PR #14, opened 2026-04-03
feat: A2UI Workflow Runtime — agents emit JSON workflow specs, runtime executes them interactively

najmuzzaman-mohammad:
## Summary

Agents can now generate interactive workflow experiences as JSON specs instead of requiring hardcoded Go TUI code. The email triage app (550 LOC of Go) becomes a ~50 line JSON spec.

**Phase 0: Prerequisites**
- Fixed broken skill create/invoke HTTP contracts (channel.go → broker.go)
- Extracted shared template utilities from composio_workflows.go to internal/workflow/ (DRY)

**Phase 1: Spec Validation**
- WorkflowSpec types with 5 interaction primitives: select, confirm, edit, submit, run
- Validator with cycle detection, transition validation, provider checks
- Email-triage (branching) and deploy-check (linear) specs validated

**Phase 2: Core Runtime**
- State machine with 7 states and error recovery (retry/skip/halt)
- ActionProvider, AgentDispatcher, StateStore interfaces (dependency injection)
- WorkflowView bubbletea model (full-screen TUI, spinner, action hints)
- A2UI interaction layer: ComponentAction, focus tracking, selection state

**Phase 3: Expansions**
- History & Replay (per-workflow + global event logging)
- Dry Run Mode (preview actions without executing)
- Workflow Versioning (linear versions, manifest, rollback)
- Sub-workflow Composition (depth-limited to 3, cycle detection)
- S4 Accelerator (prompt template + validation loop for agent generation)
- Analytics Dashboard (cross-workflow stats from event logs)

**Phase 4: Integration**
- Skill invoke returns workflow spec, channel launches workflow view
- Save-as-skill posts workflow definition to broker

## Test Coverage

72 tests across the workflow package covering:
- Spec parsing and validation (16 tests)
- Runtime state machine (26 tests)
- History, versioning, composition, generation, analytics (30 tests)
- All 18 packages pass

## Pre-Landing Review

- CEO Review: CLEAR (6 scope expansions accepted)
- Eng Review: CLEAR (3 issues resolved, 0 critical gaps)
- Design Review: CLEAR (4/10 → 8/10, 8 design decisions added)
- 2x Codex outside voice reviews incorporated

## Test plan
- [x] All Go tests pass (18 packages, 72 workflow tests)
- [x] go build ./... clean
- [x] Existing tests unbroken (composio_workflows refactor verified)


FranDias:
## Staff review — do not merge

Runtime/validator are sound as independent pieces, but wiring between runtime → view → providers has **four independent bugs that break both canonical example workflows**. 72 tests pass but they exercise the runtime API in isolation — they don't drive the view, which is where the bugs live.

### CRITICAL — Transitions to submit/run steps silently hang

`internal/workflow/view.go:6142` — when an action's `transition` points to a `submit` or `run` step (common — both examples do this), the view gets stuck forever.

**Flow:** user presses `d` → `HandleAction` has no `Execute` → calls `transitionTo("dismiss")` → runtime now on `dismiss` submit step in `StateActive` → view returns, no `tea.Cmd` issued. `handleKey` bails because `state != StateAwaitingInput`. `handleActionResult` only fires on `actionResultMsg`, nothing dispatched one. Submit step's `Execute` never runs.

**Breaks both example workflows:**
- `emailTriageSpec`: `d`/dismiss path is dead
- `deployCheckSpec`: `review → notify-team → execute-deploy` dead at `notify-team`

Integration tests at `integration_test.go:3122-3143` assert `step.Execute` is populated after transition but never call `CompleteAction` — that's why the bug slipped through.

**Fix:** After any transition that lands on submit/run with `step.Execute`, dispatch `v.executeAction` as a `tea.Cmd`. Extract the logic from `handleActionResult` into a helper and call it from `handleKey` and `Init`.

### CRITICAL — `AgentDispatcher` configured but never called

`runtime.go:3571-3572` stores a dispatcher, `channel.go:348` wires up a real one, but **nothing reads `r.agentDispatcher`**. Only `.Dispatch(...)` call sites are tests. `run` steps with just `agent: "triage"` (no `execute`) do nothing — workflows hang on agent steps.

Worse, `validator.go:5523-5528` explicitly permits this shape, and `generate.go:2454` documents it in the schema ("Has agent/agentPrompt OR workflow field"). Agent-emitted specs following the documented schema silently hang.

**Fix:** At parse time, synthesize `Execute{Provider: ProviderAgent, Slug: step.Agent, Prompt: step.AgentPrompt}` for run steps with only `Agent` set. Or reject them in the validator. Don't ship with half-wired state.

### CRITICAL — `resolveDataRefs` broken for multi-segment paths

`cmd/wuphf/channel_workflow_providers.go:832-850` — `strings.TrimPrefix(ref, "/")` then `dataStore[key]`. A ref like `/triageResult/draftReply` looks up the literal key `"triageResult/draftReply"` in the flat map. Silent failure (returns `{"ref": "..."}` wrapper).

Also only matches top-level `{"ref": "/path"}` wrappers — nested `data: {"body": {"ref": "/draft"}}` isn't resolved.

The canonical `emailTriageSpec` uses this exact pattern:
```json
{"provider": "composio", "action": "GMAIL_SEND_EMAIL",
 "data": {"ref": "/triageResult/draftReply"}}
```

Today it sends Gmail a payload of `{"ref": "/triageResult/draftReply"}` instead of the actual draft. **The canonical send-email flow is silently broken.**

**Fix:** Use `resolvePointerPath(dataStore, ref)` (already exists in `runtime.go:3950`). Recurse into nested maps/slices. Surface error on unresolved refs.

### CRITICAL — `CompleteAction` clobbers top-level data store keys

`runtime.go:3769-3777` — writes the result once under stepID, then **merges every top-level key into the root store**. Since `composioActionProvider.Execute` wraps every response as `{"provider", "action", "dry_run", "response"}`, every action unconditionally overwrites `/provider`, `/action`, `/dry_run`, `/response`.

Consequences:
- Any spec using `/response` as a dataRef gets clobbered by the next action
- Agent-emitted spec with an output key named `action` gets silently overwritten
- No way to reference a previous action's output — each one overwrites `/response`

**Fix:** Drop the top-level merge (lines 3774-3776). Downstream steps reference via `/<stepID>/<key>` (already populated). If you want a "last result" convenience, name it `/_last_result` explicitly.

### IMPORTANT — 30-second synchronous HTTP call inside bubbletea `Update`

`cmd/wuphf/channel.go:361-363` — `hydrateDataSources` runs synchronously inside the bubbletea event loop. Freezes the TUI (sidebar, notifications, key input) for up to 30s per data source, serially. Three slow data sources = 90s frozen.

**Fix:** Return a `tea.Cmd` that hydrates in a goroutine, posts `channelWorkflowReadyMsg` when done. Show loading state.

### IMPORTANT — Retry / skip UI is non-functional

`view.go:6306`, `view.go:6454` — shows `[r] Retry [s] Skip [Esc] Cancel` when in StateError, `[Enter] Return [s] Save as skill` in StateDone. **None of these keys are wired.** `handleKey` returns early for any state other than `StateAwaitingInput`. Also `handleActionResult` discards `CompleteAction` error, so transient failures show a hung spinner.

**Fix:** Wire the keys, or remove the misleading UI hints.

### IMPORTANT — No schema version, no action allowlist

No `schemaVersion` field on `WorkflowSpec`. Saved specs in `$CONFIG/workflows/interactive/*.v*.json` will drift silently as semantics change.

TODOS.md:9 acknowledges any agent-emitted spec can invoke any Composio action. `hydrateDataSources` fires these **automatically at workflow launch before any user interaction**. Malicious/confused agent specs can send email, post to Slack, exfiltrate data before user sees any "confirm" screen.

**Fix:**
1. Add `SchemaVersion` field, validator rejects unknown versions
2. Spec-level `allowedActions` allowlist enforced by validator, OR per-action user prompt on first use remembered per-spec
3. At minimum, surface to user *before launching* the full list of actions/data sources the spec will hit

### IMPORTANT — Double transition after non-execute actions

`view.go:6279`, `runtime.go:3729` — `HandleAction` calls `r.transitionTo(transition)`, then `handleKey` calls `v.runtime.Transition(transition)` again. Currently idempotent but fragile.

**Fix:** Pick one — either `HandleAction` returns target stepID (view transitions), or `HandleAction` transitions and view doesn't call `Transition` again.

### SUGGESTIONS

- `LogExecution` and the events.jsonl log exist in tests but **no production code path writes to them**. Observability claim in PR description is infrastructure-only.
- `channel_workflow_providers.go:870` — `fmt.Sprintf("...%q...", ...)` for JSON is close to but not identical to JSON. Use `json.Marshal`.
- `view.go:6269-6276` — item-removal heuristic ("first action is primary") is spec-coupled. Add explicit `ActionSpec.ConsumesItem bool`.
- `channel.go:340-342` — broker token from world-readable `/tmp/wuphf-broker-token`. Guard behind env flag or move path.

---

A single view-level test that drives `emailTriageSpec` through `list → approve → triage → confirm-send → y → list → d → list`, asserting provider + agent calls, would have caught all four criticals.

FranDias (commented):
## Staff Review

### [CRITICAL] Broker token in world-readable `/tmp/wuphf-broker-token`

Any process on the machine can read the token and impersonate the user. **Fix merged via #19** (moved to `~/.wuphf/broker-token`).

### [CRITICAL] Bare JSON workflow injection from any agent message

Any message containing `{` + `"steps"` + `"id"` gets parsed as executable workflow. A compromised agent can inject arbitrary workflow specs. **Fix merged via #19** (bare JSON detection removed).

### [IMPORTANT] `HandleAction` errors silently swallowed

`view.go` ~line 886: all errors treated as "unknown key" — real state machine bugs are invisible. Distinguish sentinel "no matching key" from actual errors.

### [IMPORTANT] Edit mode drops non-ASCII characters

`view.go` ~line 838: `len(key) == 1` only matches ASCII. Use `msg.Runes` like the interview handler does.

### [IMPORTANT] Double transition after `HandleAction`

`HandleAction` already calls `transitionTo` internally, then view calls `Transition()` again. Remove the redundant call.

### [SUGGESTION] `stepHistory` accessed without lock in `View()` — data race
### [SUGGESTION] Two `WorkflowSpec` types with same name in different packages — rename one

PR #74, opened 2026-04-15
fix(openclaw): make the bridge actually work against a real OpenClaw daemon

najmuzzaman-mohammad:
## Summary

The OpenClaw bridge shipped in #73 could not talk to a real OpenClaw daemon. Verified against OpenClaw 2026.4.14 on loopback, four protocol mismatches surface immediately. This PR fixes all of them and proves the fix end-to-end against a running daemon.

**Bugs fixed:**

1. **Token-only auth gets ZERO scopes.** The gateway wipes all requested scopes for token-authed clients without a device identity (`server.impl:21832`). Every session method errored out `missing scope: operator.read`. We now generate an Ed25519 keypair (persisted at `~/.wuphf/openclaw/identity.json`, 0600), sign a v3 pipe-delimited payload bound to the server-issued nonce, and present as a device — the gateway auto-approves silently on loopback.
2. **Protocol version.** Sent `minProtocol:1, maxProtocol:2` → server hard-rejects (`server.impl:22163` requires 3). Bumped to 3.
3. **`client.id` / `client.mode` are constant enums.** Free-form strings fail JSON schema validation. Hardcoded to `gateway-client` / `backend` (matches OpenClaw's own default client at `client-CJI3Hi9b.js:518`).
4. **`hello-ok` is wrapped in a `res` frame.** Real server sends `{type:"res", id, ok:true, payload:{type:"hello-ok", protocol:3, ...}}`, not a bare `{type:"hello-ok"}`. Rewrote `doHandshake` to read the wrapped form.

Also discovered and fixed in the process:
- `SessionRow.SessionKey` → `.Key` (the real daemon field is `key`)
- Removed `SessionsHistory` — method doesn't exist on the real daemon
- Removed the delta/streaming code path — `sessions.messages.subscribe` only emits finalized transcript entries; there is no streaming channel
- Bridge now filters role=user echoes (the daemon re-pushes our own outbound messages), forwards only role=assistant

## Real-daemon receipts

`OPENCLAW_TOKEN=… go run ./cmd/wuphf-oc-probe` against a local `openclaw daemon` listening on `ws://127.0.0.1:18789`:

```
deviceId: 290fc10f07b9387c ...
dialed + handshake ok
found 1 session(s)
first session: key=agent:main:main displayName="" kind=direct
subscribed
sent: runId=probe-1776255224647274000 status=started messageSeq=2
  EVT message role="user" text="hello from wuphf Go probe"
  EVT message role="assistant" text="⚠️ Agent failed before reply: No API key found for provider ..."
PASS
```

The assistant's text is an OpenClaw config error (no OpenAI key), not a bridge error — that's a valid round-trip session.message event, proving the whole pipeline.

## Test plan

- [x] `go test -race ./...` — all 21 packages pass
- [x] `go run ./cmd/wuphf-oc-probe` against real daemon — PASS (see receipts above)
- [x] E2E bridge test updated to match real protocol (challenge → wrapped hello-ok → role-aware events)
- [x] Unit tests for DeviceID stability, signature round-trip, wire format of BuildDeviceAuthPayloadV3

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,8 @@ Already running [OpenClaw](https://openclaw.ai) agents? You can bring them into 
 
 Inside the office, run `/connect openclaw`, paste your gateway URL (default `ws://127.0.0.1:18789`) and the `gateway.auth.token` from your `~/.openclaw/openclaw.json`, then pick which sessions to bridge. Each becomes a first-class office member you can `@mention`. OpenClaw agents keep running in their own sandbox; WUPHF just gives them a shared office to collaborate in.
 
+WUPHF authenticates to the gateway using an Ed25519 keypair (persisted at `~/.wuphf/openclaw/identity.json`, 0600), signed against the server-issued nonce during every connect. OpenClaw grants zero scopes to token-only clients, so device pairing is mandatory — on loopback the gateway approves silently on first use.
+
 ## External Actions
 
 To let agents take real actions (send emails, update CRMs, etc.), WUPHF ships with two action providers. Pick whichever fits your style.
--- a/cmd/wuphf-oc-probe/main.go
+++ b/cmd/wuphf-oc-probe/main.go
@@ -1,0 +1,108 @@
+// wuphf-oc-probe is a tiny smoke test that drives the OpenClaw bridge client
+// against a running local daemon. Build and run it with:
+//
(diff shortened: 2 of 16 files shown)


FranDias (approved):

PR #83, opened 2026-04-16
[codex] internalize autonomy coaching and surface office activity

najmuzzaman-mohammad:
## What changed
This PR internalizes the autonomy coaching work into the product instead of relying on the operator to steer every loop manually, and it makes the Web UI show far more of the office's live work.

The main changes are:
- tighten live-business task hygiene so proof packets, review bundles, and other internal theater stop counting as progress in live lanes
- strengthen the from-scratch launch path and autonomy prompts so the office creates missing capabilities and takes the next real business step
- improve broker/runtime durability around task routing, worktree assignment, resume behavior, Codex usage accounting, and panic recovery in background delivery loops
- add office-wide activity visibility in the UI so humans can see active lanes, blockers, recent actions, decisions, due automations, and live agent work without hunting through channels
- preserve recent mainline functionality from per-agent providers, scoped memory, GBrain backend support, OpenClaw bridge work, token usage tagging, and the website/pages rollout

## Why it changed
The evals exposed two core problems:
1. the office still needed too much coaching to avoid planning theater and take real business actions
2. even when the agents were doing interesting work, the UI hid too much of it, which made the system feel opaque and untrustworthy

This PR pushes the coached behavior into the runtime and makes the visible control plane match what the agents are actually doing.

## Impact
- live-business lanes are stricter about what counts as completion
- from-scratch runs are more directive-first and less template-theater prone
- usage accounting now reports real Codex token burn instead of zeros
- the activity UI now shows office-wide movement instead of only the current channel's narrow slice
- CI is no longer blocked on missing checked-in review-bundle artifacts because those tests are now hermetic

## Root cause
The office/runtime had an artifact-first bias: it treated internal evidence and scaffolding as acceptable progress, and the UI was scoped too narrowly to the current channel to show the real work already happening elsewhere in the office.

## Validation
Passed locally:
- `go test ./... -count=1`
- `go build ./...`
- `node` inline script parse for `web/index.html`
- `git diff --check`

Additional verification:
- audited overlapping merged PRs from the last 3 days and restored the one dropped behavior found during merge resolution: the panic guards from PR #70

## Notes
- This branch includes a merge from current `main` and preserves the recent provider, memory, OpenClaw, usage-tagging, website, and GBrain work landed there.


FranDias (approved):

PR #96, opened 2026-04-16
feat(web): migrate 13k-line vanilla JS to React + TypeScript

najmuzzaman-mohammad:
## Summary

Full migration of the WUPHF web UI from a single 13,486-line `index.html` to a modular React 19 + TypeScript codebase. The legacy file is preserved as `web/index.legacy.html`; the Go server serves the Vite build output when it exists, falling back to the legacy file otherwise — zero breaking changes.

## Stack

- **Vite 6** — build + dev server
- **React 19 + TypeScript 5.8** — UI + type safety
- **TanStack Query** — server state, replaces manual fetch/poll loops
- **TanStack Router** — type-safe client routing (registered, hash routing still drives navigation)
- **Zustand** — client state (current channel, theme, DM mode, sidebar)

## Architecture (50 source files)

```
web/src/
├── api/client.ts              Typed API client (31 methods, mirrors WuphfAPI)
├── stores/app.ts              Zustand store (all global UI state)
├── hooks/                     TanStack Query hooks + useAgentStream (SSE) + useKeyboardShortcuts
├── components/
│   ├── layout/                Shell, Sidebar, ChannelHeader, DisconnectBanner
│   ├── messages/              MessageFeed, Composer (w/ slash cmds), ThreadPanel, DMView
│   ├── sidebar/               AgentList, ChannelList, AppList, UsagePanel
│   ├── apps/                  9 app panels (Tasks, Policies, Calendar, Skills, Activity, Receipts, Studio, HealthCheck, Requests)
│   ├── onboarding/            Wizard (6-step), SplashScreen
│   ├── agents/                AgentPanel, AgentWizard
│   ├── search/                SearchModal
│   ├── channels/              ChannelWizard
│   └── ui/                    Toast (global showNotice)
├── lib/                       format, markdown renderer, constants, router
└── styles/                    Ported CSS (global, layout, messages, agents, search, onboarding)
```

## Go server change (8 lines)

```go
// Prefer web/dist/ (Vite build output) when it exists.
if distDir := filepath.Join(webDir, "dist"); dirExists(distDir) {
    webDir = distDir
}
```

Falls back to `web/` if `dist/` is missing. Legacy behavior preserved.

## Feature parity

- All **31 legacy API methods** covered by typed client
- All **9 app panels** implemented
- Message feed: grouping, date separators, markdown, reactions, status/routing/skill cards
- **17 slash commands** implemented (legacy had ~20 — missing `/connect`, `/insert`, `/rewind` for future)
[...]

PR #98, opened 2026-04-16
feat(web): React + TypeScript migration with embedded assets (v2)

najmuzzaman-mohammad:
## Summary

Second attempt at migrating the 13,486-line vanilla JS web UI to React + TypeScript. First attempt (#96) was reverted (#97) due to a blank-page regression. This PR fixes the root cause AND a pre-existing 404 bug on fresh installs.

### Fixes

1. **Blank-page bug (cause of #97 revert):** The previous `App.tsx` called `useKeyboardShortcuts()` after two conditional early returns, violating React's Rules of Hooks. In production, this caused React to throw silently during state transitions. Fixed: all hooks are now called unconditionally at the top of each component.

2. **404 on fresh installs:** `.goreleaser.yml` shipped only the binary, no `web/` directory. Users installing via `curl | bash` got 404 on every route. Fixed: embed `web/dist/` into the binary with `//go:embed all:web/dist`. Serve from embedded FS when filesystem assets are missing.

### Fail-loud diagnostics

If any of these safeguards fire, the user sees a red banner with the error — never a blank page:

- Static skeleton in `index.html` before React mounts
- `window.error` + `unhandledrejection` handlers that render to DOM
- 10s boot-timeout warning if React doesn't mount
- React `ErrorBoundary` around the full app
- `try/catch` around `createRoot().render()` with visible error output

### Stack

- **Vite 6** (build + dev server)
- **React 19 + TypeScript 5.8**
- **TanStack Query** (server state, replaces manual poll loops)
- **TanStack Router** (registered, hash routing still drives nav)
- **Zustand** (client state)

### CI changes

- `ci.yml`: new `web` job runs `npm ci && npm run typecheck && npm run build`, uploads dist artifact; `go` job downloads it before running tests
- `release.yml`: builds React before goreleaser so the embedded bundle contains real assets

### Verification performed

All verified in a real browser (via gstack's `$B` tool):

- [x] TypeScript clean (`tsc --noEmit` exit 0)
- [x] Vite build clean (318kb JS / 94kb gzip, 31kb CSS)
- [x] Go build clean (exit 0)
- [x] Go vet clean (exit 0)
- [x] Binary embeds web/dist: 39MB binary, works from empty directory
- [x] `curl http://localhost:7891/` returns 200 with full React HTML
- [x] React mounts in headless Chromium: `#root` has children, 7 agents visible, all 9 app panels in sidebar
- [x] Tasks panel renders real data: IN PROGRESS / OPEN / BLOCKED lanes with actual tasks
- [x] Requests panel renders real pending approvals with PENDING badges
- [x] Works with `--provider claude-code`
- [x] Works with `--provider codex`
- [x] Broker runs 1+ minute without crashes under both providers

Screenshots of working UI attached in the PR thread.

### Commits on this branch

```
a3484db5 feat(web): migrate to React + TypeScript with embedded assets
```

Squashed from the iterative work in the worktree.


FranDias (approved):

PR #101, opened 2026-04-16
fix(web): resolve React error #31 when clicking agent panel

abeeb-rai:
## Summary
- **Bug**: Clicking any agent in the sidebar opened the AgentPanel which rendered `agent.provider` (a `ProviderBinding` object from the backend) directly as JSX, causing React error #31 ("Objects are not valid as a React child — object with keys {}")
- **Root cause**: Backend serializes `provider` as a struct (`{kind, model, openclaw}`), which becomes `{}` for default-provider agents. The frontend typed it as `string` and rendered it directly.
- **Fix**: Updated `OfficeMember.provider` type to `ProviderBinding | string`, extract `provider.kind` as a string before rendering, and skip the row when no value exists.

## Test plan
- [ ] Launch `./wuphf --memory-backend none --unsafe`
- [ ] Open http://localhost:7891
- [ ] Click on any agent in the sidebar — AgentPanel should open without crashing
- [ ] Verify provider info shows correctly for agents with a configured provider
- [ ] Verify no provider row appears for agents with default/empty provider

### DIFF
--- a/web/src/api/client.ts
+++ b/web/src/api/client.ts
@@ -156,6 +156,11 @@ export function toggleReaction(msgId: string, emoji: string, channel: string) {
 
 // ── Members ──
 
+export interface ProviderBinding {
+  kind?: string
+  model?: string
+}
+
 export interface OfficeMember {
   slug: string
   name: string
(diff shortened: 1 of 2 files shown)


FranDias (approved):

drourke (approved):

najmuzzaman-mohammad (approved):

PR #117, opened 2026-04-17
fix(web): make pending requests actionable + show count badge

najmuzzaman-mohammad:
## Summary
Follow-up to #116 — three bugs all visible from the chat 409.

1. **Answer buttons missing.** Broker returns `options`, type + renderer said `choices`. Now reads either; renders title/context/BLOCKING badge; primary styling on recommended option.
2. **No badge on Requests sidebar.** Operator had no signal answers were waiting. AppList polls `/requests` every 5s and badges the pending count.
3. **Composer 409 was a dead toast.** "request pending; answer required" flashed a 4s toast with no recovery path. Composer `onError` detects that case and routes to the Requests app.

Bonus: scheduler jobs have no `id` (broker uses `slug`). Calendar and Activity "Due automations" were giving every job the same `undefined` key. Fall back to slug, then index.

## Test plan
- [x] `npm run typecheck` clean
- [x] Sidebar shows "Requests · 13" badge against current broker
- [x] Requests panel renders answer buttons (Connect One / Composio / Skip / etc.) with recommended option highlighted
- [x] Sending a message while a request is pending toasts "Answer the pending request to resume chat." and switches to Requests app
- [x] Calendar renders 24 jobs without React key warnings

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #119, opened 2026-04-17
fix(web): inline interview bar above composer (matches TUI)

najmuzzaman-mohammad:
## Summary
The full-screen overlay from #117 had three real bugs:

1. Options with `requires_text` returned the broker hint as a 400. UI surfaced it as a toast with no way to type the answer.
2. No escape hatch — truly blocking with no close button trapped users when every option needed text.
3. Modal blocked the entire app instead of behaving like the TUI prompt above the input.

Replaced with `InterviewBar` mounted inline above the Composer in both the channel view and DM view. Mirrors the TUI:

- BLOCKING badge, @from, channel, `N/M` counter, prev/next arrows for cycling the queue
- Picking an option with `requires_text` switches the bar to text input mode, using the broker's `text_hint` as placeholder; submit posts `custom_text`
- Close (×) skips locally and POSTs `/signals kind=pause` so the office stops waiting
- Recommended option keeps primary styling
- `answerRequest` accepts optional `custom_text`; `InterviewOption` carries `requires_text` + `text_hint`

## Test plan
- [x] Bar renders above composer with `1/9` counter against current broker
- [x] Prev/next arrows cycle without dismissing
- [x] Picking a `requires_text` option switches to a textarea with the broker's hint
- [x] Close button pauses the office and toasts the user
- [x] Bar disappears when all blocking requests are answered/skipped


FranDias (approved):

PR #127, opened 2026-04-17
feat(updatecheck): passive "new version available" banner

FranDias:
## Summary
- Prints a one-line upgrade nudge at interactive startup when the installed binary is behind the latest GitHub release.
- Uses the `releases/latest` redirect (no payload) + a 24h disk cache at `~/.wuphf/update-check.json` — keeps the "no telemetry" promise on the website.
- Suppressed for dev builds (`""`, `"dev"`, `"0.1.0"`) and when `WUPHF_NO_UPDATE_CHECK=1`. Only fires on TUI/web launches, not subcommands or non-interactive (`--cmd`, piped) flows.

## Why
Users installing via `curl | sh` or `npx wuphf` currently have no in-product signal when a new release lands — v0.8.1's broken install silently stranded users on a dead binary. This closes that gap without adding analytics.

## Test plan
- [x] `go test ./internal/updatecheck/...` — 19 subtests green (version comparison, env handling, cache TTL, dev-build suppression, end-to-end redirect via `httptest.NewServer`).
- [x] `go build ./...` clean.
- [ ] Manual: run a stamped build (`-ldflags "-X …buildinfo.Version=0.1.0"` → no banner; set cache to a higher latest → banner prints).

### DIFF
--- a/cmd/wuphf/main.go
+++ b/cmd/wuphf/main.go
@@ -14,6 +14,7 @@ import (
 	"github.com/nex-crm/wuphf/internal/config"
 	"github.com/nex-crm/wuphf/internal/team"
 	"github.com/nex-crm/wuphf/internal/teammcp"
+	"github.com/nex-crm/wuphf/internal/updatecheck"
 	"github.com/nex-crm/wuphf/internal/workspace"
 )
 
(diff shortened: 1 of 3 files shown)


FranDias:
Dropping: this adds a phone-home to github.com which contradicts the website's 'no phone-home' promise. If we want passive update warnings later, we should either reword the landing-page copy first or gate this behind an opt-in env var.

PR #139, opened 2026-04-17
feat(team): spawn interactive claude panes in web mode, fall back to headless

najmuzzaman-mohammad:
## Summary
- Web mode was always routing agent turns through \`claude --print\`. Under Anthropic's recent policy, that path consumes the separate headless/extra-usage quota instead of the user's normal Claude Code subscription.
- Launcher now tries to spawn one interactive \`claude\` pane per agent in a detached tmux session (same machinery TUI mode already uses), and dispatch sends messages via \`tmux send-keys\`. Agent replies come back through the existing \`team_broadcast\` MCP tool — no new output-capture wiring needed.
- Fallback: if \`tmux\` is not on PATH or the spawn fails, we keep the existing headless \`claude --print\` path and surface the billing tradeoff to the user via stderr + a #general system message.

## Key pieces
- \`Launcher.paneBackedAgents\` bool gates the new path.
- \`trySpawnWebAgentPanes\` — detached \`tmux new-session\` with a placeholder pane 0 (so pane 1..N match \`agentPaneTargets()\`), then reuses \`spawnVisibleAgents\` / \`spawnOverflowAgents\`.
- \`shouldUseHeadlessDispatch\` replaces the two \`l.webMode\` branches in \`sendTaskUpdate\` and \`sendChannelUpdate\`.
- \`reportPaneFallback\` surfaces the tradeoff so the user knows why they're on the expensive quota and how to fix it.

## Tradeoffs (acknowledged)
- Losing the live per-tool-use \"thinking…\" indicator in the web UI: pane mode only surfaces full agent posts via \`team_broadcast\`. A later PR can add \`tmux capture-pane\` polling to restore it.
- Requires tmux on the server. Local / self-hosted is fine; a hosted deployment would need tmux installed in the image.

## Test plan
- [x] \`go build ./...\`
- [x] \`go test ./internal/team/...\` (plain run clean)
- [x] New unit test \`TestShouldUseHeadlessDispatch\` covers codex / web-headless / web-pane / tui-pane matrix
- [ ] Manual: \`wuphf web\` with tmux installed → CEO and specialists reply to DMs and untagged messages via interactive panes
- [ ] Manual: \`wuphf web\` with tmux uninstalled → fallback kicks in, system message appears in #general

## Known flakiness (pre-existing on main)
- \`-race\` suite occasionally trips on \`TempDir RemoveAll cleanup\`. Verified present without this PR's changes. Not addressed here.

### DIFF
--- a/internal/team/dispatch_test.go
+++ b/internal/team/dispatch_test.go
@@ -75,6 +75,33 @@ func TestMemberEffectiveProviderKind_DefaultsToClaudeWhenAllEmpty(t *testing.T) 
 	}
 }
 
+func TestShouldUseHeadlessDispatch(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name             string
+		provider         string
+		webMode          bool
+		paneBackedAgents bool
+		want             bool
+	}{
+		{"tui mode, claude → pane", "claude-code", false, false, false},
+		{"tui mode, codex → headless", "codex", false, false, true},
+		{"web mode, no panes → headless", "claude-code", true, false, true},
+		{"web mode with panes → pane", "claude-code", true, true, false},
+		{"web mode, codex always headless even if panes flag set", "codex", true, true, true},
+	}
+	for _, tt := range tests {
+		l := &Launcher{
+			provider:         tt.provider,
+			webMode:          tt.webMode,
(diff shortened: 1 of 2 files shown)


FranDias (approved):

PR #141, opened 2026-04-17
feat(team): stream interactive claude pane output to the web UI

najmuzzaman-mohammad:
## Summary

Follow-up to #139, which spawns one interactive `claude` pane per agent in
tmux when WUPHF runs in web mode. PR #139 left the live output gap: the
web UI's per-agent stream stayed empty during a turn because the tmux
pane output was never wired into `broker.AgentStream(slug)`.

This PR adds a lightweight per-pane poller that bridges tmux to the web
UI stream:

- One goroutine per pane-backed agent. Shells out to
  `tmux -L wuphf capture-pane -p -J -t <paneTarget>` on a 1s interval.
- ANSI escape sequences and common control chars (CR, BEL) are stripped
  with a small inlined regex. No new dependencies.
- Line-level diffing against the previous snapshot. Claude's TUI
  re-renders frequently (cursor-position tricks), so the poller tracks a
  line multiset and only emits lines that appear in the new capture but
  weren't in the previous one, preserving their order. Blank lines are
  dropped.
- Pushes plain text to `AgentStream(slug).Push`, matching the existing
  headless Claude stdout convention (`internal/team/headless_claude.go`).
  The web UI's `StreamLineView` already renders raw-text lines via the
  `stream-line-raw` branch.
- Per-line cap of 64KB with a `...[truncated]` marker. Tolerates
  transient tmux errors with a 2s backoff; stops polling a pane after
  5 consecutive failures and logs once.
- Only runs when `l.paneBackedAgents == true`. TUI mode and the codex
  runtime are untouched.
- Tied to `l.headlessCtx` so shutdown cancels all pollers cleanly.

## Known limitation

Claude's TUI box-drawing characters and partial frame rewrites still
leak through as cosmetic noise. Cleaner options (stream-json, TUI-aware
post-processing) are out of scope here — this PR unblocks the "live"
feeling today.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/team/... ./internal/provider/...` — unit
  tests for `stripANSI` and `diffPaneLines` pass; pre-existing
  operation-matrix failure on main is unrelated.
- [x] `go test -race` on the new tests passes.
- [ ] Manual: launch WUPHF in web mode with tmux available, send an
  @mention, confirm the agent's stream panel starts showing live
  Claude output within ~1s.
- [ ] Manual: kill a pane mid-turn, confirm polling stops after 5
  consecutive failures and logs to stderr without spinning.

### DIFF
--- a/internal/team/launcher.go
+++ b/internal/team/launcher.go
@@ -3971,6 +3971,11 @@ func (l *Launcher) LaunchWeb(webPort int) error {
 		l.resumeInFlightWork()
 	}
 
+	// Stream tmux pane output to the web UI's per-agent stream so users see
+	// live Claude TUI activity (thinking, tool calls, responses) during a
+	// pane-backed turn. No-op when paneBackedAgents is false.
+	l.startPaneCaptureLoops(l.headlessCtx)
+
 	go l.notifyAgentsLoop()
 	go l.notifyTaskActionsLoop()
 	go l.notifyOfficeChangesLoop()
--- a/internal/team/pane_capture.go
+++ b/internal/team/pane_capture.go
@@ -1,0 +1,172 @@
+package team
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"regexp"
+	"strings"
+	"time"
+)
+
+// Pane capture defaults. Values are intentionally not user-configurable:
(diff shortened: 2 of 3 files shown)


FranDias (approved):

PR #148, opened 2026-04-18
feat(web): restore agent-panel enable/disable toggle + Remove

najmuzzaman-mohammad:
## Summary

The Vanilla JS → React migration (PR #96) dropped two agent-panel actions. Backend endpoints (`/channel-members`, `/office-members`) were never touched, so this is a pure UI restore.

- **Enable/Disable toggle** — per-channel switch at the top of the agent panel. Posts `/channel-members {action: enable|disable, channel, slug}`. This is the "move an agent in/out of a channel" semantics.
- **Remove agent** — destructive action gated on `!built_in && slug !== 'ceo'` to mirror broker validation (`broker.go:5458`, `:5924`). Uses the existing `confirm()` dialog host.
- **OfficeMember type** — surfaced the `built_in` and `disabled` fields the broker already serializes (`broker.go:253`, `:7045`).
- **Styles** — ported `.agent-toggle` + `.agent-panel-stat` from `index.legacy.html`; they were never copied into `web/src/styles/agents.css`.

## What I skipped (and why)

- **Header bell icon** — dead button in legacy (no handler, no listener, no notification source). The Requests app in the sidebar already shows a pending-count badge (`AppList.tsx:24`). Resurrecting the bell would be placebo UI.
- **Edit agent** (the third legacy panel action) — `AgentWizard.tsx` is create-only. Teaching it to accept an existing agent + wire a PATCH flow is a separate chunk of work.

## Edge cases

Broker's `enable` only lifts the `Disabled` flag; it does not add an agent to `ch.Members`. If the user toggles ON an agent that isn't already a member, the server state won't change. The panel handles this with a 3s timeout that clears the optimistic override so the switch snaps back to the real state instead of getting stuck.

## Test Coverage

No web unit-test framework in `web/src/` (same as PR #144). Verified via:
- `tsc --noEmit` — clean
- `vite build` — clean (413 KB JS / 120 KB gzipped)
- Smoke e2e (`web/e2e/tests/smoke.spec.ts`) only asserts `.agent-panel` is visible when an agent is selected; class preserved.

## Test plan

- [ ] Open panel on a non-built-in agent: toggle appears showing current state for the active channel
- [ ] Toggle off → agent disabled in channel, success toast, sidebar polling reflects it
- [ ] Toggle on (for a member) → agent enabled, success toast
- [ ] Toggle on for a non-member (edge case) → switch reverts after 3s, no error toast
- [ ] Remove on non-built-in → confirm dialog → success toast → panel closes, agent vanishes from `/office-members`
- [ ] Open panel on CEO → no toggle, no Remove button
- [ ] Open panel on any `built_in` agent → toggle shown, no Remove button

### DIFF
--- a/web/src/api/client.ts
+++ b/web/src/api/client.ts
@@ -180,7 +180,7 @@ export interface OfficeMember {
   task?: string
   channel?: string
   provider?: ProviderBinding | string
-  /** Broker-provided: serialized as `built_in`. Built-ins (e.g., CEO) cannot be removed. */
+  /** Broker-provided: serialized as `built_in`. Built-ins cannot be removed. (CEO is guarded by a separate slug check.) */
   built_in?: boolean
   /** Per-channel disabled state when the list is sourced from `/members?channel=…`. */
   disabled?: boolean
--- a/web/src/api/client.ts
+++ b/web/src/api/client.ts
@@ -180,6 +180,10 @@ export interface OfficeMember {
   task?: string
   channel?: string
   provider?: ProviderBinding | string
+  /** Broker-provided: serialized as `built_in`. Built-ins (e.g., CEO) cannot be removed. */
+  built_in?: boolean
+  /** Per-channel disabled state when the list is sourced from `/members?channel=…`. */
+  disabled?: boolean
 }
 
 export function getOfficeMembers() {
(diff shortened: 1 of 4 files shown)


FranDias (approved):

PR #153, opened 2026-04-19
ci: adopt nex-crm/.github scan-secrets workflow

FranDias:
## Summary

Wires wuphf into the org-wide [secretlint scan workflow](https://github.com/nex-crm/.github/blob/main/.github/workflows/scan-secrets.yml). Defense-in-depth over `lefthook.yml`'s existing grep-based `no-secrets` pre-commit — the grep only catches `api_key='...'`-shaped strings, while secretlint's `preset-recommend` covers AWS, GitHub, Slack, GCP, Linear, npm, Stripe, 1Password, Anthropic, DB connection strings, JWTs, and more.

- New `.github/workflows/secret-scan.yml` calls the reusable at v0.4.0.
- New root-level `package.json` solely to host `secretlint` + `preset-recommend` (the reusable expects `node_modules/.bin/secretlint`). `web/` and other subdirs keep their own package.json unchanged.
- The Postgres connection-string rule flagged one line in `cmd/wuphf/import.go` — a hardcoded `postgres:postgres@localhost` default-dev fallback for the local orchestrator import path. Suppressed inline with a scoped `secretlint-disable-next-line` comment rather than globally disabling the rule.

## What's *not* in scope (flagged as follow-ups)

- **`lint-go` / `test-go`** — the existing `ci.yml` does more than the reusables currently cover (govulncheck, artifact deps, `install.sh` smoke with a mock GitHub resolver). Adopting would be a downgrade. Consider adding `govulncheck` as an input to `lint-go.yml` in a future bump.
- **`commitlint`** — wuphf doesn't enforce Conventional Commits today; adding it is a policy change, not a CI lift.
- **`vhs`** — each of nex-cli/wuphf/home-automation runs VHS differently (tape count, gate style, golden-image storage). A shared workflow would be lowest-common-denominator garbage. Keeping the bespoke setup.
- **`lefthook.yml` secret-scan upgrade** — out of scope. The grep hack can be replaced with `./node_modules/.bin/secretlint` once this PR lands (single source of truth with CI). Separate PR.

## Test plan

- [x] Ran `./node_modules/.bin/secretlint '**/*'` locally — clean after the inline suppression.
- [ ] Verify CI `Secret Scan` passes on this PR.
- [ ] Verify CI catches an injected fake secret (optional; can smoke-test post-merge).

### DIFF
--- a/.github/workflows/secret-scan.yml
+++ b/.github/workflows/secret-scan.yml
@@ -1,0 +1,19 @@
+name: Secret Scan
+
+# Org-wide secretlint scan on PR-changed files (plus full scan on pushes to
+# main). Source: https://github.com/nex-crm/.github/blob/main/.github/workflows/scan-secrets.yml
+#
+# This is a belt-and-suspenders addition: lefthook's `no-secrets` pre-commit
+# hook (see lefthook.yml) already catches the common `api_key='...'` pattern
+# at commit time, but it's a local grep and won't catch token shapes that
+# don't match the regex. secretlint's preset-recommend covers AWS, GitHub,
+# Slack, Google, npm, Stripe, and more.
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  scan:
+    uses: nex-crm/.github/.github/workflows/scan-secrets.yml@v0.4.0
--- a/.secretlintrc.json
+++ b/.secretlintrc.json
@@ -1,0 +1,7 @@
+{
+  "rules": [
+    {
+      "id": "@secretlint/secretlint-rule-preset-recommend"
+    }
+  ]
+}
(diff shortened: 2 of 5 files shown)


FranDias:
Closing — GitHub Actions restriction discovered: **public repos cannot consume reusable workflows hosted in private repos**, regardless of `access_level: organization` on the private repo.

wuphf is public; `nex-crm/.github` is private → the run completes instantly with `conclusion: failure` and zero jobs. The failure mode is silent (no actionable error message in the Actions UI).

Options for a future attempt:
- Make `nex-crm/.github` public (it contains only reusable workflow YAML + composite actions, no secrets).
- Inline the secretlint scan directly into wuphf's `ci.yml` — defeats the "single source of truth" goal but works regardless of `.github`'s visibility.
- Split `.github` into a public companion (`nex-crm/.github-public`) for public consumers and keep org-private workflows where they are.

Flagged for the team. Not pursuing in this PR.

PR #154, opened 2026-04-19
ci: add secretlint PR-diff scan (inlined from nex-crm/.github)

FranDias:
## Summary

Adds a secretlint PR-diff + push-to-main scan. Defense-in-depth over `lefthook.yml`'s existing `no-secrets` pre-commit, which is a local grep and only catches `api_key='...'`-shaped strings. secretlint's `preset-recommend` covers AWS, GitHub, Slack, GCP, Linear, npm, Stripe, 1Password, Anthropic, DB connection strings, JWTs, etc.

## Why inlined, not a reusable-workflow call

wuphf is public; [`nex-crm/.github`](https://github.com/nex-crm/.github) is private. GitHub Actions **silently** fails to resolve reusable workflows from a private repo when the caller is public — instant run failure, zero jobs, no actionable error UI. The previously-closed [#153](https://github.com/nex-crm/wuphf/pull/153) surfaced this.

Rather than block on making `.github` public, the workflow logic is inlined here. Kept in structural sync with [the reusable source of truth](https://github.com/nex-crm/.github/blob/main/.github/workflows/scan-secrets.yml).

## What this adds

- `.github/workflows/secret-scan.yml` — full SHA-pinned checkout + setup-bun + deps cache + secretlint-on-changed-files
- Root-level `package.json` + `bun.lock` hosting `secretlint` + `@secretlint/secretlint-rule-preset-recommend` (workflow expects `node_modules/.bin/secretlint`). `web/` and other subdirs keep their own package.json unchanged.
- `.secretlintrc.json` extending `preset-recommend`
- One inline suppression in `cmd/wuphf/import.go:166` — a hardcoded `postgres:postgres@localhost` default-dev fallback for the local orchestrator import path. Suppressed scoped to the rule rather than disabled globally.

## Test plan

- [x] `./node_modules/.bin/secretlint '**/*'` locally: clean after the inline suppression
- [ ] CI `Secret Scan` passes on this PR
- [ ] Optional smoke: open throwaway PR with a fake `ghp_...` token and confirm CI catches it (can do post-merge)

## Not in this PR

- `lefthook.yml` still uses the local grep `no-secrets` hook. Future: replace with `./node_modules/.bin/secretlint` so local and CI use the same tool. Kept out of scope to keep the diff reviewable.
- `lint-go` / `test-go` / `commitlint` adoption: the existing `ci.yml` already does more than the org reusables would (govulncheck, install.sh smoke with a mock-GitHub resolver, nuanced SHA pins with rationale). Adopting those would be a downgrade.

## Summary by CodeRabbit

* **Chores**
  * Added automated secret-scanning to CI for pull requests and main branch to improve repository security.
  * Configured secretlint with the recommended rule preset to enforce secret-detection rules.
  * Introduced a dev-tooling package manifest to manage secret-scanning dependencies and tooling.

### DIFF
--- a/.github/workflows/secret-scan.yml
+++ b/.github/workflows/secret-scan.yml
@@ -73,27 +73,45 @@ jobs:
           PR_BASE_REF: ${{ github.base_ref }}
         run: |
           set -euo pipefail
-          # Exclude patterns: binary files + lockfiles that tend to produce noise.
-          EXCLUDE='(bun\.lock|package-lock\.json|\.png$|\.jpg$|\.jpeg$|\.gif$|\.ico$|\.woff2?$|\.ttf$|\.eot$|\.svg$|\.pdf$)'
+          # Exclude patterns: binary files + lockfiles that tend to produce
+          # noise. Anchored with `(^|/)...$` for bare filenames so that
+          # `docs/bun.lock.md` or `tools/package-lock.json.bak` aren't
+          # silently skipped. Extensions are anchored with trailing $.
+          EXCLUDE='((^|/)(bun\.lock|package-lock\.json)$|\.(png|jpe?g|gif|ico|woff2?|ttf|eot|svg|pdf)$)'
 
+          # NUL-delimited end-to-end so filenames with spaces / quotes /
+          # backslashes don't get split by xargs later. `git diff -z` and
+          # `git ls-files -z` both emit NUL-separated paths.
           if [ "$EVENT_NAME" = "pull_request" ]; then
-            files=$(git diff --name-only --diff-filter=ACMR "origin/${PR_BASE_REF}...HEAD")
+            git diff -z --name-only --diff-filter=ACMR "origin/${PR_BASE_REF}...HEAD" > /tmp/raw-files.z
           elif git rev-parse --verify HEAD~1 >/dev/null 2>&1; then
-            files=$(git diff --name-only --diff-filter=ACMR HEAD~1..HEAD)
+            git diff -z --name-only --diff-filter=ACMR HEAD~1..HEAD > /tmp/raw-files.z
           else
-            files=$(git ls-files)
+            # First push to main / shallow history edge: scan everything.
+            # One-time cost; flagged in coderabbit review.
+            git ls-files -z > /tmp/raw-files.z
           fi
 
-          files=$(echo "$files" | grep -v -E "$EXCLUDE" || true)
+          # Apply excludes. Pipeline converts NUL→NL for grep, then NL→NUL
+          # for xargs -0. Guard the grep step on non-empty EXCLUDE so an
+          # accidentally-empty regex doesn't silently match-all (grep -v ''
+          # returns 0 with all lines passing — we'd end up scanning the
+          # lockfile noise the exclude exists to suppress).
+          if [ -n "$EXCLUDE" ]; then
+            tr '\0' '\n' < /tmp/raw-files.z | grep -v -E "$EXCLUDE" | tr '\n' '\0' > /tmp/changed-files.z || true
+          else
+            cp /tmp/raw-files.z /tmp/changed-files.z
+          fi
 
-          if [ -z "$files" ]; then
+          if [ ! -s /tmp/changed-files.z ]; then
             echo "skip=true" >> "$GITHUB_OUTPUT"
           else
-            echo "$files" > /tmp/changed-files.txt
             echo "skip=false" >> "$GITHUB_OUTPUT"
           fi
 
       - name: Run secretlint
         if: steps.changed.outputs.skip != 'true'
         shell: bash
-        run: cat /tmp/changed-files.txt | xargs node_modules/.bin/secretlint
+        # -0 consumes NUL-delimited paths; -r (--no-run-if-empty) belts-
+        # and-braces over the `skip` gate above.
+        run: xargs -0 -r -a /tmp/changed-files.z node_modules/.bin/secretlint
--- a/.github/workflows/secret-scan.yml
+++ b/.github/workflows/secret-scan.yml
@@ -35,7 +35,7 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: oven-sh/setup-bun@735343b667d3e6f658f44d0eca948eb6282f2b76 # v2.0.2
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
         with:
           bun-version: "1"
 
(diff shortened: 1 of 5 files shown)


.github/workflows/secret-scan.yml
```diff
@@ -0,0 +1,99 @@
...
+
+          if [ "$EVENT_NAME" = "pull_request" ]; then
+            files=$(git diff --name-only --diff-filter=ACMR "origin/${PR_BASE_REF}...HEAD")
+          elif git rev-parse --verify HEAD~1 >/dev/null 2>&1; then
+            files=$(git diff --name-only --diff-filter=ACMR HEAD~1..HEAD)
+          else
+            files=$(git ls-files)
+          fi
+
+          files=$(echo "$files" | grep -v -E "$EXCLUDE" || true)
```
FranDias: **SUGGESTION — `grep -v -E "$EXCLUDE" || true` masks grep failures.**

If `$EXCLUDE` somehow ends up empty (env var not expanded, workflow edit typo, etc.), `grep -v -E ""` matches everything and returns 0 — no-op, silently scanning nothing. The `|| true` then hides any other error class (broken regex, missing grep, etc.).

Probably fine given EXCLUDE is hard-coded, but the `|| true` is there because grep returns 1 when no lines match, which is a legit case (all changed files were excluded → empty result). Cleaner:

```bash
if [ -n "$EXCLUDE" ]; then
  files=$(echo "$files" | grep -vE "$EXCLUDE" || true)
fi
```

This is in the shared reusable too (same logic). If you change it here, mirror upstream.

.github/workflows/secret-scan.yml
```diff
@@ -0,0 +1,99 @@
...
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "$files" > /tmp/changed-files.txt
+            echo "skip=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Run secretlint
+        if: steps.changed.outputs.skip != 'true'
+        shell: bash
+        run: cat /tmp/changed-files.txt | xargs node_modules/.bin/secretlint
```
FranDias: **IMPORTANT — `xargs node_modules/.bin/secretlint` will chunk on long file lists.**

`xargs` splits into multiple secretlint invocations when the arg list exceeds `ARG_MAX` (~2MB on Linux). secretlint then exits nonzero on any one chunk that has findings — fine. But xargs's default exit code when any child returns nonzero is 123, not the child's exit code. CI will fail (good), but the error message won't clearly attribute to which chunk.

More importantly, `xargs` without `-d '\n'` splits on any whitespace. If any changed file has a space in its name, secretlint will be invoked with the filename fragmented, leading to `ENOENT` errors and a confusing failure.

Safer:
```bash
xargs -d '\n' -r node_modules/.bin/secretlint < /tmp/changed-files.txt
```

The `-r` (`--no-run-if-empty`) guards against the zero-file case too (though the `if: skip != 'true'` gate already handles that).

Upstream reusable has the same shape — worth fixing there too.

cmd/wuphf/import.go
```diff
@@ -163,6 +163,7 @@ func importFromLegacyDB(portOverride int) (importedBrokerState, int, int, error)
 		}
 	}
 
+	// secretlint-disable-next-line @secretlint/secretlint-rule-database-connection-string
```
FranDias, line 170: Good use of scoped suppression — the `@secretlint/secretlint-rule-database-connection-string` disable is narrow enough that an AWS key or GitHub PAT accidentally put on this line would still be flagged. Verified locally.

One consistency note: the `postgres://postgres:postgres@localhost:%d/...` fallback only fires when running against a local dev orchestrator. A `TODO(fd)` or link to the onboarding doc that explains this default might save future contributors from wondering if the creds are meaningful. Not blocking.

FranDias (commented):
Verified locally: secretlint with the preset-recommend catches the PostgreSQL connection string, and the `secretlint-disable-next-line @secretlint/secretlint-rule-database-connection-string` directive correctly suppresses it — scoped to just the DB-connection rule, not other detectors. Good. A couple of minor concerns about the inline workflow itself.

.github/workflows/secret-scan.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Determine changed files
+        id: changed
+        shell: bash
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          PR_BASE_REF: ${{ github.base_ref }}
+        run: |
+          set -euo pipefail
+          # Exclude patterns: binary files + lockfiles that tend to produce noise.
+          EXCLUDE='(bun\.lock|package-lock\.json|\.png$|\.jpg$|\.jpeg$|\.gif$|\.ico$|\.woff2?$|\.ttf$|\.eot$|\.svg$|\.pdf$)'
```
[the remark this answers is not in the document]
FranDias: Fixed in cf14747. Also ported upstream to nex-crm/.github#18 (fetch-changed-files composite + scan-secrets default) — verified against a scratch repo with `docs/bun.lock.md` as a decoy: it's now correctly kept, while a real `bun.lock` at root is still excluded.

.github/workflows/secret-scan.yml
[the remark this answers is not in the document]
FranDias: Fixed in cf14747 — switched to NUL-delimited end-to-end (`git diff -z` → file → `xargs -0 -r -a`). Same fix also landed on the upstream composite in nex-crm/.github#18 (which refactors scan-secrets.yml to use the composite). Verified locally: `src/has space.ts` now survives as a single arg to xargs instead of being split.

.github/workflows/secret-scan.yml
```diff
@@ -0,0 +1,99 @@
...
+
+          if [ "$EVENT_NAME" = "pull_request" ]; then
+            files=$(git diff --name-only --diff-filter=ACMR "origin/${PR_BASE_REF}...HEAD")
+          elif git rev-parse --verify HEAD~1 >/dev/null 2>&1; then
+            files=$(git diff --name-only --diff-filter=ACMR HEAD~1..HEAD)
+          else
+            files=$(git ls-files)
+          fi
```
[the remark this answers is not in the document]
FranDias: Acknowledged, keeping the `git ls-files` fallback as-is. wuphf's lefthook `no-secrets` grep has been running on every commit since the repo started, so a first-push-to-main scan-all here is unlikely to surface a secret the local hook already passed. If it does, that's exactly the finding we want out of the whole exercise. Not changing the behavior in this PR — filing as a future `continue-on-error` gate if it ever bites us.

cmd/wuphf/import.go
FranDias: Added a short contextual comment above the suppression in cf14747 explaining when the code path fires (external orchestrator import only, not production). Saves the next contributor from the exact "are these creds meaningful" detour.

PR #191, opened 2026-04-21
fix: tmux command-too-long + boot-error UX hardening (v0.0.6.0)

najmuzzaman-mohammad:
## Summary

Fixes the tmux `command too long` regression from PR #139 that was silently dropping every office into the headless `claude --print` extra-usage quota. Also hardens related UX so future failures surface the right signal instead of a generic one.

**Root cause.** `claudeCommand` inlined the full per-agent system prompt (5–10 KB) as a single argument to `tmux split-window`. Tmux's internal command-parse buffer rejected oversize args before `sh -c` ever ran. Pane spawn always failed, fallback always fired.

**Changes:**

- **`internal/team/launcher.go`** — `claudeCommand` writes the prompt to `$TMPDIR/wuphf-prompt-<slug>.txt` (0o600) and passes `--append-system-prompt-file <path>`. Signature changes to `(string, error)` so 5 call sites can bubble write failures into a clean headless fallback instead of silently launching an agent with no system prompt. `reportPaneFallback` now takes `tmuxInstalled bool` and distinguishes "tmux not installed → install it" from "tmux installed but rejected → file a bug". `Launcher.Kill()` now calls `cleanupAgentTempFiles()` so broker tokens and prompts don't linger in `$TMPDIR`.
- **`cmd/wuphf/path_shadow.go`** — detector ignores candidates that resolve into the running binary's own directory. Fixes the npm install's false positive where `/opt/homebrew/bin/wuphf` (a symlink to `wuphf.js`) appeared to shadow the native `wuphf` binary in the same `node_modules/wuphf/bin` dir.
- **`web/index.html`** — boot-error watchdog hardened: first overlay wins (no more destructive replacement of specific errors by the generic "Boot timeout"), any fatal flips `bootDone` so the watchdog is a no-op once a diagnostic is on screen, and when the watchdog does fire its detail now includes `document.readyState`, `location.hash`, and which `/assets/index-*` bundles finished downloading.
- **`web/src/main.tsx`** — catch branch signals `window.__wuphfBootDone?.()` so a React mount failure isn't duplicated by the 10s watchdog 10 seconds later.

## Test Coverage

```
NEW CODE PATHS                                       TESTS
[MOD] claudeCommand()
  ├── uses --append-system-prompt-file               [★★★]
  ├── stays under 4 KB tmux limit                    [★★★] regression
  ├── writes prompt file with correct content        [★★★]
  └── returns error when write fails                 [★★★]
[NEW] writeAgentPromptFile()                         [★★★] (via above)
[NEW] cleanupAgentTempFiles()                        [★★★]
[NEW] paneFallbackMessages(tmuxInstalled)            [★★★] both branches
[MOD] detectPathShadows(npm sibling case)            [★★★]
[MOD] main.tsx catch branch                          [★★]  spec-by-example
[MOD] index.html watchdog                            [★★]  spec-by-example
```

Coverage gate: PASS (~85%, above 80% target).
Tests: 51 → 52 web test files (+186 LOC vitest), +253 LOC Go test file.

Legend: ★★★ behavior + edge + error | ★★ happy path. The web watchdog is inline HTML JS (can't be imported), so its contract is pinned via spec-by-example in `web/src/main.test.tsx`.

## Pre-Landing Review

No blocking issues. Full `/plan-eng-review` ran on initial scope; all findings addressed before first commit.

## Test Results

- Go: all 27 packages pass (`go test ./...`, ~100s internal/team)
- Web: 51 files, 232 tests pass (vitest, 18s)
- Build: clean

## Versions bumped

- `VERSION`: 0.0.5.3 → 0.0.6.0
- `package.json` + `web/package.json`: 0.0.6.0

## Why users care

- **Subscription quota restored.** Office agents ship in real tmux panes again instead of consuming the Anthropic extra-usage quota every turn.
- **Boot-timeout "heisenbug" eliminated.** The CPU starvation from 4+ parallel `claude --print` spawns at launch — which was likely causing the user-reported web UI "Boot timeout" — goes away when pane-backed mode works.
- **Better diagnostics when anything else goes wrong.** Boot-error overlay now carries `readyState`, `hash`, and bundle-load status so a real boot failure can be debugged, not guessed at.
- **No more false PATH shadow warning** on every `npx wuphf`.

## Test plan

- [x] Backend unit + regression tests green
- [x] Web unit tests green (232/232)
- [ ] Manual: fresh `npx wuphf` launch with this build — confirm interactive pane mode activates (no "pane-backed mode unavailable" stderr)
- [ ] Manual: prod web UI boot — no overlays


FranDias (approved):

PR #197, opened 2026-04-21
feat(entity): cross-entity graph — briefs auto-link related entities

najmuzzaman-mohammad:
## Summary
- Parses `[[kind/slug]]` and bare `[[slug]]` wikilinks out of every fact at append time and appends to an append-only adjacency log at `team/entities/.graph.jsonl`. Reader coalesces duplicate edges.
- Extends the brief synthesizer to deterministically append a trailing `## Related` section generated from the graph (top 10 by recency). LLM prompt tells the model not to write that section manually; any Related block it emits is stripped before ours is appended.
- New `GET /entity/graph?kind=&slug=&direction={in,out,both}` endpoint powers a minimal `<EntityRelatedPanel/>` that mounts next to `FactsOnFile` on wiki entity pages (out-edges only in v1, capped at 5, refetches on `entity:fact_recorded`).
- New `entity_graph_query` MCP tool so agents can ask "who's connected to X" without reading the raw jsonl.
- Writes ride the existing WikiWorker via a new `IsEntityGraph` discriminator — same single-writer invariant as `IsEntityFact`, no wiki-index regen.

## Out of scope (per roadmap)
- Fuzzy name resolution (exact-match + wikilinks only).
- Graph visualisation (a list is fine for v1).
- Auto-contradiction detection across edges.

## Test plan
- [x] `go build ./...`
- [x] `go test ./... -count=1` (all team + teammcp packages green, including new `entity_graph_test.go` and `broker_entity_graph_test.go`)
- [x] `go test ./internal/team/ -race -run 'EntityGraph|ExtractRefs|RecordFactRefs|StripRelated|RenderRelated'`
- [x] `go vet ./...`
- [x] `cd web && bun run build`
- [x] `cd web && bun run test` (vitest, 241 passed — includes new `EntityRelatedPanel.test.tsx`)
- [ ] Manual: record a fact like `Works at [[companies/acme]]` on `people/sarah` and confirm the Related panel shows `companies/acme` after the next synthesis.

Work worktree: `.worktrees/cross-entity-graph`. Non-overlap respected: no edits to migration, scanner, human-auth, or WikiEditor files; additive-only changes to `broker.go` and `wiki_worker.go`.

PR #200, opened 2026-04-21
fix(entity-graph): post-review hardening for PR #197

najmuzzaman-mohammad:
## Summary

Follow-ups from the multi-perspective review after #197 landed.

### 🔴 HIGH — `stripRelatedSection` silent data loss

**Before:** first `## Related` in the body truncated everything from that point to EOF. A pathological LLM response that emitted the heading mid-document, inside a code fence, or in a "Related projects" section silently dropped the brief body.

**Now:** the managed block is wrapped in sentinels emitted by `renderRelatedSection`:

```
<!-- wuphf:related:start -->
## Related

- [[companies/acme]]
<!-- wuphf:related:end -->
```

`stripRelatedSection` matches strictly between the sentinels (strict path). For legacy pre-sentinel briefs and LLM-injected sections, it falls back to a narrow shape match: the heading must be the LAST top-level heading, AND everything after it must be bullet items or blank lines. Any prose disqualifies the match.

### 🔴 HIGH — Latent deadlock in `RecordFactRefs`

**Before:** `g.mu` held across the blocking `EnqueueEntityGraph` call (channel send + up-to-`wikiWriteTimeout` reply wait). No current reader takes `g.mu`, so it was latent — but any future read path that acquired the same mutex would deadlock.

**Now:** build `buf` under the lock, release, then enqueue.

### 🟡 MEDIUM — `readAll()` partial-slice on scanner error

Was logging the error and returning accumulated edges with `nil` error. Now propagates the error so `Coalesce`/`Query` can surface a truncated read instead of silently returning stale results.

### 🟡 MEDIUM — Missing test coverage

- **`TestHandleEntityFact_GraphRecordFailureDoesNotBreakFactWrite`** — locks in the "graph is additive, never a constraint on the fact log" contract. Adds a `graphRecordFactRefs` test seam; overrides it with an error-returning stub; asserts POST `/entity/fact` still returns 200 with a valid fact envelope.
- **`<EntityRelatedPanel>` SSE refetch test** — captures the `subscribeEntityEvents` callback, fires it, asserts `fetchEntityGraph` is called again and the new edge renders. The SSE wiring was structurally present but behaviorally unverified (every existing test stubbed the subscriber to no-op).

## Verification

- `go build ./...` clean
- `go vet ./...` clean
- `go test ./internal/team -run "StripRelated|RecordFactRefs|EntityGraph|HandleEntityFact"` green
- `cd web && npx vitest run` — 259/259 across 53 files (5/5 on EntityRelatedPanel)
- One pre-existing race flake in `TestHeadlessTurnCompletedDurablyRejectsCodingTurnWithoutTaskStateOrEvidence` reproduces on untouched main; unrelated.

### DIFF
--- a/internal/team/broker_entity.go
+++ b/internal/team/broker_entity.go
@@ -30,6 +30,14 @@ import (
 	"time"
 )
 
+// graphRecordFactRefs is the test seam for the cross-entity graph hook in
+// handleEntityFact. Production code calls graph.RecordFactRefs; tests
+// override this var to inject errors and verify the
+// "graph failure keeps fact write intact" contract.
+var graphRecordFactRefs = func(ctx context.Context, graph *EntityGraph, fact Fact) ([]EntityRef, error) {
+	return graph.RecordFactRefs(ctx, fact)
+}
+
 // SubscribeEntityBriefEvents returns a channel of brief-synthesized events
 // plus an unsubscribe func.
 func (b *Broker) SubscribeEntityBriefEvents(buffer int) (<-chan EntityBriefSynthesizedEvent, func()) {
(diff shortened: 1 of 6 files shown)


FranDias (approved):

PR #194, opened 2026-04-21
fix(build): embed VERSION so local builds stop reporting v0.1.0

najmuzzaman-mohammad:
## Summary

- Bare `go build ./cmd/wuphf` reported `wuphf v0.1.0` no matter what, because `internal/buildinfo.Version` was a hardcoded default and no build path (README, DEVELOPMENT.md, FORKING.md, `.github/workflows/ci.yml`, `.goreleaser.yml`, uat scripts) injected a real value via `-ldflags`.
- Move `VERSION` into `internal/buildinfo/` and embed it, so contributor builds fall back to the file's contents instead of a stale literal. Root `VERSION` is now a symlink into the package so `cat VERSION` still works.
- Keep the `-ldflags -X` path working: `Version` / `BuildTimestamp` start empty and `Current()` falls back to the embed only when `-X` wasn't set. (First attempt initialized at package scope and silently clobbered `-X`.)
- Wire `.goreleaser.yml` ldflags to `{{.Version}}` / `{{.Date}}` so tagged releases report the actual git tag, not the embedded fallback.
- Bump `VERSION` from the stale `0.0.3.0` to `0.47.4-dev` (latest tag is `v0.47.4`).

## Test plan

- [x] `go build -o wuphf ./cmd/wuphf && ./wuphf --version` → `wuphf v0.47.4-dev` (was `v0.1.0`)
- [x] `go build -ldflags "-X '.../buildinfo.Version=0.47.5'" -o wuphf ./cmd/wuphf && ./wuphf --version` → `wuphf v0.47.5` (goreleaser path still overrides)
- [x] `go build ./...` clean
- [x] `go test ./internal/team/...` passes (broker tests read `buildinfo.Current()`)
- [ ] CI green on this PR
- [ ] After merge + next auto-release, a tagged release binary reports the real tag instead of the embedded fallback

PR #193, opened 2026-04-21
feat: compounding intelligence loop — playbook executions synthesize back into the playbook

najmuzzaman-mohammad:
## Summary

Closes the loop on v1.3 playbooks. Today executions accumulate in the append-only jsonl log but the playbook body itself never changes — the next agent to run the playbook gets the original v1 instructions with zero accumulated wisdom. This PR adds a broker-level synthesizer that mirrors how v1.2 entity briefs work: threshold-triggered LLM synthesis rewrites the source playbook's trailing \"## What we've learned\" section, preserving the author's main body verbatim. The existing auto-recompile hook then regenerates the SKILL.md so the next agent starts smarter.

## What changed

**Backend — `internal/team/`:**
- `playbook_synthesizer.go`: new broker-level worker. Same debounce + coalesce + inflight-follow-up pattern as `entity_synthesizer.go`. Threshold env `WUPHF_PLAYBOOK_SYNTHESIS_THRESHOLD` (default 3). Timeout `WUPHF_PLAYBOOK_SYNTHESIS_TIMEOUT` (default 45s).
- `mergePreservingAuthorBody`: defense in depth — even if the LLM ignores the prompt and rewrites the body, we splice only its \"## What we've learned\" section onto the original author body.
- Commits as `archivist` via the existing wiki write queue. The auto-recompile hook in `wiki_worker.go` then regenerates `SKILL.md` with the updated body — no new plumbing needed.
- Broker wiring in `broker.go` + `broker_playbook.go`: ensure/Start/Stop lifecycle, POST `/playbook/synthesize`, GET `/playbook/synthesis-status`, `playbook:synthesized` SSE event.
- `OnExecutionRecorded(slug)` hook called from `handlePlaybookExecution` — threshold auto-trigger without a separate background scanner.

**MCP tool — `internal/teammcp/playbook_tools.go`:**
- `playbook_synthesize_now` — gated on `WUPHF_MEMORY_BACKEND=markdown`. Lets an agent force synthesis after a particularly useful outcome.

**Frontend — `web/src/`:**
- `api/playbook.ts`: `synthesizeNow`, `fetchSynthesisStatus`, `subscribePlaybookSynthesizedEvents`.
- `components/wiki/PlaybookExecutionLog.tsx`: \"Last synthesis: 2h ago · N executions\" badge + \"Re-synthesize\" button with pending/success/error states. Auto-refreshes on SSE.
- `styles/wiki.css`: footer styling uses existing wiki tokens, no new palette.

## Design decisions

[...]

PR #192, opened 2026-04-21
feat: human-editable wiki — in-browser editor on /wiki/* articles

najmuzzaman-mohammad:
## Summary

Adds a Wikipedia-style **Edit source** flow to every `/wiki/*` article so the founder can draft and edit the team encyclopedia directly in the browser — no agent middleman, no terminal editor.

- New broker endpoint `POST /wiki/write-human` with **optimistic SHA-based concurrency** (409 + current bytes on conflict).
- Fixed `Human <human@wuphf.local>` git identity, mirroring the `ArchivistAuthor` pattern; reuses the single-writer `WikiWorker` queue.
- Web: **Edit source** tab on articles → plain-markdown textarea + commit message + save. Conflict banner offers one-click reload. Catalog gets a **+ New article** modal with local path validation. `Byline` shows a distinct "Human" pill when `last_edited_by === "human"`.

## What changed

**Backend**
- `internal/team/human_commit.go` — `HumanAuthor` constant, `ErrWikiSHAMismatch`, `Repo.CommitHuman` (pre-check → write → index regen → commit, atomic like `Repo.Commit`).
- `internal/team/wiki_worker.go` — new `IsHuman` + `ExpectedSHA` on `wikiWriteRequest`, `EnqueueHuman()`, `handleWikiWriteHuman` HTTP handler returning 409 with `current_sha` + `current_content`.
- `internal/team/wiki_article.go` — `ArticleMeta.CommitSHA` so the editor can round-trip expected_sha.
- `internal/team/broker.go` — register `/wiki/write-human` on the authed mux (HTTP-only; MCP never reaches it).

**Frontend**
- `web/src/api/wiki.ts` — `writeHumanArticle()` helper + `WriteHumanConflict` envelope; `WikiArticle.commit_sha` surfaced to consumers.
- `web/src/components/wiki/WikiEditor.tsx` — textarea + edit-summary + save/cancel + 409 banner.
- `web/src/components/wiki/NewArticleModal.tsx` — section + slug + title; mirrors backend `validateArticlePath` shape so users hear bad input before an HTTP round-trip.
- `web/src/components/wiki/HatBar.tsx` — new "Edit source" tab between Talk and History.
- `web/src/components/wiki/WikiArticle.tsx` — renders the editor on `tab === 'edit'`, refetches on save, returns to Article view.
- `web/src/components/wiki/Byline.tsx` — amber "Human" pill for `authorSlug === 'human'`.
[...]

PR #201, opened 2026-04-21
fix(team): wait for wiki worker side goroutines before tempdir cleanup

najmuzzaman-mohammad:
## Summary

The `test` job has been red on main since #192 and #193 landed. Symptom on Linux CI:

\`\`\`
testing.go:1369: TempDir RemoveAll cleanup:
  unlinkat /tmp/<Test>/.../wiki.bak/.git: directory not empty
\`\`\`

Ten-plus tests (\`TestCommitHumanHappyPath\`, \`TestNotebookWriteHappyPath\`, \`TestWikiSearchReturnsHits\`, \`TestFactLog_MalformedLineRecovery\`, \`TestSynthesizer_StopPreventsNewJobs\`, \`TestPlaybookSynthesizer_StopPreventsNewJobs\`, etc.) fail this way on every CI run. They pass locally on macOS because the timing window is wider.

Root cause: \`WikiWorker\` spawned a detached backup-mirror goroutine (\`maybeScheduleBackup\`), and test fixtures used \`defer cancel()\` without waiting for drain. Drain exited on \`ctx.Done\`, but the backup goroutine — and any in-flight drain write — could still be touching the repo while \`t.TempDir\` RemoveAll ran.

## Fix

1. Track the backup mirror goroutine in \`sideGoroutines\` (it was fire-and-forget; auto-recompile was already tracked).
2. \`defer sideGoroutines.Wait()\` + \`defer close(drainDone)\` in drain so drain only returns once all detached helpers have finished, and tests can observe full shutdown.
3. Move the \`ReplyCh\` send in \`process()\` to a deferred send at the end of the function. Without this, a caller could return from \`Enqueue\` before \`sideGoroutines.Add(1)\` for backup/auto-recompile had happened, so \`WaitForIdle\` would see an incorrectly-low counter.

Expose \`Done()\` and \`WaitForIdle()\` so tests can pick the right synchronization point.

Update the five shared fixture teardowns (\`newStartedWorker\`, \`newStartedNotebookWorker\`, \`newFactLogFixture\`, \`newSynthFixture\`, \`newPlaybookSynthFixture\`) to wait on \`<-worker.Done()\` instead of \`worker.Stop()\`. \`Stop\` was unreliable because drain's own defer flips \`running → false\`, so \`Stop\`'s \`if !running.Swap(false)\` branch could short-circuit the \`sideGoroutines.Wait\`.

## Blocks

Unblocks PR #194 (embed VERSION fix), which is dead in the water on the same flake.

## Test plan

- [x] \`go test ./internal/team/...\` — pass (68s, no flake)
- [x] \`go test -count=5 -run 'TestCommitHuman...|TestNotebook...|TestWiki...|TestSynth...|TestFactLog...|TestPlaybookSynth...'\` — pass (5x in a row)
- [x] \`go test ./...\` — full repo pass
- [ ] CI green on this PR (Linux is where the flake reproduces)

### DIFF
--- a/internal/team/entity_facts_test.go
+++ b/internal/team/entity_facts_test.go
@@ -35,7 +35,7 @@ func newFactLogFixture(t *testing.T) (*FactLog, *WikiWorker, func()) {
 	log := NewFactLog(worker)
 	return log, worker, func() {
 		cancel()
-		worker.Stop()
+		<-worker.Done()
 	}
 }
 
--- a/internal/team/entity_synthesizer_test.go
+++ b/internal/team/entity_synthesizer_test.go
@@ -60,7 +60,7 @@ func newSynthFixture(t *testing.T, llmStub func(ctx context.Context, sys, user s
 	return synth, factLog, worker, pub, func() {
 		synth.Stop()
 		cancel()
-		worker.Stop()
+		<-worker.Done()
 	}
 }
 
--- a/internal/team/notebook_worker_test.go
+++ b/internal/team/notebook_worker_test.go
@@ -55,7 +55,7 @@ func newStartedNotebookWorker(t *testing.T) (*WikiWorker, *Repo, *recordingNoteb
 	worker.Start(ctx)
 	return worker, repo, pub, func() {
 		cancel()
-		worker.Stop()
+		<-worker.Done()
 	}
 }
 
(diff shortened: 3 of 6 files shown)


FranDias (approved):

PR #202, opened 2026-04-21
fix(tests): track orphaned goroutines so Stop() waits before TempDir cleanup

najmuzzaman-mohammad:
## Summary

Fixes the `TempDir RemoveAll cleanup: directory not empty` flake class that has been leaking across CI runs since #193 (compounding loop) and #197 (cross-entity graph) landed.

Three goroutines were spawned but never tracked on a WaitGroup, so `Stop()` returned while filesystem operations were still in flight.

## Changes

1. **`entity_synthesizer.go`** — the re-schedule goroutine in `runJob` (fires after a completed synthesis to pick up queued follow-ups) called `go func() { time.Sleep(10ms); EnqueueSynthesis(...) }` outside of `s.wg`. Now tracked on `s.wg` + exits early on `s.stopCh` so Stop() doesn't wait the full 10ms per orphan.
2. **`playbook_synthesizer.go`** — identical re-schedule pattern, identical fix.
3. **`wiki_worker.go`** — `maybeScheduleBackup` fires `go func()` outside of `sideGoroutines`. Added so Stop() waits for the mirror copy.

## Verification

`go test ./internal/team` full suite green in 70s — first clean run in a while.

No production behavior change.


FranDias (approved):

PR #206, opened 2026-04-21
fix(launcher): overflow pane spawn errors + codex-bound agents + MCP tool audit

najmuzzaman-mohammad:
## Why this regressed after blueprints

The overflow pane-spawn path has silently ignored \`tmux new-window\` errors since PR #139 (98983ae4). Pre-blueprint rosters fit within the 5 visible slots, so \`spawnOverflowAgents()\` was a no-op — the bug was invisible. Blueprints shipped 8-agent rosters (founding-team: ceo, pm, fe, be, ai, designer, cmo, cro); agents 6–8 now hit the overflow path, where silent \`tmux new-window\` failures became observable as:

- \`Agents: pane capture for <slug> stopped after 5 failures: tmux capture-pane: exit status 1\`
- Agents that "did work but reply never came"
- Missing "X is typing..." indicators

### 1. Overflow spawn captures errors and falls back
\`spawnOverflowAgents\` now uses \`CombinedOutput()\`, logs any tmux failure with detail, and records the slug in a new \`failedPaneSlugs\` map. \`agentPaneTargets()\` skips recorded slugs so the capture loop doesn't poll ghost targets. Same treatment for \`spawnVisibleAgents\`' secondary splits (pane 1 is still mandatory — a first-pane failure aborts).

### 2. Codex-bound agents never get Claude panes
\`trySpawnWebAgentPanes\` only checked the install-wide provider. In a mixed roster (install=claude-code, some members=codex) the codex-bound agents still received claude panes running the wrong binary. New \`paneEligibleOfficeMembers()\` filters them out of both visible and overflow selection upstream. \`agentPaneTargets()\` also skips them via \`skipPaneForSlug()\` as defense in depth.

### 3. MCP tool calls tee into the per-agent stream
Adds a \`ReceivingMiddleware\` in \`teammcp\` that wraps every \`tools/call\`. On invocation it POSTs \`{slug, phase=call, tool, args}\` to a new broker endpoint \`/agent-tool-event\`. On completion it POSTs \`phase=result\` (truncated summary) or \`phase=error\`. The broker appends events to \`AgentStream(slug)\`, which the existing web SSE consumer already displays. This gives the UI an audit trail of which tool each agent called, visible even when pane capture is disabled or fails.

## Test plan
- [x] \`go build ./...\`
- [x] \`go test ./internal/team/ ./internal/teammcp/ -count=1\` passes (pre-existing \`TestDetectRuntimeCapabilities\` cleanup flake unrelated)
- [ ] Dev smoke: founding-team blueprint → no "pane capture stopped after 5 failures" on stderr; CEO tool calls show in agent stream; designer (overflow slot) replies arrive in channel
- [ ] Mixed provider: bind one agent to codex → no Claude pane created for it; headless codex dispatch still works

## Summary by CodeRabbit

* **New Features**
  * Added a founding team roster option when starting a new organization from scratch, including preset roles (CEO, GTM Lead, Engineer, PM, Designer).
  * Added member status indicators displaying recent routing activity.
  * Implemented tool call event tracking and logging for agent operations.

* **Bug Fixes**
  * Improved resilience when agent initialization fails with automatic fallback mechanisms.
  * Enhanced error detection and recovery for agent processes.

### DIFF
--- a/internal/team/broker.go
+++ b/internal/team/broker.go
@@ -1272,6 +1272,7 @@ func (b *Broker) StartOnPort(port int) error {
 	mux.HandleFunc("/v1/logs", b.requireAuth(b.handleOTLPLogs))
 	mux.HandleFunc("/events", b.handleEvents)
 	mux.HandleFunc("/agent-stream/", b.requireAuth(b.handleAgentStream))
+	mux.HandleFunc("/agent-tool-event", b.requireAuth(b.handleAgentToolEvent))
 	mux.HandleFunc("/web-token", b.handleWebToken)
 	// Onboarding: state/progress/complete + prereqs/templates/validate-key + checklist.
 	// completeFn posts the first task as a human message and seeds the team.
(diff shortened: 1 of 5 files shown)


FranDias (approved):

PR #210, opened 2026-04-22
fix(ui): quote-reply on inline thread replies opens thread panel pre-quoted

najmuzzaman-mohammad:
## Summary

PR #209 landed quote-reply *inside* the thread panel but missed the sibling case: **inline thread replies in the main channel feed** had hover actions for Reply-in-thread and Copy-link only, with no quote affordance. You saw a reply under a parent, hovered it, and had no way to quote it.

This wires a third `[Quote-reply]` button into the inline-reply hover toolbar. Clicking it:

1. Stores the reply id in a new `pendingQuoteId` slot on the app store.
2. Opens the thread panel on the reply's parent.
3. The panel's new `useEffect` consumes `pendingQuoteId` once its `replies` array contains the target — seeds its local `quoting` state, focuses the composer, clears the pending marker.

No broker changes, no API changes, no store schema break. The parent hover toolbar intentionally keeps Reply-in-thread only — quoting the parent is the default behavior inside the panel, so adding a button for it would be redundant.

## Files

| File | Change |
|---|---|
| `web/src/stores/app.ts` | `pendingQuoteId` + setter |
| `web/src/components/messages/MessageFeed.tsx` | `openThreadWithQuote(parentId, quoteId)` helper, passed as `onQuoteReply` on reply bubbles |
| `web/src/components/messages/ThreadPanel.tsx` | consume `pendingQuoteId` in a second effect, don't let the `activeThreadId` effect clobber it |

## Test plan

Verified in the dev build (`wuphf-dev` on :7900):

- [x] Hover an inline thread reply → toolbar shows `[Reply in thread] [Quote-reply] [Copy link]`
- [x] Click Quote-reply → thread panel opens, chip reads "Replying to @ceo We're alive…", composer placeholder = "Reply to @ceo…"
- [x] Click Reply-in-thread on the same reply → panel opens with NO quote chip, placeholder = "Reply to thread…"
- [x] Quote-reply inside the thread panel (from PR #209) still works
- [x] `tsc -b` clean, `vite build` clean

## Summary by CodeRabbit

* **New Features**
  * Improved quote reply interactions: Selecting "quote reply" on a message now automatically opens the parent thread and pre-selects the quoted message in the composer for streamlined conversation flows.

### DIFF
--- a/web/src/components/messages/MessageFeed.tsx
+++ b/web/src/components/messages/MessageFeed.tsx
@@ -27,10 +27,19 @@ type FeedElement =
 export function MessageFeed() {
   const currentChannel = useAppStore((s) => s.currentChannel)
   const setActiveThreadId = useAppStore((s) => s.setActiveThreadId)
+  const setPendingQuoteId = useAppStore((s) => s.setPendingQuoteId)
   const collapsedThreads = useAppStore((s) => s.collapsedThreads)
   const toggleThreadCollapsed = useAppStore((s) => s.toggleThreadCollapsed)
   const containerRef = useRef<HTMLDivElement>(null)
   const prevLengthRef = useRef(0)
+
+  // Quoting an inline reply pops the thread panel open on the reply's parent
+  // and tells the panel which message to pre-quote. The panel reads
+  // pendingQuoteId from the store and seeds its local `quoting` state.
+  const openThreadWithQuote = (parentId: string, quoteId: string) => {
+    setPendingQuoteId(quoteId)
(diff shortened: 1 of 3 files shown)


najmuzzaman-mohammad:
Closing — user clarified quote-reply should only exist inside the thread panel, not on main-feed inline replies. PR #209 already ships the in-panel version (verified: hover a reply inside the thread panel → [Quote-reply] button → click sets quoting, chip shows 'Replying to @ceo', composer placeholder updates to 'Reply to @ceo…').

PR #218, opened 2026-04-22
fix(team): @-tags and DMs route to specialist, not CEO

najmuzzaman-mohammad:
## Summary

Two bugs caused explicit @-tags and DMs to silently route to CEO instead of the targeted specialist. User-observed symptoms:
1. Tagging any non-CEO agent → CEO responds instead of the specialist.
2. CEO tagging a specialist → nobody responds.
3. DMs to specialists → dropped silently.

**Root causes (both in `internal/team/launcher.go`):**

1. **Channel-membership filter ran before the explicit @-tag check.** `notificationTargetsForMessage` applied the `enabledMembers` filter in `allowTarget` / `addImmediate` and in the focus-mode human-tag loop. A newly hired specialist is in `b.members` but not yet in `ch.Members` for `#general`, so the filter dropped them and CEO absorbed the message. The existing "explicit @-tag always wins" short-circuit was unreachable.
2. **`activeSessionMembers` was pack-gated.** It returned only agents listed in `l.pack.Agents`, so agents hired via the web wizard (`POST /office-members`) after launch were silently excluded from `agentNotificationTargets`. DMs and @-tags targeting wizard-hired agents dropped into the void.

**Fix:**
- `allowTarget` and `addImmediate` now compute `explicit := containsSlug(msg.Tagged, slug)` and bypass `enabledMembers` when true.
- Focus-mode human-tag loop simplified — the `isEnabled` check is redundant inside `for _, slug := range msg.Tagged`.
- `activeSessionMembers` now returns pack order first, then any broker-only members appended after.

## Test Coverage

9 regression tests in `internal/team/mention_routing_bug_test.go` covering both bugs at target-computation and full-dispatch layers:

```
Layer A (target computation, no broker):
  TestBug_HumanTagsSpecialist_CollaborativeMode_SpecialistIsImmediate
  TestBug_CEOTagsSpecialist_CollaborativeMode_SpecialistIsImmediate
  TestBug_HumanDMsSpecialist_CollaborativeMode_SpecialistIsImmediate

Layer B (full dispatch via deliverMessageNotification → headless queue):
  TestBug_HumanTagsSpecialist_Dispatch_SpecialistReceivesTurn
  TestBug_CEOTagsSpecialist_Dispatch_SpecialistReceivesTurn
  TestBug_HumanDMsSpecialist_Dispatch_SpecialistReceivesTurn

Wizard-hired agent scenarios (reproduce symptom 3 exactly as user reported):
  TestBug_DMToWizardHiredPM_Dispatch
  TestBug_TagWizardHiredPM_InGeneral_Dispatch

Root-cause assertion:
  TestBug_RootCause_ChannelMembershipFilterDropsExplicitMention
```

All 9 tests fail on `origin/main` with diagnostics pinpointing each root cause. All pass with the fix. Full `go test ./internal/...` passes (team suite: 85s).

## Pre-Landing Review

48 lines changed across one file, plus a test file. No SQL, no LLM input, no new error paths. The `seen` map in `activeSessionMembers` guards against duplicate slugs. Comments explain the why (user-observed bug, invisible silent drop) rather than the what. Clean.

## Verification

Dev build on isolated ports (7899 broker / 7900 web, `HOME=~/.wuphf-dev-home`), tested visually:

- **Tag in #general:** `@pm please scope the new onboarding flow` — PM's headless worker dispatched, entry appears in `~/.wuphf-dev-home/.wuphf/logs/headless-codex-pm.log`. CEO remained idle ("waiting for work").
- **DM to PM:** `Browser DM test — please acknowledge.` — PM's Claude worker executed (live-output panel showed the `RATE_LIMIT_EVENT` payload, which is a real Claude API response, i.e. PM's turn really fired). CEO not dispatched.

Pre-fix, neither path would ever land in PM's log — the notification was silently absorbed by CEO.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/...` passes (all packages, 85s)
- [x] 9 new regression tests fail on main, pass with fix
- [x] Visual browser test: @-tag in #general wakes PM (not CEO)
- [x] Visual browser test: DM to PM wakes PM (not CEO)
- [x] Full team suite passes with no new regressions

## Release 0.0.6.1

* **Bug Fixes**
  * Fixed @-mention routing in #general channel to deliver messages to intended recipients
  * Fixed direct message and explicit @-mention targeting to reach recently added specialists
  * Ensured muted members remain unreachable even when explicitly mentioned


FranDias (approved):

PR #223, opened 2026-04-22
fix(team): auto-promote @slug text for humans; silence tmux no-server noise

najmuzzaman-mohammad:
## Summary

Follow-up to #218. Two symptoms reported against v0.0.6.1:

**Symptom 1** — user typed `@pm do you know of Lenny's PM fit frameworks?` in #general. The UI did not render `@pm` as a tag chip while typing. Message posted. CEO received it and re-tagged PM in its own reply. PM itself was never woken.

**Root cause** — the web composer does not always commit typed `@slug` text into an explicit tag chip, so the POST body arrives with \`tagged: []\`. The broker's auto-promote block (\`broker.go:7072\`) ran only for agent senders — the original comment was that humans might type \`@\` conversationally. In practice every human \`@agent\` is intended as a tag, and #218's routing fix never sees \`msg.Tagged\` containing \`pm\`, so it falls through to \`addImmediate(lead)\` → CEO.

**Fix** — auto-promote for every sender except \`system\`. \`extractMentionedSlugs\` already restricts to registered agent slugs (\`b.findMemberLocked(slug) != nil\`), so \`@joedoe\` (not an agent) stays plain text.

**Symptom 2** — console spammed:
\`\`\`
office_reseeded: respawn panes failed: spawn first agent: exit status 1 (tmux: no server running on /private/tmp/tmux-501/wuphf)
\`\`\`
on every onboarding reseed. Web/headless mode never has a tmux server; attaching to it is expected to fail and the headless dispatch path handles delivery silently. Logging the normal code path as a recurring error misled debugging.

**Fix** — \`respawnPanesAfterReseed\` silences \`isNoSessionError\`. Permission denied, exec-not-found, and other real failures keep logging.

## Test Coverage

5 regression tests in \`internal/team/mention_auto_promote_test.go\`:

\`\`\`
TestAutoPromote_HumanTypedAtPM_PromotesToTagged        (the user's bug)
TestAutoPromote_HumanTypedAtNonAgent_LeavesUntagged    (conversational @ stays untouched)
TestAutoPromote_AgentTypedAtPM_StillWorks              (no regression in pre-existing behavior)
TestAutoPromote_ExplicitTagRespected                   (composer-committed tag chip, no dup)
TestRespawnPanesAfterReseed_NoSessionErrorSilenced     (table-driven error classification)
\`\`\`

All pass. Full \`go test ./internal/team/\` green except \`TestPostHeadlessFinalMessageIfSilentPostsFinalOutput\` which is pre-existing on \`origin/main\` (verified via \`git stash\` bisect).

## Test plan

- [x] 5 new tests pass
- [x] Full team suite green (ignoring pre-existing flake)
- [x] CHANGELOG + VERSION + VHS snapshots updated for v0.0.6.2

## Release Notes v0.0.6.2

* **Bug Fixes**
  * @pm mentions from human authors are now properly recognized and tagged in messages
  * Expected tmux server errors no longer generate failure logs in web/headless environments


FranDias (dismissed):

FranDias (approved):

PR #224, opened 2026-04-22
fix(onboarding): restore Skip on first-task step; unblock readiness CTA

najmuzzaman-mohammad:
## Summary

- Restores the **Skip for now** button on the first-task step (`What should the team work on first?`) where the optional question is asked. It had been moved to the readiness summary, so the optional nature wasn't visible at the point of the question.
- Skip now advances to the readiness review (clearing the task field) instead of finishing onboarding directly to the workspace, so users still see the configuration summary before launching.
- On the readiness step, the primary CTA (**Get started**) is no longer disabled when the task field is empty. With an empty task it auto-routes through the skip path, so an optional field never blocks completion of onboarding.
- Net diff: 1 file, 26 insertions, 14 deletions in `web/src/components/onboarding/Wizard.tsx`.

## Test plan

- [ ] Reach the `What should the team work on first?` step in the wizard.
- [ ] Click **Skip for now** → readiness summary opens with no task carried over.
- [ ] On the readiness step with an empty task, **Get started** is enabled and completes onboarding.
- [ ] Type a task → click **Review setup** → **Get started** posts the task.
- [ ] Type a task → go **Back** → click **Skip for now** → readiness opens with empty task.

### DIFF
--- a/web/src/components/onboarding/Wizard.tsx
+++ b/web/src/components/onboarding/Wizard.tsx
@@ -1039,7 +1039,9 @@ interface TaskStepProps {
   taskText: string
   onChangeTaskText: (v: string) => void
   onNext: () => void
+  onSkip: () => void
   onBack: () => void
+  submitting: boolean
 }
 
 function TaskStep({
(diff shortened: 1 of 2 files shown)


FranDias (approved):

PR #226, opened 2026-04-22
ci: add /approve comment workflow

FranDias:
## Summary
Adds `.github/workflows/approve-on-comment.yml` so org members can approve this repo's PRs by commenting `/approve <reason>`.

Calls reusable workflow `nex-crm/.github/.github/workflows/approve-on-comment.yml@main`.

## Why
Part of the org-wide rollout to satisfy the Sprinto "Branch protection rules should be enforced for admins" control. Validated end-to-end on nex-crm/intelligence (App token, org-membership check, self-approval guard all confirmed working).

## Rollout order
1. This PR merges
2. Branch protection with 1 required review enabled on `main` (separate PR)
3. Team can approve via `/approve <reason>` comment on any PR authored by a different org member

## Summary by CodeRabbit

* **New Features**
  * Added automated pull request approvals. Organization members can now comment `/approve` on a pull request to trigger an automatic approval review, with optional reasoning included in the approval.

### DIFF
--- a/.github/workflows/approve-on-comment.yml
+++ b/.github/workflows/approve-on-comment.yml
@@ -1,74 +1,11 @@
 name: Approve on Comment
 
-# Inlined (not using private nex-crm/.github reusable) because this is a
-# public repo and public repos cannot consume private reusable workflows.
-# Keep in sync with nex-crm/.github/.github/workflows/approve-on-comment.yml
-
 on:
   issue_comment:
     types: [created]
 
 jobs:
   approve:
-    if: |
-      github.event.issue.pull_request != null &&
-      startsWith(github.event.comment.body, '/approve')
-    runs-on: ubuntu-latest
-    timeout-minutes: 5
-    permissions:
-      pull-requests: write
-      issues: write
-    steps:
-      - name: Mint GitHub App token
-        id: app-token
-        uses: actions/create-github-app-token@v1
-        with:
-          app-id: ${{ vars.APPROVER_APP_ID }}
-          private-key: ${{ secrets.APPROVER_APP_PRIVATE_KEY }}
-
-      - name: Verify commenter is org member
-        env:
-          GH_TOKEN: ${{ steps.app-token.outputs.token }}
-          COMMENTER: ${{ github.event.comment.user.login }}
-          ORG: ${{ github.repository_owner }}
-        run: |
-          if ! gh api "orgs/${ORG}/members/${COMMENTER}" --silent 2>/dev/null; then
-            echo "::error::@${COMMENTER} is not a member of ${ORG} — /approve denied"
-            exit 1
-          fi
-
-      - name: Reject self-approval
-        env:
-          COMMENTER: ${{ github.event.comment.user.login }}
-          PR_AUTHOR: ${{ github.event.issue.user.login }}
-        run: |
-          if [ "${COMMENTER}" = "${PR_AUTHOR}" ]; then
-            echo "::error::@${COMMENTER} cannot /approve their own PR"
-            exit 1
-          fi
-
-      - name: Submit approval review
-        env:
-          GH_TOKEN: ${{ steps.app-token.outputs.token }}
-          PR_NUMBER: ${{ github.event.issue.number }}
-          REPO: ${{ github.repository }}
-          COMMENTER: ${{ github.event.comment.user.login }}
-          BODY: ${{ github.event.comment.body }}
-        run: |
-          reason="${BODY#/approve}"
-          reason="${reason# }"
-          [ -z "$reason" ] && reason="(no reason provided)"
-          gh pr review "${PR_NUMBER}" \
-            --repo "${REPO}" \
-            --approve \
-            --body "Approved by @${COMMENTER} via \`/approve\`: ${reason}"
-
-      - name: React to comment
-        env:
-          GH_TOKEN: ${{ steps.app-token.outputs.token }}
-          COMMENT_ID: ${{ github.event.comment.id }}
-          REPO: ${{ github.repository }}
-        run: |
-          gh api -X POST \
-            "repos/${REPO}/issues/comments/${COMMENT_ID}/reactions" \
-            -f content=+1 >/dev/null
+    uses: nex-crm/.github/.github/workflows/approve-on-comment.yml@v1
+    secrets:
+      APPROVER_APP_PRIVATE_KEY: ${{ secrets.APPROVER_APP_PRIVATE_KEY }}
--- a/.github/workflows/approve-on-comment.yml
+++ b/.github/workflows/approve-on-comment.yml
@@ -1,11 +1,78 @@
 name: Approve on Comment
 
+# Inlined (not using private nex-crm/.github reusable) because this is a
+# public repo and public repos cannot consume private reusable workflows.
+# Keep in sync with nex-crm/.github/.github/workflows/approve-on-comment.yml
+# (current canonical SHA tracked via the v1 tag on that repo).
+
 on:
   issue_comment:
     types: [created]
 
 jobs:
   approve:
-    uses: nex-crm/.github/.github/workflows/approve-on-comment.yml@v1
-    secrets:
-      APPROVER_APP_PRIVATE_KEY: ${{ secrets.APPROVER_APP_PRIVATE_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/approve-on-comment.yml
```diff
@@ -0,0 +1,74 @@
...
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  approve:
+    if: |
+      github.event.issue.pull_request != null &&
+      startsWith(github.event.comment.body, '/approve')
```
FranDias: **IMPORTANT** — `startsWith(github.event.comment.body, '/approve')` is a loose prefix match, so `/approved`, `/approvers`, `/approve-later`, or even a typo like `/approvelgtm` all trigger the full workflow. In the best case the commenter is an org member and their off-topic comment produces an accidental approving review on the PR (and a `+1` reaction). On a PR with only one required review, that approval is enough to unblock merge.

This is the same issue CodeRabbit raised, and it's a real functional bug rather than a style nit — the command grammar should be anchored.

**Fix:** either tighten the expression gate, e.g.

```yaml
if: |
  github.event.issue.pull_request != null &&
  (github.event.comment.body == '/approve' ||
   startsWith(github.event.comment.body, '/approve '))
```

or keep the loose gate and add an explicit regex check as the first step that fails closed on anything not matching `^/approve($|[[:space:]])`. Whichever you pick, back-port to `nex-crm/.github/.github/workflows/approve-on-comment.yml` so the other 14 repos in the rollout inherit the fix automatically.

.github/workflows/approve-on-comment.yml
```diff
@@ -0,0 +1,74 @@
...
+          private-key: ${{ secrets.APPROVER_APP_PRIVATE_KEY }}
+
+      - name: Verify commenter is org member
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          COMMENTER: ${{ github.event.comment.user.login }}
+          ORG: ${{ github.repository_owner }}
+        run: |
+          if ! gh api "orgs/${ORG}/members/${COMMENTER}" --silent 2>/dev/null; then
+            echo "::error::@${COMMENTER} is not a member of ${ORG} — /approve denied"
```
FranDias: **IMPORTANT** — `gh api orgs/${ORG}/members/${COMMENTER}` returns 204 only when the authenticated identity can **see** that user's membership. With `GH_TOKEN` set to the App token, the App needs `Organization permissions → Members: Read` for this endpoint to return 204 for users whose org membership is set to private. If the App only has the default install scopes (or only `Members: None`), then any teammate with private org visibility will hit 404 here and be told they're "not a member of nex-crm" when they actually are.

You mentioned this was validated end-to-end on `nex-crm/intelligence` — worth confirming the validator's org membership is actually **public** (GitHub defaults to public on recent joiners but respects per-user overrides), otherwise the happy path hides this failure mode. Quick check: `gh api orgs/nex-crm/members/<teammate> --silent` using the App token against a teammate whose membership is private.

**Fix (if the App doesn't have Members:Read):** either grant the App that org permission, or switch to `gh api orgs/${ORG}/memberships/${COMMENTER}` which returns the membership object regardless of visibility (still requires the App to have org read, but behaves consistently). Worth documenting the required App permissions in `docs/APPROVER_BOT_SETUP.md` either way so future installs don't silently regress this.

.github/workflows/approve-on-comment.yml
```diff
@@ -0,0 +1,74 @@
...
+
+# Inlined (not using private nex-crm/.github reusable) because this is a
+# public repo and public repos cannot consume private reusable workflows.
+# Keep in sync with nex-crm/.github/.github/workflows/approve-on-comment.yml
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
```
FranDias, line 12: **SUGGESTION** — No `concurrency:` stanza on the job. If someone double-clicks `/approve`, or two org members race on the same PR, two jobs will run in parallel, both mint App tokens, both submit approving reviews, and both POST the `+1` reaction. GitHub de-dupes identical approving reviews from the same App actor (so there's no *actual* duplicate-approval risk), and the reaction endpoint is idempotent, so the blast radius is "two workflow runs in the Actions tab" rather than a correctness issue — hence suggestion, not blocking.

Clean version if you want it:

```yaml
jobs:
  approve:
    concurrency:
      group: approve-on-comment-${{ github.event.issue.number }}
      cancel-in-progress: false
```

`cancel-in-progress: false` is deliberate — cancelling mid-approval is worse than just queuing.

.github/workflows/approve-on-comment.yml
```diff
@@ -0,0 +1,74 @@
+name: Approve on Comment
+
+# Inlined (not using private nex-crm/.github reusable) because this is a
+# public repo and public repos cannot consume private reusable workflows.
+# Keep in sync with nex-crm/.github/.github/workflows/approve-on-comment.yml
```
FranDias: **SUGGESTION** — The "Keep in sync with nex-crm/.github/..." comment is the only thing preventing drift across the 15-repo rollout, and comments drift fastest. Consider a lightweight CI check in `nex-crm/.github` that diffs the `steps:` section of this inlined copy against the reusable and fails on drift — either via a dedicated workflow or by folding it into an existing scheduled job. Alternatively, stash a SHA/checksum of the reference file in a header comment here and have `secret-scan.yml` (or a new tiny workflow) assert it on every PR. Not blocking this PR — but with 15 copies, drift is a when-not-if.

FranDias (commented):
Reviewed the inlined workflow against the private reusable at `nex-crm/.github/.github/workflows/approve-on-comment.yml` — the bodies of every step (`if:` guard, permissions, env wiring, membership check, self-approval guard, reason parsing, review submission, reaction) are byte-identical. Only deltas are the header comment block and the trigger (`issue_comment` vs `workflow_call`), both expected for the public-repo inlining. Verified no collision with existing workflows (`auto-release`, `ci`, `pages`, `release`, `secret-scan`) — none use `issue_comment`, no name clash.

Secret handling is clean: `APPROVER_APP_PRIVATE_KEY` flows only through `create-github-app-token@v1` inputs, never into a shell `env:`. Command-injection surface via `BODY` is safe — it's passed through `env:` (GitHub's recommended pattern) and shell parameter expansion does not re-parse substituted values, so `/approve; rm -rf /` or `$(whoami)` land as literal text in the approval body. `set -e` + `pipefail` are on by default in Actions' bash; the `if ! gh api … --silent` pattern short-circuits `-e` correctly. `pull_request_target` is not needed — `issue_comment` always runs the base-branch workflow, which is the safer pattern for public repos.

Two moderate-severity items worth addressing in this PR (and back-porting to the reusable so the rest of the 15-repo rollout inherits the fix), plus two nits. Inline below.

FranDias:
## Review addressed

Going through findings from the initial review now that the rollout consolidation has landed:

- **`startsWith('/approve')` false-positive** (IMPORTANT) — tightened matcher to `body == '/approve' || startsWith(body, '/approve ')` in [nex-crm/.github#31](https://github.com/nex-crm/.github/pull/31). `/approved` / `/approvers` no longer trigger.
- **Org-member check visibility** (IMPORTANT) — the `nex-approver` App has `Organization Members: Read`, so private memberships resolve correctly. Verified against the installation permissions.
- **No `concurrency:` stanza** (SUGGESTION) — added in [#31](https://github.com/nex-crm/.github/pull/31). Group is `approve-on-comment-${issue.number}`, `cancel-in-progress: false` so an in-flight approval submit isn't killed mid-run.
- **Drift between inlined copies and canonical reusable** (SUGGESTION) — the 4 public-repo inlined copies were re-synced in dedicated follow-ups (wuphf#228 for this repo). Every one now byte-matches `nex-crm/.github@v1` except the trigger section, which differs by design.

Additional hardening landed since: `create-github-app-token` bumped from `@v1` to `@v3.1.1` in [#34](https://github.com/nex-crm/.github/pull/34); CRLF trailing-\\r bug fixed in [#33](https://github.com/nex-crm/.github/pull/33); self-approval explicitly allowed (accepted risk, documented in [`docs/AUDIT_EVIDENCE.md`](https://github.com/nex-crm/.github/blob/main/docs/AUDIT_EVIDENCE.md)).

PR #229, opened 2026-04-22
fix(prod): React loop, pane-capture recovery, stale-turn livelock

najmuzzaman-mohammad:
## Summary

User reported \`@pm\` stuck in "typing" with no reply on v0.61.7. Root-cause trace surfaced three compounding bugs:

1. **Web UI in an infinite \`setState\` loop.** \`useCommands\` returned \`toAutocomplete(data)\` on every call, producing a fresh array reference even when React Query's \`data\` was cached. Composer's Autocomplete has a \`useEffect\` watching \`items\` (derived from \`commands\`), so every render rebuilt items → re-fired the effect → set \`acItems\` → re-rendered Composer → looped. React eventually bailed with \`Maximum update depth exceeded\` and the UI became unresponsive. **Fix:** \`useMemo\` the derived array against \`[data, isError]\`.

2. **Pane capture gave up forever on office-reseed churn.** \`startPaneCaptureLoops\` spawned one goroutine per pane target; after 5 consecutive tmux \`capture-pane\` failures it logged \`stopped after 5 failures\` and exited permanently. Office reseeds routinely invalidate pane ids, so agents' live-output feeds went dark until the whole office restarted. **Fix:** after hitting the failure cap, sleep 30s, re-resolve the current pane target from the launcher (new \`resolvePaneTargetForSlug\` helper), and resume. Clear the prev-line cache so the re-surfaced pane pushes a fresh snapshot.

3. **Headless codex queue spun in a stale-cancel loop.** CEO's log showed dozens of \`stale-turn: cancelling active turn after 0s\` lines per session. Enqueue could preempt an active turn that had barely started, under clock skew or malformed-turn conditions. **Fix:** require both the configured stale threshold AND a 2s minimum-turn-age floor for cancellation. Breaks the tight loop without blocking legitimate preemption.

## Test Coverage

- **\`useCommands\`:** pinned deterministic output contract for \`toAutocomplete\` and documented the referential-stability requirement (the bug this PR fixes).
- **\`TestEnqueueHeadlessCodexTurnCancelsStaleTurn\`:** updated to override the new \`headlessCodexMinTurnAgeBeforeCancel\` constant so its 20 ms stale threshold still exercises the cancel path.
- Verified React loop gone by building locally (dev server, source-mapped) and reloading: 0 errors in 4s vs hundreds/sec before.

## Test plan

- [x] \`go test ./internal/team/\` passes (excluding two pre-existing flakes: \`TestPostHeadlessFinalMessageIfSilentPostsFinalOutput\` and \`TestPlaybookSynthesizer_IntegrationTriggersAutoRecompile\`, both verified passing in isolation)
- [x] \`vitest run src/hooks/useCommands.test.ts\` → 9 passed
- [x] Dev server reload: browser console shows 0 React errors (previously: infinite flood)
- [x] Snapshot drift: \`testdata/vhs/help.txt\` and \`version.txt\` updated to v0.0.6.3


FranDias (approved):

PR #231, opened 2026-04-22
fix(team): serialize pane dispatches so rapid @tags don't clobber each other

najmuzzaman-mohammad:
## Summary

User observed: tagging @pm twice in 3 minutes, PM answered one question and silently swallowed the other.

**Root cause:** \`sendNotificationToPane\` in \`internal/team/launcher.go\` runs \`/clear\` + type + Enter directly against tmux with no queueing. When two notifications arrive close together for the same agent, the second \`/clear\` wipes claude's in-flight input from the first before it was submitted. Claude sees whatever state survives the race and answers one question (at best).

**Fix:** per-slug dispatch queue.
- New \`queuePaneNotification(slug, paneTarget, notification)\` replaces direct \`sendNotificationToPane\` calls at both sites (\`sendChannelUpdate\` and \`sendTaskUpdate\`).
- One worker goroutine per slug drains its queue, inserting a 3-second minimum gap between \`/clear\` cycles so claude's TUI has time to return to prompt-ready.
- Different agents dispatch in parallel — the gap is per-slug, not a global fence.
- Worker exits atomically when queue is empty; next enqueue spawns a fresh goroutine.

\`paneDispatchMinGap\` is a \`var\` so the regression test can shorten it without changing prod behaviour. \`launcherSendNotificationToPane\` is a package-level seam for tests to observe dispatch ordering without shelling out to real tmux.

## Test Coverage

\`internal/team/pane_dispatch_queue_test.go\`:
- **SerializesPerSlug**: two rapid enqueues for one slug → the second dispatch fires at least \`paneDispatchMinGap\` after the first. Proves the race is closed.
- **DifferentSlugsRunInParallel**: two different slugs → both dispatches complete well inside one min-gap. Proves the gap isn't cross-slug (a slow pane for one agent wouldn't starve others).

## Test plan

- [x] Full \`go test ./internal/team/\` passes (one pre-existing flake, unrelated: \`TestPostHeadlessFinalMessageIfSilent\`)
- [x] Both new tests green in isolation and as part of the full suite


FranDias (dismissed):

FranDias (approved):

PR #232, opened 2026-04-22
feat(team): headless claude --print as default dispatch for web and TUI

najmuzzaman-mohammad:
## Summary

Makes headless \`claude --print\` per-turn the default dispatch path for both \`wuphf\` (web) and \`wuphf --tui\`. The legacy per-agent long-lived interactive tmux pane path is preserved as an internal fallback primitive but is no longer invoked at startup.

Motivation: Anthropic re-sanctioned headless CLI reuse in the 2026-04 OpenClaw policy note — it runs on the user's normal subscription quota with no separate extra-usage charge. Unifying dispatch across web and TUI matches how the Codex runtime already worked and removes tmux as a reliability-sensitive dependency in the common path. tmux is still required for \`--tui\` because the channel-view TUI itself lives in tmux; the web UI no longer needs it.

### Changes

- \`shouldUseHeadlessDispatch\` keys off \`paneBackedAgents\`, not \`webMode\` — TUI mode goes through the headless queue by default.
- \`Launch()\` (TUI) skips \`spawnVisibleAgents\` / \`primeVisibleAgents\` and calls \`resumeInFlightWork\` directly. The tmux session and channel-view pane still exist so the user has something to attach to.
- \`LaunchWeb\` calls \`resumeInFlightWork\` unconditionally and no longer invokes \`trySpawnWebAgentPanes\` at startup.
- \`trySpawnWebAgentPanes\` is retained as an internal fallback primitive (reachable if dispatch ever needs to promote to panes at runtime) with updated comments.
- \`resumeInFlightWork\` routing aligned with dispatch routing: keyed off \`!paneBackedAgents\` to prevent silently dropping resume packets when panes aren't up (caught via \`/review\`).
- \`enqueueHeadlessCodexTurnRecord\` lazily inits its three headless maps so test-constructed Launchers that skip \`NewLauncher\` don't panic when routed through headless.
- Durability-error path in \`runHeadlessCodexQueue\`: when the provider returns success but leaves no durable task state, block the task instead of retrying (the agent already had its turn).
- Messaging + docs updated to drop the stale "extra-usage quota" framing.

### Files

- \`internal/team/launcher.go\` — TUI + web dispatch defaults flipped, pane-spawn messaging rewritten.
- \`internal/team/headless_codex.go\` — map init, durability-error routing, \`isDurabilityError\` flag.
- \`internal/team/headless_codex_test.go\` — suppress lead-wake in queue-retry test to prevent interference.
- \`internal/team/resume.go\` — resume routing keyed off \`paneBackedAgents\`.
- \`internal/team/dispatch_test.go\` — test cases updated to match new defaults.
- \`internal/team/resume_test.go\` — regression test for TUI+claude resume routing.
- \`internal/team/launcher_tmux_fix_test.go\` — stale comment refresh.
- \`internal/config/config_test.go\` — trailing newline.
- \`README.md\` + \`CHANGELOG.md\` — documentation updates.

## Test plan

- [x] \`go build ./...\` clean.
- [x] \`go vet ./...\` clean.
- [x] \`go test ./internal/config/... ./internal/team/... ./cmd/wuphf\` green (modulo two pre-existing flakes: \`TestAutoPromote_EndToEnd_HumanTagsPM_DispatchesToPM\` under \`-race\` and \`TestPlaybookSynthesizer_IntegrationTriggersAutoRecompile\` under full-suite load — both verified pre-existing against pristine tree).
- [x] New regression test \`TestResumeInFlightWorkTUIClaudeRoutesHeadless\` pins the resume-gating invariant.
- [ ] \`wuphf\` (web) manual smoke on dev ports — boots without tmux, agents respond headlessly.
- [ ] \`wuphf --tui\` manual smoke — channel pane renders, agents respond headlessly, resume works after restart.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to WUPHF will be documented in this file.
 
+## [Unreleased]
+
+### Changed
+
+- **Headless `claude --print` is now the default dispatch path for both `wuphf` (web) and `wuphf --tui`.** Anthropic re-sanctioned headless CLI reuse in the 2026-04 OpenClaw policy note, and it runs on the user's normal subscription quota — no separate extra-usage charge. Every turn now dispatches as a fresh `claude --print` invocation, matching how the Codex runtime already worked and unifying dispatch across both modes. The previous per-agent long-lived interactive tmux pane path is preserved as an internal fallback primitive (reachable if dispatch ever needs to promote to panes at runtime) but is no longer invoked at startup. tmux is still required for `--tui` since the channel-view TUI runs in tmux; the web UI no longer needs it.
+- **Pane-fallback messaging updated to reflect the new default.** The stderr banner and `#general` system post no longer frame headless as "extra-usage quota" — it isn't, anymore. Messaging now reads as: pane-backed fallback attempted but unavailable → continuing with the default headless path on your normal subscription.
+
 ## [0.0.7.1] - 2026-04-23
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ One command. One shared office. CEO, PM, engineers, designer, CMO, CRO — all v
 
 ## Get Started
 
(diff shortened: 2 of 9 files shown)


FranDias (dismissed):

FranDias (dismissed):

FranDias (approved):

PR #233, opened 2026-04-22
fix(team): coalesce rapid @-tags instead of /clear mid-turn

najmuzzaman-mohammad:
## Summary

Follow-up to #231's pane-dispatch serialization. Observed in prod after v0.61.10: user tagged \`@pm\` twice 13 seconds apart. The 3-second min-gap allowed the second dispatch to fire — but claude's turn takes ~30 seconds, so the second \`/clear\` cut off claude's first reply mid-generation. PM answered only the second question; the first vanished.

## Fix

Instead of trying to detect claude-idle via pane scraping (unreliable — panes run at tiny 1x4 geometries in practice and claude renders to its own internal buffer that tmux capture-pane can't parse cleanly), the queue now **coalesces bursts**:

- Each slug tracks a **60-second coalesce window** against its last send.
- A new notification arriving within that window **MERGES** into the pending next dispatch's \`Notification\` field (prompts separated by \`\\n\\n---\\n\\n\`) rather than triggering a fresh \`/clear\`.
- The worker, after popping the queue head, waits out the remainder of the coalesce window before sending — any mergers that land during the wait get picked up via a re-peek of the queue head at pop time.
- Single tags still dispatch immediately. Coalesce only engages when claude is in flight from a recent send.

Net effect: bursts of \`@pm\` tags for the same agent become **one** claude turn that answers every question. No more dropped questions, no more mid-turn \`/clear\`.

## Test Coverage

\`internal/team/pane_dispatch_queue_test.go\`:
- **CoalescesBurstsIntoOneDispatch**: two rapid tags produce 2 dispatches total — the first solo prompt, and the second with the merged content. Verifies both prompts land.
- **SingleTagDispatchesImmediately**: no-burst baseline; the fast path fires in < \`coalesceWindow/2\`.
- **DifferentSlugsRunInParallel**: per-slug coalesce doesn't cross-lock — a slow agent doesn't starve others.

## Test plan

- [x] \`go test -run TestQueuePane ./internal/team/\` — 3 new tests green
- [x] Full \`go test ./internal/team/\` passes except pre-existing flake (\`TestPostHeadlessFinalMessageIfSilent\`, verified passing in isolation)


FranDias (approved):

PR #235, opened 2026-04-23
ci(approver): swap App-token for FRAN_GOD_MODE PAT

FranDias:
## What

Swap `.github/workflows/approve-on-comment.yml` from the GitHub-App-token path to the `FRAN_GOD_MODE` PAT.

## Why

First live `/approve` invocation on PR #234 failed at the `actions/create-github-app-token` step:

```
Error: The 'client-id' (or deprecated 'app-id') input must be set to
a non-empty string.
```

The workflow was referencing `vars.APPROVER_APP_ID` + `secrets.APPROVER_APP_PRIVATE_KEY`, neither of which are configured on this repo (and no run of this workflow has ever succeeded — history is skipped, skipped, skipped, failure).

This repo already has a `FRAN_GOD_MODE` PAT provisioned. Using it directly drops the mint step and the two App-config requirements.

## Added safety net

A preflight step fails with a human-readable error if `secrets.FRAN_GOD_MODE` is missing, so next time this breaks the diagnostic is one line instead of a node stack trace.

## Caveat (documented inline)

GitHub rejects approvals submitted by the PR's own author even if the token belongs to a different identity they control. The PAT account and the PR author must be distinct for the self-approve flow to actually land an approval. If `FRAN_GOD_MODE` is tied to the same account as the PRs being self-approved, this still won't approve — we'd need a different account's token.

## Test plan

- [ ] Merge this.
- [ ] On any open PR (e.g. #234), post `/approve` as a member of nex-crm.
- [ ] Workflow run should mint-free, verify org membership, submit review.
- [ ] If approval submits but is rejected because author==token-owner, we know the next move is a bot-account PAT.

### DIFF
--- a/.github/workflows/approve-on-comment.yml
+++ b/.github/workflows/approve-on-comment.yml
@@ -2,8 +2,14 @@ name: Approve on Comment
 
 # Inlined (not using private nex-crm/.github reusable) because this is a
 # public repo and public repos cannot consume private reusable workflows.
-# Keep in sync with nex-crm/.github/.github/workflows/approve-on-comment.yml
-# (current canonical SHA tracked via the v1 tag on that repo).
+#
+# Auth: uses the FRAN_GOD_MODE PAT (org-wide, scoped for repo + read:org)
+# rather than a GitHub App token. Simpler — no APPROVER_APP_ID / APPROVER_
+# APP_PRIVATE_KEY config needed. The PAT must be tied to an identity that
+# can approve the target PR under the repo's branch-protection rules
+# (GitHub rejects approvals submitted by the PR's own author even when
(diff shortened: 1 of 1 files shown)


FranDias:
Wrong direction — Fran caught it: `FRAN_GOD_MODE` is Fran's PAT, and GitHub won't let the PR author's own token approve the PR, which is precisely why the GitHub App (a distinct identity) is the right path. The workflow wiring (`vars.APPROVER_APP_ID` + `secrets.APPROVER_APP_PRIVATE_KEY`) is correct. The repo-level var and secret just aren't configured yet — that's a one-time setup via `gh variable set` / `gh secret set` or the GitHub UI.

PR #239, opened 2026-04-23
fix(web): composer caret stays aligned when @-tagging

najmuzzaman-mohammad:
## Summary

User observation after v0.63.1: caret position looks weird in the composer when tagging someone. Each typed character after an \`@agent\` chip sits a few pixels behind its rendered position.

## Root cause

The mirror-overlay chips picked up the same \`.mention\` CSS used in message bubbles:
- \`padding: 0 5px\` (messages.css base)
- \`padding: 1px 4px\` (nex theme override)
- \`font-size: 0.9em\`
- \`display: inline-block\`

These change the rendered horizontal width of a chip relative to the raw characters (\`@pm\` = 3 characters) in the textarea behind it. The textarea's caret is positioned by char count; the mirror positions the chip by rendered width. After every chip the two diverge by ~8–10px, and the visible caret lags typed characters.

## Fix

Composer-scoped override — \`.composer-mirror .mention\` — keeps the background highlight but strips all layout-affecting properties:
- \`display: inline\` (not \`inline-block\`)
- \`padding: 0\`
- \`font-size\`, \`line-height\`, \`font-weight\`: \`inherit\`
- \`border-radius: 0\`

Raw characters in the mirror take the exact width the textarea draws. Caret aligns character-for-character.

Message-bubble chips unchanged — the base \`.mention\` rule still applies outside the composer.

## Test plan

- [x] Type \`@planner hey\` in composer: chip shows background highlight, caret sits exactly after the typed character
- [x] Move caret mid-text with \`setSelectionRange\`: position matches visible text
- [x] Message bubble chips still render with the original styled pill treatment (verified on \`@planner are you done?\` in #general)
- [x] Computed styles: both \`.composer-input\` and \`.composer-mirror .mention\` at \`fontSize: 15px, lineHeight: 22.5px\`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to WUPHF will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Caret no longer drifts when typing past an `@agent` chip in the composer.** The mirror-overlay treatment rendered mention chips with 4–5px horizontal padding and `font-size: 0.9em`, so each chip took more width than the raw text in the textarea behind it. The caret fell behind the typed characters by a few pixels after every chip. Composer chips now inherit the textarea's font metrics exactly (15px, no padding, `display: inline`) and only apply a background highlight — layout-neutral, so character-for-character alignment with the textarea is preserved. Message-bubble chips keep the original styled pill treatment.
+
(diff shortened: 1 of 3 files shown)


FranDias (approved):

PR #250, opened 2026-04-23
fix(team): reduce agent token bloat + dispatch reliability

najmuzzaman-mohammad:
## Summary

- **Headless by default**: removes the mandatory tmux pane spawn at startup; `claude --print` per turn runs on the user's normal subscription quota (Anthropic re-sanctioned this in the 2026-04 OpenClaw policy note)
- **Duplicate broadcast filter**: 30 s window + 0.85 similarity threshold drops paraphrased re-posts from the same agent in the same thread, cutting token burn on chatty turns
- **Prompt cache stability**: sorts `officeMembers` and `activePolicies` by slug before building the prompt so the prefix is byte-identical across spawns and prompt-cache hits land reliably
- **Tool hygiene rules in prompt**: agents are told all `team_*`/`human_*` tools are already registered — no `ToolSearch` needed — closing the most common cause of wasted turns
- **Wizard-hired agent dispatch**: `activeSessionMembers` now includes broker-only members (agents hired mid-session via the wizard) so @-tags and DMs no longer silently route to the CEO
- **Ghost pane guard**: `agentPaneTargets` returns early when `paneBackedAgents` is false, preventing ghost `PaneTarget` strings that bypass the headless dispatcher and drop messages into dead tmux sessions
- **Activity watchdog**: reaps agents stuck in `active`/`thinking` after a crash; `activityWatchdogEnabled` flag lets TestMain disable it so short-lived test brokers do not accumulate goroutines

## Test plan

- [ ] `go test ./internal/team/... -timeout 120s` passes green
- [ ] Launch office in web mode, confirm agents dispatch headless without tmux (no pane-spawn attempt at startup)
- [ ] Hire an agent via wizard mid-session, confirm @-tags reach that agent
- [ ] Confirm no ghost-pane dispatch when `paneBackedAgents = false`
- [ ] Enable prompt caching (`ANTHROPIC_PROMPT_CACHING=1`), run two sequential turns, confirm cache hit reported

### DIFF
--- a/internal/team/broker.go
+++ b/internal/team/broker.go
@@ -535,7 +535,8 @@ type Broker struct {
 	lastAgentRateLimitPrune time.Time
 	agentLogRoot            string // override for tests; empty means agent.DefaultTaskLogRoot()
 
-	stopCh chan struct{} // closed by Stop(); signals background goroutines to exit
+	stopCh   chan struct{} // closed by Stop(); signals background goroutines to exit
+	stopOnce sync.Once
 }
 
 func taskNeedsLocalWorktree(task *teamTask) bool {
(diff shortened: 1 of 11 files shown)


FranDias (dismissed):

FranDias (dismissed):

FranDias (approved):

PR #252, opened 2026-04-23
fix(wiki): Pam rides the tab bar — visible across the wiki, desk on the divider

najmuzzaman-mohammad:
## Summary

- **Placement**: Pam's desk now sits flush on the bottom divider line of the Wiki / Notebooks / Reviews tab bar. Previously she was absolutely positioned inside `.wk-article-col`, which has `overflow-y: auto` — that silently clipped her avatar whenever she tried to rise above the column top, which is why she looked like she was hiding behind the article.
- **Visibility**: hoisted from `WikiArticle` up to `WikiTabs` so she renders across every wiki surface (catalog, article, audit), not just article views.
- **Cross-tab behavior**: on Notebooks / Reviews tabs she renders as disabled scenery with an "Open an article to use Pam." empty state in the menu, so she reads as wiki chrome without misfiring.
- **Refresh plumbing**: article refetch after a Pam action is now App → WikiTabs (Pam host) for the callback, App → Wiki → WikiArticle for the refresh nonce. Same UX, state just lifted one level.

## Files

- `web/src/components/wiki/Pam.tsx` — `articlePath: string \| null`, disabled-state empty menu.
- `web/src/components/wiki/WikiTabs.tsx` — renders `<Pam>` inside the nav.
- `web/src/components/wiki/Wiki.tsx` — drops Pam render; accepts `externalRefreshNonce` prop.
- `web/src/components/wiki/WikiArticle.tsx` — accepts `externalRefreshNonce`, wired into the fetch effects.
- `web/src/App.tsx` — owns `articleRefreshNonce` + computes `pamArticlePath`; passes through.
- `web/src/styles/pam.css` — `.wiki-tabs` gets `position: relative; overflow: visible`; `.pam-wrap` uses `bottom: 0` so the desk's bottom sits exactly on the divider.

## Test plan

- [x] `npm test` — 302/302 web tests pass
- [x] Dev build (`wuphf-dev --broker-port 7897 --web-port 7898`, isolated `HOME`) — verified visually:
  - Pam visible on the Wiki catalog, articles, audit, Notebooks, and Reviews
  - Desk bottom sits on the tab-bar divider line
  - Menu opens; "Open an article to use Pam." shown when off the Wiki tab or on catalog/audit
  - Article enrichment action still runs and refreshes the article

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -105,6 +105,10 @@ function MainContent() {
   const notebookAgentSlug = useAppStore((s) => s.notebookAgentSlug)
   const notebookEntrySlug = useAppStore((s) => s.notebookEntrySlug)
   const setNotebookRoute = useAppStore((s) => s.setNotebookRoute)
+  // Pam's onActionDone bumps this; Wiki re-fetches article + history when
+  // the prop changes. Lifted up here because Pam lives inside the tab bar
+  // (so her desk can rest on the divider line).
+  const [articleRefreshNonce, setArticleRefreshNonce] = useState(0)
 
   if (!currentApp && isDMChannel(currentChannel, channelMeta)) {
     return <DMView />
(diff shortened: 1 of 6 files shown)


FranDias (approved):

PR #254, opened 2026-04-23
feat(wiki): Slice 2 Thread B — extractor closes §7.4 substrate guarantee via fact-log JSONL

najmuzzaman-mohammad:
## Summary

- Closes the substrate-rebuild leak on the extraction path: before this PR, `rm -rf .wuphf/index/` + `ReconcileFromMarkdown` re-found zero extracted facts because the extractor wrote the index but not the markdown.
- Extraction now commits an append-only JSONL fact log (`wiki/facts/{kind}/{slug}.jsonl`) under the `archivist` git identity, batched per-entity, routed through `WikiWorker` so the single-writer invariant stays intact.
- `ReinforcedAt` excluded from `CanonicalHashFacts` (same-content hashes match across re-extractions) but included in `CanonicalHashAll` (end-to-end drift detection). §7.4 of `docs/specs/WIKI-SCHEMA.md` now documents the asymmetry.

## Files touched

- `internal/team/entity_commit.go` — new `Repo.AppendFactLog` that reads, appends, and commits a fact-log file inside the repo mutex.
- `internal/team/wiki_worker.go` — new `IsFactLogAppend` request flag + `EnqueueFactLogAppend` public method routing through the existing single-writer queue.
- `internal/team/wiki_extractor.go` — `factLogPath(kind, slug)` helper + `persistFactLogs` batching new (non-reinforced) facts per-entity and calling `EnqueueFactLogAppend` after `SubmitFacts` succeeds. Append failures go to DLQ without double-failing the extractor.
- `internal/team/wiki_index.go` + `internal/team/wiki_index_sqlite.go` — `CanonicalHashFacts` clones each fact with `ReinforcedAt=nil` before hashing, on both backends.
- `docs/specs/WIKI-SCHEMA.md` §7.4 — documents the extractor closure + hash asymmetry.
- `internal/team/wiki_extractor_closure_test.go` — `TestExtractionSurvivesReboot`, `TestExtractionAppendsJSONL`, `TestReinforcementHashInvariance`.

## Out of scope (Slice 3)

- **Ghost-entity briefs.** The extractor mints `IndexEntity` rows in-memory but doesn't commit a `team/{kind}/{slug}.md` brief — brief authorship belongs to `EntitySynthesizer`. `CanonicalHashAll` doesn't fully round-trip through extraction alone; entity briefs enter the hash only after synthesis runs. Documented in the reboot test's comment. Slice 3 should unify the two paths so `CanonicalHashAll` round-trips end-to-end.

## Verification

- `golangci-lint run --timeout=5m ./...` → 0 issues.
- `go test ./internal/team/ -count=1 -timeout 120s` passes (one pre-existing flake `TestWikiIndex_ReconcileMutexNarrow` — 50ms timing assertion under full-suite load, passes in isolation, unrelated to Thread B).
- Thread B tests at `-count=10` deterministic green.

## Test plan

- [ ] `go test ./internal/team/ -run \"TestExtractionSurvivesReboot|TestExtractionAppendsJSONL|TestReinforcementHashInvariance\"` passes
- [ ] `go test ./internal/team/ -run TestExtractionSurvivesReboot -count=10` passes deterministically
- [ ] `golangci-lint run --timeout=5m ./...` → 0 issues
- [ ] On a fresh broker: extract 5 facts, check `wiki/facts/people/<slug>.jsonl` has 5 lines, `git log --author=archivist` shows 1 commit per entity

### DIFF
--- a/docs/specs/WIKI-SCHEMA.md
+++ b/docs/specs/WIKI-SCHEMA.md
@@ -311,6 +311,12 @@ fact_id = sha256(artifact_sha + "/" + sentence_offset + "/" + norm(subject) + "/
 
 `rm -rf .wuphf/index/` → restart broker → boot reconcile runs → SQLite + bleve re-indexed from markdown. The result must be **logically identical**, not byte-identical. Logical identity means: `SELECT * FROM facts ORDER BY id` produces the same canonical hash pre- and post-rebuild.
 
+**Every code path that introduces a new fact must append it to `wiki/facts/{kind}/{slug}.jsonl` via `WikiWorker.EnqueueFactLogAppend` under the `archivist` identity.** The extraction loop, human `save_as_insight`, and any future synthesis-time fact mint all honor this contract — markdown is the source of truth, and a fact that lives only in the derived cache violates the rebuild guarantee.
+
+Reinforcement is a read-side concept: when the same `fact_id` is re-extracted, only `reinforced_at` advances in the index. No new JSONL line is appended, and the on-disk fact log remains canonical.
+
+**`ReinforcedAt` hash policy.** `CanonicalHashFacts` EXCLUDES `reinforced_at` from its input, so two extraction runs on the same artifact (the second one purely bumping `reinforced_at`) produce an identical hash. That makes `CanonicalHashFacts` the load-bearing rebuild-invariance check. `CanonicalHashAll` INCLUDES `reinforced_at` (alongside entities, edges, and redirects) and so advances whenever any layer — reinforcement included — changes. Use `CanonicalHashAll` for end-to-end drift detection and `CanonicalHashFacts` for the rebuild contract test.
+
 ---
 
 ## 8. Decay & temporal validity
--- a/docs/specs/WIKI-SCHEMA.md
+++ b/docs/specs/WIKI-SCHEMA.md
@@ -317,6 +317,8 @@ Reinforcement is a read-side concept: when the same `fact_id` is re-extracted, o
 
 **`ReinforcedAt` hash policy.** `CanonicalHashFacts` EXCLUDES `reinforced_at` from its input, so two extraction runs on the same artifact (the second one purely bumping `reinforced_at`) produce an identical hash. That makes `CanonicalHashFacts` the load-bearing rebuild-invariance check. `CanonicalHashAll` INCLUDES `reinforced_at` (alongside entities, edges, and redirects) and so advances whenever any layer — reinforcement included — changes. Use `CanonicalHashAll` for end-to-end drift detection and `CanonicalHashFacts` for the rebuild contract test.
 
+**Append-failure closure.** When `EnqueueFactLogAppend` fails (queue saturated, local I/O, git error), the failure is routed to the DLQ under the dedicated `fact_log_persist` category — NOT `provider_timeout`. The replay path retries the APPEND only, reading the current fact log and appending any missing `fact_id`s. Re-running extraction would treat the fact as reinforcement per §7.3 and never write the missing JSONL line, permanently breaking §7.4 for that fact. See §11.13 for category details.
+
 ---
 
 ## 8. Decay & temporal validity
@@ -456,14 +458,22 @@ The DLQ (`wiki/.dlq/extractions.jsonl`) holds extraction failures that are eligi
   "artifact_path": "wiki/artifacts/chat/3f9a21bc.md",
   "kind": "chat",
   "last_error": "json_parse_error: expected { got \"\`\`\`json\\n{\"",
-  "error_category": "parse | provider_timeout | validation",
+  "error_category": "parse | provider_timeout | validation | fact_log_persist",
   "retry_count": 2,
   "max_retries": 5,
   "first_failed_at": "2026-04-22T13:00:00Z",
   "last_attempted_at": "2026-04-22T15:10:00Z",
-  "next_retry_not_before": "2026-04-22T15:20:00Z"
+  "next_retry_not_before": "2026-04-22T15:20:00Z",
+  "fact_log_append": {
+    "kind": "person",
+    "slug": "sarah-chen",
+    "artifact_sha": "3f9a21bc",
+    "jsonl_lines": "{\"id\":\"a3f9b2c14e8d\",...}\n"
+  }
 }
 ```
+
+`fact_log_append` is populated only for `error_category = "fact_log_persist"`. Carries the exact state needed for the append-replay path to reconstruct the `AppendFactLog` call without re-running extraction — see the §7.4 closure guarantee below.
 
 #### Tombstone rows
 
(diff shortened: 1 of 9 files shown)


FranDias (dismissed):

najmuzzaman-mohammad:
## Review blockers addressed

All three blockers are fixed, plus the worth-fixing items. New commits:
- `30e3b4ab` fix(wiki): route fact-log append failures to a dedicated DLQ category
- `3f6112b2` fix(wiki): tighten fact-log path regex and align Append permissions
- `6269bd4a` test(wiki): fact-log-persist DLQ category, replay recovery, reconcile proof
- `340be7b4` docs(wiki): document fact_log_persist DLQ category in §11.13 and §7.4

### 1. CRITICAL — `ErrQueueSaturated` silently drops facts from the JSONL substrate

**Fix.** New category `DLQCategoryFactLogPersist` carries a `FactLogAppendPayload { Kind, Slug, ArtifactSHA, JSONLLines }`. `ReplayDLQ` now dispatches by category: persist entries go through `replayFactLogAppend`, which reads the current fact log, dedupes by `fact_id`, and appends any still-missing lines via `EnqueueFactLogAppend`. The replay is fully idempotent under retry — an already-landed line is detected and the replay resolves as a no-op with a tombstone. Extraction-class entries continue to go through `ExtractFromArtifact` unchanged.

The DLQ row key for persist entries is composed as `factlog:{kind}:{slug}:{artifact_sha}` (see `FactLogAppendSHA`) so two concurrent per-entity append failures from one artifact do not clobber each other under `readLatestStateLocked` last-write-wins.

**Tests proving it.**
- `TestExtractAppendSaturationLandsInFactLogPersistDLQ` — saturation lands under the right category with the right payload (and explicitly asserts it is NOT `provider_timeout`).
- `TestReplayDLQRecoversFactLogAppend` — end-to-end: saturate → DLQ row lands → replay writes the line → fresh `NewWikiIndex` + `ReconcileFromMarkdown` finds the fact.
- `TestReplayDLQFactLogAppendIsIdempotent` — line already on disk → replay is a no-op.
- `TestExtractAppendSaturationDistinctEntitiesDoNotCollide` — two simultaneous per-entity failures produce two distinct DLQ rows.

### 2. HIGH — Wrong DLQ category for append failures

**Fix.** `persistFactLogs` now routes append errors to the new `DLQCategoryFactLogPersist` via `queueFactLogPersistDLQ`. `DLQCategoryProviderTimeout` is no longer reachable from the append path. Category table added to §11.13 documenting what each one routes to.

**Tests proving it.** `TestExtractAppendSaturationLandsInFactLogPersistDLQ` explicitly asserts `ErrorCategory != DLQCategoryProviderTimeout`.

### 3. CRITICAL — `TestExtractionSurvivesReboot` test validity

**Finding.** `ReconcileFromMarkdown` already walks `wiki/facts/**/*.jsonl` via `walkFactLogs` (`wiki_index.go:436-444`) and feeds every line into both `store.UpsertFact` and `text.Index`. That path was correct. What the reboot test was missing was a direct, isolated proof that the rebuild literally consumes the JSONL file — without any extractor state in memory to blur the assertion.

**Fix (option b, strengthened).** Added `TestReconcileFromMarkdownReadsFactLogJSONL`: seeds a single `TypedFact` on disk by hand (no worker, no extractor, no prior in-memory state), asserts a fresh `NewWikiIndex` does NOT have the fact, calls `ReconcileFromMarkdown`, then asserts `GetFact` AND BM25 `Search` both surface it. Plus `TestReplayDLQRecoversFactLogAppend` closes the loop end-to-end: "queue saturation → DLQ → replay → rm-rf index → reconcile → fact searchable."

### Worth-fixing items — all landed

- **Partial batch abandon.** `serializeFactsAsJSONL` now logs and skips the failing fact, serializing every successfully-marshalable sibling. Prior code returned on the first failure and dropped the whole `(kind, slug)` group.
- **`factLogPathPattern` too permissive.** Tightened to `^wiki/facts/[a-z][a-z0-9-]*/[a-z0-9][a-z0-9-]*\.jsonl$`, mirroring `entityFactPathPattern`. Also tightened `factLogNewSchema` in `wiki_index.go` to match, keeping the validate-at-commit and classify-at-reconcile regexes aligned.
- **Permission inconsistency.** `AppendFactLog` now uses 0o700/0o600, matching `CommitEntityFact`.

### Verification

- `go test -race ./internal/team/...` for every wiki substrate test (`TestExtract*`, `TestReplayDLQ*`, `TestReconcile*`, `TestFact*`, `TestDLQ*`, `TestReinforcement*`, `TestWikiIndex*`, `TestPathRouting*`, `TestCanonicalHash*`) passes clean across multiple runs.
- `gofmt -w`, `goimports -w`, `go vet ./...` all clean.
- Single-writer invariant preserved — every fact-log mutation still goes through `WikiWorker.EnqueueFactLogAppend` (the test seam merely swaps the function the extractor calls; the production wiring is unchanged).
- Fact_id determinism untouched.

### Design calls

1. **Replay-append idempotency via on-disk read.** Alternative was to trust the DLQ row to know whether it had landed. I read the fact log instead — it is the canonical record of truth per §7.4, and any replay that disagreed with what is on disk would itself be a bug. Ten lines of I/O, zero new invariants.
2. **Composite DLQ key `factlog:{kind}:{slug}:{artifact_sha}`.** The DLQ schema keys by `artifact_sha`, and one artifact can produce fact-log failures for multiple entities. I did not want to reshape the on-disk DLQ schema, so the synthesized SHA form keeps every other code path untouched while giving the persist entries their own key-space.
3. **Test seam `setAppendFactLog` on the Extractor.** Race-safe saturation simulation via worker-lifecycle games turned out to be fragile. A 2-line function pointer injection is far cleaner and keeps the production call-site identical (the seam defaults to `worker.EnqueueFactLogAppend`).

FranDias (dismissed):

najmuzzaman-mohammad:
## Reinforcement regression fix pushed (`46829c5a`)

**Root cause**: `reconcileFactLog` was overwriting the in-memory fact row with the JSONL row verbatim. `ReinforcedAt` is never serialized to JSONL by design (§7.3 — the append-only log stays immutable; reinforcement is a runtime signal bumped on second extraction).

When a post-append reconcile fires after a reinforcement bump, it read JSONL seeing `ReinforcedAt=nil` and clobbered the in-memory bump. Symptom on CI: `TestExtractIdempotentOnRepeat` → "expected reinforced_at to be set after second extract". Real failure mode: every JSONL reconcile path (boot, DLQ replay, external git sync) silently un-reinforces facts.

**Fix**: before `UpsertFact`, if the decoded row has `ReinforcedAt=nil` AND the store already has the fact with `ReinforcedAt` set, carry forward the in-memory value.

```go
if f.ReinforcedAt == nil {
    if existing, ok, _ := w.store.GetFact(ctx, f.ID); ok && existing.ReinforcedAt != nil {
        f.ReinforcedAt = existing.ReinforcedAt
    }
}
```

**Regression guard**: `TestReconcileFactLogPreservesInMemoryReinforcement` in `wiki_extractor_persist_dlq_test.go` covers the general case, not just the narrow `TestExtractIdempotentOnRepeat` symptom.

Local: `go test -race -count=1` green on all 4 affected tests.

FranDias (dismissed):

FranDias (approved):

PR #258, opened 2026-04-23
[codex] fix workspace shred fresh start

najmuzzaman-mohammad:
## Summary

This changes workspace shred into a true fresh-start operation. The wipe now removes runtime history and local state that could repopulate stale messages: logs, sessions, provider session state, Codex headless state, calendar, local markdown wiki memory, office state, workflows, team state, onboarding, and company identity.

It still preserves user work and credentials: task worktrees, `config.json`, and OpenClaw device identity.

## Root Cause

The web shred endpoint deleted selected files on disk while the live broker and agent loops kept running with old in-memory state. Because logs and sessions were intentionally preserved, and the broker could continue after shred, old messages could remain visible or be written back after restart.

## User Impact

After shredding from Settings, WUPHF now stops after returning a successful response. Relaunching `wuphf` reopens onboarding from a clean workspace instead of carrying old messages or logs forward. The web client also uses a bounded request timeout so the UI does not spin forever if the broker fails to respond.

## Validation

- `go test ./internal/workspace ./internal/team`
- `go test ./cmd/wuphf`
- `npm run typecheck`
- `git diff --check`

### DIFF
--- a/cmd/wuphf/main.go
+++ b/cmd/wuphf/main.go
@@ -51,10 +51,11 @@ func printSubcommandHelp(sub string) {
 		fmt.Fprintln(os.Stderr, "wuphf shred — burn the whole workspace down")
 		fmt.Fprintln(os.Stderr, "")
 		fmt.Fprintln(os.Stderr, "Stops the running session, clears broker state, and deletes the team")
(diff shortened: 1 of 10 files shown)


FranDias (approved):

PR #259, opened 2026-04-23
feat(wiki-web): Slice 2 Thread E — polish for /lookup, /lint, resolve flow

najmuzzaman-mohammad:
## Summary

Slice 2 Thread E of the wiki intelligence port — frontend-only polish for the Thread A–D user touch points.

- **ResolveContradictionModal pending state.** Clicked pick button shows a subtle neutral spinner + "Resolving…" label inline; all three picks + Cancel go disabled while the POST is in flight. `aria-busy` set on the pending button for a11y. No layout shift — spinner is an inline 12×12 element with margin-right for the label gap. On error, the existing inline banner surfaces the message (no error toast).
- **Post-resolve success toast with commit sha.** After `resolveContradiction` returns, fires a `success`-type toast with copy `Resolved. Commit {short_sha}.` (7-char sha, git's default width). No commit-viewer route exists yet, so sha is text in the toast body rather than a link — spec §4 explicitly allows this.
- **/lint autocomplete entry.** Extends `FALLBACK_SLASH_COMMANDS` (and the matching broker-icon map) with `/lint` and `/lookup`. Descriptions match the Thread E copy spec verbatim (`"Review wiki for contradictions and stale facts."` for lint, `"Cited answer from the team wiki"` for lookup). The broker already registers both with `webSupported=true`, so live deployments worked — this brings the offline fallback back in line.
- **/lookup loading line.** Replaces the skeleton blocks in `CitedAnswer`'s loading state with a single Fraunces-italic "Searching the wiki…" line. Matches DESIGN-WIKI.md hatnote typography (§9), preserves `role="status"` + `aria-busy`, renders via new `.wk-lookup-status` class. Light mode, no amber, no dark surfaces.

Shared primitive: scoped `.wk-spinner` in `web/src/styles/wiki.css`, neutral border + muted-text top color so it reads as "working" without stealing attention. Uses the global `@keyframes spin` that already lives in `global.css`.

## Design-system calls

- DESIGN-WIKI.md forbids amber as decoration; the spinner and status line use neutral tokens only.
- No purple, no gradients, no system-ui on body text — status line uses `var(--wk-body-serif)` italic.
- Toast reuses the existing global `showNotice` / `ToastContainer` primitive (no new UI lib).
- Operator-language preserved on severity labels ("Needs attention / Worth a look / FYI" already shipped in WikiLint; no change needed).

## Typecheck + tests

- `npm run typecheck` — clean.
- `npm test` — **68 files, 388 tests passing**. New tests:
  - `ResolveContradictionModal.test.tsx` × 4 — pending spinner, inline-error (no toast), short-sha toast, empty-sha fallback toast.
  - `useCommands.test.ts` +2 — lock in the /lookup and /lint fallback copy.
  - `CitedAnswer.test.tsx` — rewrites the skeleton assertion to assert the Fraunces-italic status line.

## Test plan

- [ ] Manually hit `/lookup who is sarah?` in the composer — confirm the `wiki-lookup` view shows "Searching the wiki…" in Fraunces italic while the request is in flight.
- [ ] Type `/l` in the composer — confirm autocomplete lists `/lookup`, `/lint`, and other `/l…` commands.
- [ ] Open the wiki health check (`/lint` → contradictions finding) → click Resolve → click Fact A. Confirm the spinner renders inline on Fact A only, Cancel is disabled, and on success a toast reads "Resolved. Commit abc1234."
- [ ] Force the resolve endpoint to fail (e.g. broker offline) — confirm the error renders inside the modal banner, not a toast, and the buttons re-enable so the user can retry.

## Skipped

- Screenshots/GIFs: running headless in a worktree without a dev server, and the task spec explicitly says to skip if we can't capture them headlessly.
- Playwright E2E for /lookup spinner > 200ms (mentioned under Thread E "Tests"): Playwright is an e2e package in `web/e2e/` — would require a running dev server and broker. Unit coverage in `CitedAnswer.test.tsx` pins the loading-state markup.

## Release Notes

* **New Features**
  * Added `/lookup` and `/lint` slash commands to autocomplete.

* **UI/UX Improvements**
  * Replaced skeleton loaders with a "Searching the wiki…" status message during wiki searches.
  * Enhanced contradiction resolution modal with per-button loading states and spinners.
  * Added success toast notifications showing commit information after resolving contradictions.

### DIFF
--- a/web/src/components/wiki/CitedAnswer.test.tsx
+++ b/web/src/components/wiki/CitedAnswer.test.tsx
@@ -1,141 +1,128 @@
-import { render, screen, waitFor } from "@testing-library/react";
-import { beforeEach, describe, expect, it, vi } from "vitest";
-
-import * as apiClient from "../../api/client";
-import CitedAnswer, { type QueryAnswer } from "./CitedAnswer";
+import { describe, expect, it, vi, beforeEach } from 'vitest'
+import { render, screen, waitFor } from '@testing-library/react'
+import CitedAnswer, { type QueryAnswer } from './CitedAnswer'
+import * as apiClient from '../../api/client'
 
 const STUB_ANSWER: QueryAnswer = {
-  query_class: "status",
-  answer_markdown: "Sarah Jones is VP of Sales at Acme Corp.<sup>[1]</sup>",
+  query_class: 'status',
+  answer_markdown: 'Sarah Jones is VP of Sales at Acme Corp.<sup>[1]</sup>',
   sources_cited: [1],
   sources: [
     {
-      kind: "fact",
-      slug_or_id: "people/sarah-jones",
-      title: "Sarah Jones",
-      excerpt: "Sarah Jones is VP of Sales at Acme Corp.",
-      valid_from: "2026-01-01",
+      kind: 'fact',
+      slug_or_id: 'people/sarah-jones',
+      title: 'Sarah Jones',
+      excerpt: 'Sarah Jones is VP of Sales at Acme Corp.',
+      valid_from: '2026-01-01',
       staleness: 0.1,
-      source_path: "team/people/sarah-jones.md",
+      source_path: 'team/people/sarah-jones.md',
     },
   ],
   confidence: 0.9,
-  coverage: "complete",
+  coverage: 'complete',
   latency_ms: 42,
-};
+}
 
 const STUB_GENERAL: QueryAnswer = {
-  query_class: "general",
+  query_class: 'general',
   answer_markdown: "I don't have information about that.",
   sources_cited: [],
   sources: [],
   confidence: 0.85,
-  coverage: "none",
+  coverage: 'none',
   latency_ms: 5,
-};
+}
 
-const _STUB_PARTIAL: QueryAnswer = {
+const STUB_PARTIAL: QueryAnswer = {
   ...STUB_ANSWER,
-  coverage: "partial",
+  coverage: 'partial',
   sources_cited: [],
-};
+}
 
-describe("<CitedAnswer>", () => {
+describe('<CitedAnswer>', () => {
   beforeEach(() => {
-    vi.restoreAllMocks();
-  });
+    vi.restoreAllMocks()
+  })
 
-  it("shows loading skeleton while fetching", () => {
-    vi.spyOn(apiClient, "get").mockReturnValue(new Promise(() => {}));
-    render(<CitedAnswer query="Who is Sarah Jones?" />);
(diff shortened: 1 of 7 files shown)


FranDias (dismissed):

najmuzzaman-mohammad:
Review fixes pushed (3 commits):

- `e54e322f` — aria-busy now emits the ARIA 1.1 string token `"true"` when busy and is omitted otherwise (matches `CitedAnswer`'s existing convention). Test asserts both `getAttribute('aria-busy') === 'true'` on the clicked button and `hasAttribute('aria-busy') === false` on the siblings.
- `249306e0` — dropped the trailing period on `/lint` so it aligns with every other fragment-style description in `FALLBACK_SLASH_COMMANDS`. Confirmed via grep that no other entry carries trailing punctuation. Locked-string test updated.
- `d8034ba4` — `.wk-spinner` now references a locally-defined `@keyframes wk-spin` in `wiki.css` (mirrors `@keyframes wk-pulse` ~line 162). `@keyframes spin` stays in `global.css` — `search.css` and `onboarding.css` still depend on it.

`bun run typecheck` passes. `bun run test` passes (388/388).

FranDias (dismissed):

najmuzzaman-mohammad:
Addressed both CodeRabbit findings in d3912ce1.

**Fix 1 — setState after unmount (`ResolveContradictionModal.tsx`)**
On success, `onResolved()` calls into WikiLint which refreshes and unmounts this modal. The old `finally` block then reset `submitting` / `pendingWinner` on an unmounted component. Changed to return early after `onResolved()` and only reset state in the error branch (no `finally`).

**Fix 2 — unawaited resolve in test (`ResolveContradictionModal.test.tsx`)**
This is the "spinner visible while in-flight" test. The never-resolving promise is the point — the only reason the post-test `resolveFn!(...)` existed was to avoid trailing act warnings, but it actually caused them by scheduling state updates after the test body returned. Dropped the resolve entirely; cleanup unmounts the component and discards the pending promise. With Fix 1 in place, even a late settle would be a no-op (success path no longer touches local state), so a `waitFor` on `not.toBeDisabled()` wouldn't work anyway.

**Verification**
- `npm run typecheck` passes.
- `npm test` passes (388/388). Modal file in isolation: 4/4, no act warnings, no AbortError noise.
- Full-suite AbortError count unchanged (7 before, 7 after) — those are pre-existing noise from other tests, not introduced here.

FranDias (dismissed):

FranDias (approved):

PR #261, opened 2026-04-23
feat(wiki): Slice 2 Thread C — playbook clusters surface reinforced patterns

najmuzzaman-mohammad:
## Why

Slice 2 Thread C of the WIKI-SLICE2-PLAN: playbook synthesis currently runs per-entity. This PR extends it so clusters of reinforced facts across entities surface as cross-workspace patterns in the learnings section.

## What changed

- Added `FactStore.ListAllFacts(ctx)` with implementations on both `inMemoryFactStore` and `SQLiteFactStore`. Pure additive read-only enumeration used by cross-entity consumers; no writer path or single-writer invariant touched.
- New `internal/team/playbook_clusters.go` with `FactCluster` and `clusterReinforcedFacts(ctx, store, predicateFilter, minEntities)`. The live reinforcement signal is `TypedFact.ReinforcedAt != nil` — the extractor bumps it on re-extraction; no `reinforced_count` field exists in v1.2.
- New `internal/team/prompts/synthesis_playbook_v2.tmpl` — strict superset of the v1 programmatic prompt. Adds a `# Reinforced patterns across entities` input section and a rule-4a instruction to emit `## Patterns across entities` under the learnings heading (entity counts only, OMIT when empty).
- New `internal/team/playbook_synthesizer_v2.go` — parsed-once template + `buildPlaybookSynthUserPromptV2` + `collectReinforcedClusters`.
- `PlaybookSynthesizerConfig` gains `ClusterSource FactStore` + `ClusterMinEntities int`. When `ClusterSource` is nil, the v1 prompt path is byte-identical to pre-Thread-C behavior. Any cluster-collection or v2-render error logs and falls back to v1 — patterns are a supplement, not a gate.

## Tests

Run: `go test ./internal/team/... -run 'Cluster|Playbook' -count=1 -race`

New cases:
- `TestClusterReinforcedFacts` (table-driven): threshold boundaries, non-reinforced skip, same-entity dedup, predicate filter, nil-triplet skip, min-entities clamp, count-desc ordering.
- `TestClusterReinforcedFacts_SQLiteParity`: in-memory and SQLite backends produce identical cluster output from the same seed — read-side complement of the §7.4 rebuild contract.
- `TestClusterReinforcedFacts_NilStore`: nil store returns an error.
- `TestBuildPlaybookSynthUserPromptV2_RendersClusters` / `_EmptyClusters` / `_Deterministic`: prompt render assertions.
- `TestPlaybookSynthesisWithClusters`: end-to-end with `ClusterSource` wired — LLM stub captures the prompt, committed playbook carries the patterns section.
- `TestPlaybookSynthesisFallsBackWhenClusterSourceNil`: v1-path byte-equivalence when `ClusterSource == nil`.
- `TestPlaybookSynthesisEmptyClusterSet`: empty cluster scan still routes through v2 and the OMIT rule holds.

No evals harness exists in the codebase today, so the planned `evals/playbook_clusters.golden.json` was skipped — the deterministic template render test (`TestBuildPlaybookSynthUserPromptV2_Deterministic`) covers the same cache-key-stability property a golden would.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./internal/team/...`
- [x] `gofmt -l internal/team/ | wc -l` → 0
- [x] `go test ./internal/team/... -run 'Cluster|Playbook' -count=1 -race` → pass

## Dependency note

This branch is stacked on #254 (Thread B). It carries Thread B's commits until that PR merges. Rebase onto main once #254 lands, before merging this PR.

## Out of scope (intentional)

- SQL indexing for `(predicate, object)`. Plan documents this as acceptable at Slice 2 (bounded by ~500 facts at bench); introduce an index when the corpus passes ~10k facts.
- Evals harness wiring — does not exist yet and is tracked separately.

### DIFF
--- a/internal/team/entity_commit.go
+++ b/internal/team/entity_commit.go
@@ -198,15 +198,21 @@ func (r *Repo) CommitFactLog(ctx context.Context, slug, relPath, content, messag
 	return strings.TrimSpace(sha), len(content), nil
 }
 
-// AppendFactLog appends newlineContent to the fact-log file at relPath and
+// AppendFactLog appends additionalContent to the fact-log file at relPath and
 // commits the resulting bytes. The file is created if it does not exist.
 // `additionalContent` must be the raw bytes to append — the caller is
 // responsible for newline-terminating each JSONL record. A trailing newline
 // is added if missing so the final file always ends with "\n".
 //
-// Uses the repo-wide write lock so the read-modify-write sequence is safe
-// against concurrent appenders; the WikiWorker single-writer invariant
-// (§11.5, Anti-pattern 5) routes every caller through this path.
+// Uses the repo-wide write lock so concurrent appenders are serialised; the
+// WikiWorker single-writer invariant (§11.5, Anti-pattern 5) routes every
+// caller through this path.
+//
+// Implementation: O_APPEND on a per-open fd. Cheaper than the earlier
+// read-modify-write for prolific entities whose JSONL files can grow past a
+// few MB — each append is O(bytesWritten) instead of O(filesize). The
+// repo-wide mutex still guarantees exclusivity for the non-atomic
+// "fstat + write" sequence we need to keep the trailing-newline invariant.
 //
 // The accepted relPath shape matches Repo.CommitFactLog: wiki/facts/**/*.jsonl
 // or team/entities/*.facts.jsonl.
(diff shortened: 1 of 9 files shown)


FranDias (dismissed):

FranDias (dismissed):

najmuzzaman-mohammad:
Review findings addressed in `e08cde71`, `1e96aa62`, `ecb4d70c`.

## 1. HIGH — Unbounded `ListAllFacts` observability `e08cde71`

- Added `CountFacts(ctx) (int, error)` to the `FactStore` interface. SQLite implementation is `SELECT COUNT(*) FROM facts`, in-memory returns `len(s.facts)`.
- `clusterReinforcedFacts` now calls `CountFacts` first and emits `log.Printf("playbook clusters: fact count %d exceeds %d — consider paging", count, threshold)` when the count crosses the new `factCountWarnThreshold` constant (5000). Never fails or truncates — silent truncation would hide corpus growth from operators.
- `CountFacts` failure is non-fatal; the scan continues and the underlying `ListAllFacts` error path handles real I/O breakage.
- Test: `TestClusterReinforcedFacts_LogsAboveThreshold` lowers the threshold to 2 via the test-only `setFactCountWarnThresholdForTest` hook, seeds 3 facts, captures `log.Writer()` into a `bytes.Buffer`, and asserts the warning line is present. Paired with `TestClusterReinforcedFacts_DoesNotLogBelowThreshold` to pin the inverse so we don't spam the broker log on every synthesis run.

## 2. MEDIUM — Apply cluster cap inside `clusterReinforcedFacts` `e08cde71`

- Added a `topN int` parameter. Short-circuits after sorting: `if topN > 0 && len(out) > topN { out = out[:topN] }`. Values ≤ 0 mean unbounded (the semantic the tests asserted implicitly).
- `collectReinforcedClusters` passes `MaxClustersForPrompt`. Removed the now-redundant post-truncation in `synthesize`.
- Test: `TestClusterReinforcedFacts_TopNShortCircuits` pins `topN=1` → strongest cluster only; `topN=0` → all 3; `topN=999` → no-op.

## 3. `FactCluster.Count` godoc `e08cde71`

One-line clarification added verbatim: Count reflects distinct entities where the fact was confirmed via re-extraction (`ReinforcedAt != nil`), not all entities where it was observed.

## 4. `AppendFactLog` read-modify-write `1e96aa62`

Switched from `os.ReadFile` + `os.WriteFile` to `os.OpenFile(O_APPEND|O_CREATE|O_WRONLY)` + a single `Write`. Each append is now O(bytesWritten) instead of O(filesize). The repo-wide mutex still gates concurrent writers.

One implementation note worth flagging — macOS returns `EBADF` on `ReadAt` when the fd was opened `O_APPEND|O_WRONLY`, so the trailing-newline probe uses a short read-only `os.Open` + `ReadAt(fi.Size()-1)` *before* opening the appender. Still O(1) — reads one byte regardless of filesize.

### Thread B coupling

Checked `git log origin/feat/wiki-thread-b-extraction-closure -- internal/team/entity_commit.go` — Thread B's branch has no newer commits touching this file (AppendFactLog on both branches currently points at 641b6710, which is shared). Safe to land here.

That said, Thread B's branch is 4 commits ahead on *other* files (hash-invariance work, extractor closure). When Thread B lands, the follow-up synthesis-side PR should rebase and re-run `TestEnqueueFactLogAppendsJSONL` + `TestReinforcementHashInvariance` — both currently pass on this branch with the O_APPEND rewrite, so this is just a reviewer tripwire not a known breakage.

## 5. Test name clarity `ecb4d70c`

Renamed `TestBuildPlaybookSynthUserPromptV2_Deterministic` → `TestBuildPlaybookSynthUserPromptV2_IsIdempotent` and added a docstring pointing at Slice 3 as the home for the golden-harness semantic-quality eval. This test pins cache-key stability only.

## Verification

```
gofmt -w ... && go vet ./internal/team/...
go test ./internal/team/... -race -count=1 -run '<touched tests>' -timeout 120s
```

All touched tests pass under `-race`. Full-suite run surfaced one pre-existing race in `TestAutoPromote_EndToEnd_HumanTagsPM_DispatchesToPM` (stack rooted in `headless_codex.go:281` + `launcher.go:2874` — entirely unrelated to this PR; reproducible against `origin/feat/wiki-thread-c-playbook-clusters` with no diff applied).

FranDias (approved):

PR #264, opened 2026-04-23
[codex] Keep broker alive during workspace shred

najmuzzaman-mohammad:
## What changed

This changes the workspace shred flow so it wipes workspace state and sends the user straight back to onboarding without taking down the running broker.

On the web side, the Settings danger-zone shred action now resets app navigation state in-process and immediately re-enters onboarding instead of showing a success alert and forcing a page reload.

On the backend, the workspace wipe path now clears the live broker runtime in-process and only removes persisted broker state files rather than deleting the whole `~/.wuphf/team` directory. That keeps the broker available while still wiping the workspace data that shred is meant to remove.

## Why

Shredding a workspace was behaving like a destructive runtime teardown. In practice that could leave the broker stopped or looking stopped, which is the wrong contract for this action.

The intended behavior is simpler: wipe the workspace, keep the broker alive, and return the user to onboarding immediately.

## Impact

Users can shred a workspace from Settings and land directly in onboarding without an extra alert/reload cycle.

The broker keeps serving the web app during and after the wipe, so the session stays live and the app can transition cleanly.

## Root cause

The shred flow depended on deleting the broader runtime directory and then reloading the web app to reconstruct state later. That coupled data wipe behavior to broker lifecycle behavior.

## Validation

- `go test ./internal/workspace`
- `go test ./internal/team -run TestWebUIProxyHandlerForwardsOnboardingRoutes -count=1`
- `npm --prefix web test -- --run src/stores/app.test.ts` could not be run in this worktree because `vitest` is not installed locally (`sh: vitest: command not found`)

### DIFF
--- a/docs/specs/WIKI-SCHEMA.md
+++ b/docs/specs/WIKI-SCHEMA.md
@@ -171,7 +171,7 @@ Insights are facts that rise above the noise: status changes, decisions, pattern
   "confidence": 0.95,
   "created_at": "2026-04-22T14:32:00Z",
   "created_by": "archivist",
-  "source": "synthesis" 
+  "source": "synthesis"
 }
 ```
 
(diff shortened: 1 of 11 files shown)


FranDias (dismissed):

FranDias (approved):

PR #265, opened 2026-04-23
[codex] Add web @all tagging

najmuzzaman-mohammad:
## What changed
- added `@all` as a first-class web mention in the autocomplete and mention token renderer
- taught the web client to derive an explicit `tagged` payload from typed mentions before posting messages
- expanded `@all` to all mentionable agent slugs for both the main composer and thread replies
- added tests for `@all` parsing, expansion, and autocomplete ordering

## Why
The broker already supports all-agent wakeups, but the web client only rendered mention chips and sent raw message text. That meant `@all` was missing from the web UX, and web-authored mentions were not consistently promoted into explicit tags at send time.

## Impact
Web users can now type or autocomplete `@all` and get the same broadcast behavior available in the TUI. Regular web mentions also post a concrete `tagged` array, which keeps routing aligned with the visible chips.

## Validation
- `npm run typecheck`
- `npm test`

### DIFF
--- a/web/src/api/client.ts
+++ b/web/src/api/client.ts
@@ -176,13 +176,15 @@ export function postMessage(
   content: string,
   channel: string,
   replyTo?: string,
+  tagged?: string[],
 ) {
-  const body: Record<string, string> = {
+  const body: Record<string, string | string[]> = {
     from: "you",
     channel: channel || "general",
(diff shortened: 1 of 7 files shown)


FranDias (approved):

PR #272, opened 2026-04-24
fix(team): strip inherited GIT_DIR/GIT_CONFIG from wuphf git callers

FranDias:
## Summary

When wuphf runs inside a git hook (pre-push, lefthook, etc.), git exports `GIT_DIR` / `GIT_WORK_TREE` / `GIT_INDEX_FILE` / `GIT_CONFIG_PARAMETERS` pointing at the outer repo. Any `git` subprocess from wuphf that inherits those vars silently targets the outer repo regardless of `cmd.Dir` — because `GIT_DIR` wins over cwd for repo discovery.

**Symptom (reproduced in the wild):** a real working branch's `git reflog` ends up with thousands of alternating `wuphf: init wiki` and short commits, burying actual commits ~2900 entries deep. `Repo.Init()` in `internal/team/wiki_git.go` runs `git init -q -b main` (no-op under inherited `GIT_DIR`) then `git commit -m "wuphf: init wiki"` — which lands on the user's current working branch with the wiki tree staged against the outer index.

## Fix

Strip the problematic env vars in every wuphf codepath that spawns `git` (or spawns a coding agent that spawns `git`):

- `internal/team/worktree.go` — extend `gitCleanEnv` to also strip `GIT_CONFIG`, `GIT_CONFIG_COUNT`, `GIT_CONFIG_KEY_*`, `GIT_CONFIG_VALUE_*`, `GIT_CONFIG_PARAMETERS` (the `-c`-injection family).
- `internal/team/wiki_git.go` — `runGitLockedAs` now uses `gitCleanEnv()` instead of `os.Environ()`. This is the original smoking gun.
- `internal/team/headless_claude.go` — `buildHeadlessClaudeEnv` seeds from `gitCleanEnv()` so spawned claude agents can't inherit `GIT_DIR`.
- `internal/team/headless_codex.go` — same; transitively fixes opencode since `headless_opencode.go` reuses `buildHeadlessCodexEnv`.
- `internal/team/human_identity.go` — `runGitConfig` sets `cmd.Env = gitCleanEnv()` to block `GIT_CONFIG_PARAMETERS` hijack of `--global` reads.

## Regression test

`internal/team/wiki_git_test.go::TestRepoInitIgnoresInheritedGitDir` — creates a sacrificial outer repo with one commit, sets `GIT_DIR`/`GIT_WORK_TREE` to point at it, runs `Repo.Init()` on an unrelated path, asserts the outer HEAD is unchanged. Verified the test catches the bug when the fix is reverted (fails with exactly the real-world symptom: `"wuphf: init wiki"` commit landing on the outer HEAD).

## About the branch history

Three commits: the fix; a leaked `outer: seed` commit + initial test fix; a cleanup commit removing the leaked file and hardening the test runner's subprocess env.

The middle commit exists because the first time I ran `git push`, pre-push ran `TestRepoInitIgnoresInheritedGitDir`, and the test's own `exec.Command("git", "init"/"commit", ...)` subprocesses **inherited `GIT_DIR` from the pre-push hook itself** — reproducing the exact bug on the fix's own PR. Those inner commits landed on this branch before the leak was caught. The test now sets `cmd.Env = gitCleanEnv()` on its subprocesses and the same-class failure found in `internal/migration/writer_test.go` (`git -C <tmpdir> log` inheriting `GIT_DIR`) is fixed in the same cleanup commit. History preserved rather than rewritten (user preference: forward-only cleanup, no destructive reset).

## Test plan

- [x] `go build ./...` — clean.
- [x] `go vet ./internal/team/...` + `golangci-lint` — clean.
- [x] Regression test passes under normal env.
- [x] Regression test **fails** with the env-strip fix reverted (confirmed catches the bug).
- [x] Regression test passes under inherited `GIT_DIR` (the real-world hook scenario).
- [x] `lefthook run pre-push` green — full `go test ./...` passes end-to-end with this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Git environment variable handling to prevent unintended repository operations when executing in certain contexts (such as git hooks). This ensures commands operate on the correct repository and configuration.
  * Added regression test to verify proper repository isolation.


FranDias:
Addressed the CodeRabbit nit on `filterOutGitEnv` (and the staff-review follow-up flagged alongside it) in `59f8e6f3`:

- `team.gitCleanEnv` → `team.GitCleanEnv` (exported).
- `internal/migration/writer_test.go`'s local mirror deleted; now calls `team.GitCleanEnv()` directly.
- All in-package call sites updated; pure rename + dedup, no behaviour change.

Full pre-push suite (`go test ./...`) green on the updated branch.

PR #277, opened 2026-04-24
fix: unblock main CI + eliminate --no-verify reasons on pre-push

FranDias:
## Summary

- **fix(web)**: main CI has been red since PR #276 because `web / Install and build` runs `npm ci` but biome was added via `bun add` — `web/package-lock.json` was never resynced. Regenerated via `npm install --package-lock-only`.
- **fix(team)**: `TestExtractionBatchesFactsPerEntity` fails deterministically under the pre-push hook because the `commitCountForPath` test helper inherits `GIT_DIR` from git's hook env and `git -C <tempRepo> log` queries the outer repo. Scrub env via `GitCleanEnv()` — same pattern `internal/migration/writer_test.go` already uses.
- **fix(onboarding)**: `TestCheckOneFindsOpencodeInCommonUserBinWhenPATHIsMinimal` flaked at the hard 3s `<bin> --version` timeout under parallel `go test ./...` load. Raised to 10s — it's a one-shot env probe, not a hot path.
- **chore(hooks)**: `lefthook.yml` biome pre-commit was passing `web/package-lock.json` through, which biome hard-errors on ("No files were processed"). Added lockfiles to the exclude list. The existing `bun.lock` entry was also a silent no-op (pattern matches repo-root-relative paths, not `root:`-scoped) — added `web/…` entries for both lockfiles.
- **docs(dev)**: documented lefthook hooks in `DEVELOPMENT.md` and called out that `--no-verify` is not the fix.

Goal: you can `git push` this branch with hooks enabled and they go green on a clean checkout.

## Non-goals / not addressed

- `TestBug_HumanDMsSpecialist_Dispatch_SpecialistReceivesTurn` stays skipped (per MEMORY.md — unrelated Launcher teardown issue).
- One-off flake of `TestDefaultPrepareTaskWorktreeOverlaysCompletedSiblingTaskWorkspace` seen under full `./...` + `GIT_DIR`, didn't reproduce in 5 subsequent runs. Probably a package-level state race on the `taskWorktreeRootDir`/`brokerStatePath` override pattern; noted for follow-up.
- Consolidating web's two redundant lockfiles (`bun.lock` + `package-lock.json`) is a separate, larger decision.

## Test plan

- [x] `bun install` at repo root, `git commit` on a doc-only change — pre-commit green.
- [x] `GIT_DIR=\$(pwd)/.git go test -run TestExtractionBatchesFactsPerEntity ./internal/team/` — repros the failure before the fix, passes after.
- [x] `GIT_DIR=\$(pwd)/.git go test ./internal/team/` x5 — all green.
- [x] `GIT_DIR=\$(pwd)/.git go test ./...` x5 — all green (previously flaked onboarding + team).
- [x] `cd web && npm ci && npm run typecheck && npm run build` — all green with new lockfile.
- [x] `git push` with hooks enabled — pre-commit + pre-push green, no \`--no-verify\`.
- [ ] CI `web / Install and build` job goes green on this PR.

## Summary by CodeRabbit

* **Documentation**
  * Added Git hook workflow docs covering pre-commit and pre-push verification steps.

* **Bug Fixes**
  * Made prerequisite checks more reliable by running probes concurrently and increasing timeouts.
  * Fixed test tooling to avoid using an unintended Git environment during history lookups.

* **Chores**
  * Adjusted tooling configuration to exclude specific lockfiles from pre-commit processing.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -8,6 +8,10 @@ go build -o wuphf ./cmd/wuphf
 
 For normal app usage you do not need Bun. The local office/team MCP tools now run from the main Go binary through the hidden `wuphf mcp-team` subcommand.
 
+## Git hooks
+
+Hooks run via [lefthook](https://github.com/evilmartians/lefthook) (`lefthook.yml`). On `pre-push` the full Go test suite and a `cmd/wuphf` build are executed; on `pre-commit` gofmt/go vet/golangci-lint run against staged Go files. Do not push with `--no-verify` — fix the underlying failure instead.
+
 ## Latest Published CLI
 
 The old standalone CLI is no longer vendored in this repo.
--- a/internal/onboarding/prereqs.go
+++ b/internal/onboarding/prereqs.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"os/exec"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/nex-crm/wuphf/internal/runtimebin"
@@ -51,12 +52,25 @@ var prereqSpecs = map[string]prereqSpec{
 // CLI runtimes must be present for wuphf to actually run a turn, but all are
 // marked optional here so the user can proceed with whichever runtime
 // they have.
+//
+// Probes run concurrently. CheckOne's per-probe timeout is 10s (see comment
+// there for rationale) and CheckAll is invoked from an HTTP handler with a
+// 5s client deadline at cmd/wuphf/onboarding.go; running probes serially
+// would mean worst-case wall-clock = 7 × 10s = 70s, far past any sane HTTP
+// budget. Concurrent probes cap wall-clock at max(probe), well under the
+// client timeout. Order of `names` is preserved in the returned slice.
 func CheckAll() []PrereqResult {
 	names := []string{"node", "git", "claude", "codex", "opencode", "cursor", "windsurf"}
-	results := make([]PrereqResult, 0, len(names))
-	for _, name := range names {
-		results = append(results, CheckOne(name))
+	results := make([]PrereqResult, len(names))
+	var wg sync.WaitGroup
+	wg.Add(len(names))
+	for i, name := range names {
+		go func(i int, name string) {
+			defer wg.Done()
+			results[i] = CheckOne(name)
+		}(i, name)
 	}
+	wg.Wait()
 	return results
 }
 
--- a/internal/onboarding/prereqs.go
+++ b/internal/onboarding/prereqs.go
@@ -80,7 +80,13 @@ func CheckOne(name string) PrereqResult {
 	r.OK = true
 
 	// Best-effort version capture; ignore errors.
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	//
+	// 10s (up from 3s) to keep the probe reliable when the machine is under
+	// parallel test load — `go test ./...` can stack 20+ concurrent fork+exec
+	// calls, and a 3s window was flaky on a developer laptop running the
+	// pre-push hook. This is a one-shot `--version` probe, not a hot path;
+	// the timeout is a floor on machine health, not on binary response time.
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 	out, err := exec.CommandContext(ctx, path, "--version").Output()
 	if err == nil {
--- a/internal/team/wiki_extractor_closure_test.go
+++ b/internal/team/wiki_extractor_closure_test.go
@@ -375,9 +375,16 @@ func countNonEmptyLines(t *testing.T, path string) int {
 // commitCountForPath returns the number of commits that touched relPath.
 // Uses `git log --oneline -- <path>` so the test does not depend on any
 // internal helper.
+//
+// Clean the subprocess env so the lookup isn't hijacked by an inherited
+// GIT_DIR — when tests run under a pre-push hook, git exports GIT_DIR
+// pointing at the outer repo and `git log` would then query the outer
+// history (zero commits touching wiki/facts/**) instead of this test's
+// fixture repo. Same pattern as internal/migration/writer_test.go.
 func commitCountForPath(t *testing.T, repoRoot, relPath string) int {
 	t.Helper()
 	cmd := exec.Command("git", "-C", repoRoot, "log", "--oneline", "--", relPath)
+	cmd.Env = GitCleanEnv()
 	out, err := cmd.CombinedOutput()
 	if err != nil {
 		t.Fatalf("git log %s: %v: %s", relPath, err, out)
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -29,6 +29,9 @@ pre-commit:
         - "dist/**"
         - "node_modules/**"
         - "bun.lock"
+        - "web/bun.lock"
+        - "package-lock.json"
+        - "web/package-lock.json"
         - "**/*.gen.ts"
         - "**/*.gen.tsx"
 
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -26,6 +26,7 @@
         "zustand": "^5.0.5"
       },
       "devDependencies": {
+        "@biomejs/biome": "2.4.13",
         "@testing-library/jest-dom": "^6.9.1",
         "@testing-library/react": "^16.3.2",
         "@testing-library/user-event": "^14.6.1",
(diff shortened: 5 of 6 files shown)


FranDias (commented):
Overall: five targeted fixes that together get `git push` green without `--no-verify`. The `GitCleanEnv` fix is correct and the audit for other offenders checks out — `worktree.go`, `human_identity.go`, `wiki_git.go`, `broker_test.go`, `broker_studio_test.go`, `wiki_git_test.go`, `migration/writer_test.go` all already route through `GitCleanEnv()`, so the extractor test was the last gap in production + test code. Lefthook biome exclude addition works, but leans partly on misdiagnosis (see inline). Package-lock resync is clean — 164 lines are exactly biome@2.4.13 + 8 platform deps, no unrelated drift.

Main concern is the onboarding timeout: the 3→10s bump masks a test-fixture issue and introduces a user-facing UX regression (server-side serial probes can outrun the wizard's 5s HTTP budget). See inline.

Also note the scratchpad `FIX_HOOKS_NOTES.md` is still uncommitted in the worktree — make sure it doesn't accidentally land in a subsequent commit.

internal/onboarding/prereqs.go
```diff
@@ -80,7 +80,13 @@ func CheckOne(name string) PrereqResult {
...
 
 	// Best-effort version capture; ignore errors.
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	//
+	// 10s (up from 3s) to keep the probe reliable when the machine is under
+	// parallel test load — `go test ./...` can stack 20+ concurrent fork+exec
+	// calls, and a 3s window was flaky on a developer laptop running the
+	// pre-push hook. This is a one-shot `--version` probe, not a hot path;
+	// the timeout is a floor on machine health, not on binary response time.
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
FranDias, line 103: **IMPORTANT** — this bump interacts badly with how `CheckAll` is consumed.

`CheckAll()` at `internal/onboarding/handlers.go:376` (via `HandlePrereqs`) calls `CheckOne` **serially** for 7 binaries (`node`, `git`, `claude`, `codex`, `opencode`, `cursor`, `windsurf`). With a 10s per-probe timeout, worst-case server latency is now **70 seconds**.

Meanwhile the wizard's HTTP client in `cmd/wuphf/onboarding.go:712` has a 5-second timeout:

```go
client := &http.Client{Timeout: 5 * time.Second}
```

On completion it falls back to `defaultPrereqs()` — which reports every binary as missing and flips the wizard into "continue unverified" state. One pathologically slow `--version` invocation (even at the old 3s budget) could already push past 5s; with 10s, two slow probes guarantees it. The web wizard at `web/src/components/onboarding/Wizard.tsx:1501` has its own fetch timeout too.

Three better patterns to consider:

1. Parallelize `CheckAll` with an `errgroup` / `sync.WaitGroup` — each probe already has its own context, the aggregate wall-clock drops back to `max(probes)` instead of `sum(probes)`.
2. Keep the aggregate budget tight (e.g. 3–4s) and short-circuit on first timeout — for a `--version` probe we don't actually need a clean version string, `Found=true` is enough.
3. If this really is only a test-load problem, fix it in the test (use a pre-known `.version` sidecar file, or mock `exec.CommandContext` via a seam) rather than relaxing production timing.

As-is, this is a latent user-facing regression hiding behind a test flake fix.

internal/onboarding/prereqs.go
```diff
@@ -80,7 +80,13 @@ func CheckOne(name string) PrereqResult {
...
 	defer cancel()
 	out, err := exec.CommandContext(ctx, path, "--version").Output()
```
FranDias, line 105: **SUGGESTION** — root cause of the test flake isn't addressed.

`TestCheckOneFindsOpencodeInCommonUserBinWhenPATHIsMinimal` (`internal/onboarding/prereqs_test.go:114`) forks a `#!/bin/sh` stub script that `printf`s a version string. That happy path is sub-second even on a heavily-loaded laptop — 3s should have been enough with orders of magnitude of headroom.

What's actually happening: under `go test ./...` the rest of the suite (internal/team in particular) is fork/exec-heavy, saturating the OS process table and making each new `fork+exec` latency-spike. The test itself is fine; its environment is polluted by siblings.

A 10s bump makes the test 3.3× more tolerant but doesn't fix the underlying coupling. Options:

- Make `CheckOne` accept an injectable exec function in tests, skip the real subprocess spawn.
- Have the test use `t.Parallel()` and document the load budget explicitly.
- Write a version sidecar (e.g. `opencode.version` next to the binary) and have `CheckOne` prefer reading that over forking `--version` — would also speed up real startup.

This one can stay as-is for now given the push-unblock urgency, but the scratchpad note ("the general fix is to stop hard-timing an env probe") deserves a follow-up issue, not a `FIX_HOOKS_NOTES.md` scratchpad mention.

lefthook.yml
```diff
@@ -29,6 +29,9 @@ pre-commit:
         - "dist/**"
         - "node_modules/**"
         - "bun.lock"
+        - "web/bun.lock"
+        - "package-lock.json"
```
FranDias, line 33: **SUGGESTION** — `package-lock.json` entry is a defensive no-op and `web/bun.lock` is likely unnecessary.

Per the [lefthook `root:` docs](https://lefthook.dev/configuration/root), globs (both `glob:` and `exclude:`) always match against repo-root-relative paths, never `root:`-relative paths. So:

- `package-lock.json` matches a root-level `package-lock.json` file. **No such file exists** in this repo. Defensive/harmless, but worth a comment or deletion.
- `web/bun.lock` is also redundant for biome specifically: biome's `glob: **/*.{js,ts,jsx,tsx,json,css}` doesn't match `.lock` extensions, so `bun.lock` / `web/bun.lock` never reach biome's file list in the first place. The only entry doing real work here is `web/package-lock.json` (the extension matches the glob and biome errors on it).
- The commit message for 1ca91ef6 frames this as "the existing `bun.lock` entry was a silent no-op because lefthook exclude is repo-root-relative." The no-op conclusion is right, but the reason is different — it's the `.lock` vs `.json/.ts/.css` extension mismatch, not the `root:`-scoping. (For reference, the secretlint block has no such `glob:` restriction, so its `bun.lock` + `web/bun.lock` entries ARE both load-bearing.)

Not blocking. Suggest trimming to:

```yaml
exclude:
  - "dist/**"
  - "node_modules/**"
  - "web/package-lock.json"   # biome errors with "no files processed" when only a lockfile is staged
  - "**/*.gen.ts"
  - "**/*.gen.tsx"
```

…with a one-line comment so the next reader doesn't chase the same phantom.

DEVELOPMENT.md
```diff
@@ -8,6 +8,10 @@ go build -o wuphf ./cmd/wuphf
 
 For normal app usage you do not need Bun. The local office/team MCP tools now run from the main Go binary through the hidden `wuphf mcp-team` subcommand.
 
+## Git hooks
+
+Hooks run via [lefthook](https://github.com/evilmartians/lefthook) (`lefthook.yml`). On `pre-push` the full Go test suite and a `cmd/wuphf` build are executed; on `pre-commit` gofmt/go vet/golangci-lint run against staged Go files. Do not push with `--no-verify` — fix the underlying failure instead.
```
FranDias, line 13: **SUGGESTION** — documentation omits most of what the pre-commit hook actually runs.

The line describes only gofmt/go-vet/golangci-lint on pre-commit, but `lefthook.yml` pre-commit has 7 commands (secretlint, biome, gofmt, go-vet, golangci-lint, no-secrets, check-merge-conflicts, no-large-files) plus a separate `commit-msg → commitlint` hook. The PR's stated goal is "eliminate reasons to reach for `--no-verify`" — the most likely new-contributor `--no-verify` driver is actually secretlint failing with `Failed to load rule module` when root-level `bun install` hasn't run (per the scratchpad, this catches everyone on a fresh clone).

Minor doc fix worth landing with this PR so the single authoritative "hooks" reference covers them all, e.g.:

```md
## Git hooks

Hooks run via [lefthook](https://github.com/evilmartians/lefthook) (`lefthook.yml`).

- `commit-msg`: commitlint (conventional commits format).
- `pre-commit`: secretlint, biome (web/**), gofmt, go vet, golangci-lint, merge-conflict markers, no-large-files (>5MB), grep-based no-secrets.
- `pre-push`: full `go test ./...`, `go build ./cmd/wuphf`, and `vhs testdata/vhs/check.sh` (if vhs is installed).

Before your first commit on a fresh clone, run `bun install` at the repo root — the pre-commit tooling (secretlint, commitlint, biome) is resolved via `bunx` from root node_modules.

Do not push with `--no-verify` — fix the underlying failure instead.
```

internal/team/wiki_extractor_closure_test.go
```diff
@@ -375,9 +375,16 @@ func countNonEmptyLines(t *testing.T, path string) int {
...
+//
+// Clean the subprocess env so the lookup isn't hijacked by an inherited
+// GIT_DIR — when tests run under a pre-push hook, git exports GIT_DIR
+// pointing at the outer repo and `git log` would then query the outer
+// history (zero commits touching wiki/facts/**) instead of this test's
+// fixture repo. Same pattern as internal/migration/writer_test.go.
 func commitCountForPath(t *testing.T, repoRoot, relPath string) int {
 	t.Helper()
 	cmd := exec.Command("git", "-C", repoRoot, "log", "--oneline", "--", relPath)
+	cmd.Env = GitCleanEnv()
```
FranDias, line 387: Fix is correct — confirms the last direct-`exec.Command("git", …)` gap in the tree. I grepped:

```
internal/team/worktree.go:177,244,256,422 — runGit/runGitOutput/gitRepoRoot/gitRefExists: GitCleanEnv ✓
internal/team/human_identity.go:283 — getGitGlobalConfig: GitCleanEnv ✓
internal/team/wiki_git.go:725 — runGitLocked: GitCleanEnv ✓ (append pattern)
internal/team/broker_test.go:152 — initUsableGitWorktree: GitCleanEnv ✓
internal/team/broker_studio_test.go:474 — test git init: GitCleanEnv ✓
internal/team/wiki_git_test.go:506,526,579 — all three: GitCleanEnv ✓
internal/migration/writer_test.go:285 — gitLog: team.GitCleanEnv ✓
internal/team/wiki_extractor_closure_test.go:386 — THIS PR ✓
```

No other callers. Worth noting in the CLAUDE.md / worktree.go godoc that `GitCleanEnv()` is mandatory for any direct `git` subprocess so the next contributor doesn't add a 10th offender.

PR #282, opened 2026-04-25
fix(team): unique tmp filename for broker state save + local e2e helper

FranDias:
## Summary

Two latent issues flagged after PR #281 merged:

1. **`broker.saveLocked` rename race** — the flake that hit PR #281's `test` job (`rename broker-state.json.tmp → broker-state.json: no such file or directory`).
2. **Smoke test seed assumption** — `web/e2e/tests/smoke.spec.ts` requires `~/.wuphf/onboarded.json` to be seeded; without it the test fails confusingly. CI handles this in shell; locally devs had to read the workflow YAML to figure out the contract.

Both fixed in a single PR — small surface, related to test-isolation, and the e2e helper benefits from being able to repro this exact race during local debugging.

## Commit 1 — `fix(team): unique tmp filename for broker state save`

`saveLocked` used a fixed `<path>.tmp` filename. Two brokers writing to the same `brokerStatePath()` could interleave:

\`\`\`
A.WriteFile(path.tmp, dataA)
B.WriteFile(path.tmp, dataB)   // clobbers A's bytes (harmless)
A.Rename(path.tmp, path)       // wins; path = dataB content
B.Rename(path.tmp, path)       // FAILS: path.tmp gone
\`\`\`

Production never hits this — one broker per path. **Tests do**: 22 `_test.go` files override `brokerStatePath` and may leak goroutines, plus a process-lifetime tempdir from `worktree_guard_test.go` `init()` is shared across every test that doesn't override.

Fix: factor out `atomicWriteFile()` that uses `os.CreateTemp` to generate a unique sibling tmp filename per save. Both the state file and the `.last-good` snapshot path go through the helper. Concurrent writers to the same destination can never race on the source of a rename.

Includes a deterministic regression test (`broker_save_race_test.go`) that runs 32 concurrent broker saves at the same path. **28/32 fail** against the prior code; **0/32** against the fix.

## Commit 2 — `chore(e2e): add run-local.sh + README`

Adds `web/e2e/run-local.sh`:
- Pins `WUPHF_RUNTIME_HOME` to a per-run tempdir, sandboxing all on-disk state away from the developer's real `~/.wuphf`.
- For the shell phase, seeds `<RUNTIME_HOME>/.wuphf/onboarded.json` (same JSON CI writes) before launching wuphf.
- Launches wuphf on `27890`/`27891` (overridable via `PORT=`) so it never collides with a developer's normally-running 7891 instance.
- Builds `web/dist` and the `wuphf` binary if missing.
- Runs both wizard and smoke specs end-to-end, with cleanup.

Adds `web/e2e/README.md` documenting the two phases, the seed precondition, and the script. Updates the playwright.config.ts comment to point at them.

## Test plan

- [x] `go test -count=1 ./internal/team/` — all pass (incl. new race repro)
- [x] `go test -count=1 ./...` — all 27 packages pass
- [x] `web/e2e/run-local.sh both` — 6/6 tests pass cleanly from a fresh clone
- [x] Pre-push hook green, no `--no-verify`
- [ ] CI fully green

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed race condition in broker state persistence that could cause failures under concurrent operations.

* **Tests**
  * Added concurrency regression test to verify broker state handling stability.

* **Documentation**
  * Added guide for running end-to-end smoke tests locally.

* **Chores**
  * Added automation script for local e2e test execution with proper sandboxing.
  * Updated test configuration and documentation.

### DIFF
--- a/internal/team/broker.go
+++ b/internal/team/broker.go
@@ -3179,11 +3179,11 @@ func (b *Broker) saveLocked() error {
 	if err != nil {
 		return err
 	}
-	if err := atomicWriteFile(path, data, 0o600); err != nil {
+	if err := atomicWriteFile(path, data); err != nil {
 		return err
 	}
 	if brokerStateShouldSnapshot(state) {
-		if err := atomicWriteFile(snapshotPath, data, 0o600); err != nil {
+		if err := atomicWriteFile(snapshotPath, data); err != nil {
 			return err
 		}
 	}
@@ -3191,9 +3191,9 @@ func (b *Broker) saveLocked() error {
 }
 
 // atomicWriteFile writes data to path atomically by creating a uniquely-named
-// sibling tmp file and renaming. Each call uses a fresh tmp filename via
-// os.CreateTemp, so concurrent writers to the same destination cannot race
-// on the source path of the rename.
+// sibling tmp file (mode 0o600 via os.CreateTemp) and renaming. Each call
+// uses a fresh tmp filename, so concurrent writers to the same destination
+// cannot race on the source path of the rename.
 //
 // The previous fixed `<path>.tmp` filename was safe in production (one broker
 // owns one path) but broke the test suite: 22 *_test.go files override
@@ -3204,7 +3204,11 @@ func (b *Broker) saveLocked() error {
 // A had already renamed the shared tmp out from under it. That was the CI
 // flake on PR #281's `test` job. See broker_save_race_test.go for the
 // regression repro.
-func atomicWriteFile(path string, data []byte, perm os.FileMode) error {
+//
+// 0o600 is hard-coded because the only callers (broker state file +
+// snapshot) want exactly that mode; CreateTemp already produces it on the
+// platforms we support, so no os.Chmod is needed.
+func atomicWriteFile(path string, data []byte) error {
 	dir := filepath.Dir(path)
 	base := filepath.Base(path)
 	tmpf, err := os.CreateTemp(dir, base+".*.tmp")
@@ -3219,13 +3223,6 @@ func atomicWriteFile(path string, data []byte, perm os.FileMode) error {
 		return err
 	}
 	if err := tmpf.Close(); err != nil {
-		cleanup()
-		return err
-	}
-	// os.CreateTemp creates with mode 0o600 already, but tighten / loosen
-	// per the caller's request so this helper is interchangeable with the
-	// previous os.WriteFile call.
-	if err := os.Chmod(tmpName, perm); err != nil {
 		cleanup()
 		return err
 	}
--- a/internal/team/broker.go
+++ b/internal/team/broker.go
@@ -3179,21 +3179,59 @@ func (b *Broker) saveLocked() error {
 	if err != nil {
 		return err
 	}
-	tmp := path + ".tmp"
-	if err := os.WriteFile(tmp, data, 0o600); err != nil {
-		return err
-	}
-	if err := os.Rename(tmp, path); err != nil {
+	if err := atomicWriteFile(path, data, 0o600); err != nil {
 		return err
 	}
 	if brokerStateShouldSnapshot(state) {
-		snapshotTmp := snapshotPath + ".tmp"
-		if err := os.WriteFile(snapshotTmp, data, 0o600); err != nil {
+		if err := atomicWriteFile(snapshotPath, data, 0o600); err != nil {
 			return err
 		}
-		if err := os.Rename(snapshotTmp, snapshotPath); err != nil {
-			return err
-		}
+	}
+	return nil
+}
+
+// atomicWriteFile writes data to path atomically by creating a uniquely-named
+// sibling tmp file and renaming. Each call uses a fresh tmp filename via
+// os.CreateTemp, so concurrent writers to the same destination cannot race
+// on the source path of the rename.
+//
+// The previous fixed `<path>.tmp` filename was safe in production (one broker
+// owns one path) but broke the test suite: 22 *_test.go files override
+// brokerStatePath, plus a leaked tempdir from worktree_guard_test.go init()
+// is shared across every test that doesn't override. Two saves landing on
+// the same path could interleave like A.WriteFile / B.WriteFile / A.Rename /
+// B.Rename — and B's Rename failed with "no such file or directory" because
+// A had already renamed the shared tmp out from under it. That was the CI
+// flake on PR #281's `test` job. See broker_save_race_test.go for the
+// regression repro.
+func atomicWriteFile(path string, data []byte, perm os.FileMode) error {
+	dir := filepath.Dir(path)
+	base := filepath.Base(path)
+	tmpf, err := os.CreateTemp(dir, base+".*.tmp")
+	if err != nil {
+		return err
+	}
+	tmpName := tmpf.Name()
+	cleanup := func() { _ = os.Remove(tmpName) }
+	if _, err := tmpf.Write(data); err != nil {
+		_ = tmpf.Close()
+		cleanup()
+		return err
+	}
+	if err := tmpf.Close(); err != nil {
+		cleanup()
+		return err
+	}
+	// os.CreateTemp creates with mode 0o600 already, but tighten / loosen
+	// per the caller's request so this helper is interchangeable with the
+	// previous os.WriteFile call.
+	if err := os.Chmod(tmpName, perm); err != nil {
+		cleanup()
+		return err
+	}
+	if err := os.Rename(tmpName, path); err != nil {
+		cleanup()
+		return err
 	}
 	return nil
 }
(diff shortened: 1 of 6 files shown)


internal/team/broker.go
```diff
@@ -3179,25 +3179,63 @@ func (b *Broker) saveLocked() error {
...
+		return err
+	}
+	if err := tmpf.Close(); err != nil {
+		cleanup()
+		return err
+	}
+	// os.CreateTemp creates with mode 0o600 already, but tighten / loosen
+	// per the caller's request so this helper is interchangeable with the
+	// previous os.WriteFile call.
+	if err := os.Chmod(tmpName, perm); err != nil {
```
FranDias: **SUGGESTION** — The `os.Chmod` is effectively dead code for both current callers (saveLocked passes `0o600`, which is exactly what `os.CreateTemp` already gives you). Worth either dropping it (leanest) or keeping it but shortening the comment — the `tighten / loosen` framing suggests there's a meaningful difference, but no caller in the tree actually exercises it.

If you keep it, fine — but if a future caller passes `0o644` expecting world-readable output, the brief Chmod-then-Rename window means a reader could open the file at `path` post-rename and see 0o600 if Chmod hadn't applied yet. That's not a real concern today (no such caller), but it's worth pinning down: pass `perm` to `os.OpenFile`-equivalent semantics by chmodding *before* Write, or just drop the parameter.

FranDias (commented):
Two-part fix looks solid. Race repro is deterministic and the helper is correct end-to-end. A few notes inline; nothing blocking.

Verified:
- `os.CreateTemp` opens the file → `Write → Close → Chmod → Rename` ordering is right.
- Cleanup-on-error covers every failure path (Write/Close/Chmod/Rename), so no tmp leak on the hot paths.
- `WUPHF_RUNTIME_HOME` flows through to `internal/config.RuntimeHomeDir()` and into `onboarded.json` / `broker-state.json` paths — sandbox is real.
- `--no-open --broker-port --web-port --no-nex` all exist in `cmd/wuphf/main.go`.
- README path claim (`~/.wuphf/team/broker-state.json`) matches `defaultBrokerStatePath`.
- All required checks (test, build, web-e2e, lint, web, vhs, shellcheck, scan, release-build) green.

internal/team/broker_save_race_test.go
```diff
@@ -0,0 +1,93 @@
...
+	errs := make(chan error, goroutines)
+	wg.Add(goroutines)
+	for i := 0; i < goroutines; i++ {
+		go func(i int) {
+			defer wg.Done()
+			b := NewBroker()
+			// EnsurePlannedTask calls saveLocked synchronously under b.mu —
+			// each broker serializes its own saves, but two brokers do not
+			// share a mutex, so they race on the on-disk tmp filename.
+			_, _, err := b.EnsurePlannedTask(plannedTaskInput{
```
FranDias, line 54: **SUGGESTION** — The repro uses `EnsurePlannedTask`, which is a heavy path (entity/task/state mutation, not just a save). It works — you've shown 28/32 fails against the old code — but a more surgical repro would invoke the rename race directly:

```go
b := NewBroker()
// ... seed a minimal state ...
for i := 0; i < goroutines; i++ {
    go func() { defer wg.Done(); b.mu.Lock(); err := b.saveLocked(); b.mu.Unlock(); ... }()
}
```

Upsides: faster on slow runners (this currently does ~32 full broker bootstraps), tighter blast radius if EnsurePlannedTask grows new failure modes, and the assertion stays a clean "only the rename can fail". Not blocking — current shape is fine and the failure-mode assertion (`expected zero errors` + filename surfacing) catches the rename signature loud enough.

web/e2e/run-local.sh
```diff
@@ -0,0 +1,131 @@
...
+  (cd web && bun install --frozen-lockfile && bun run build)
+fi
+
+if [ ! -x ./wuphf ]; then
+  echo "[run-local] building wuphf binary"
+  go build -o wuphf ./cmd/wuphf
+fi
+
+echo "[run-local] installing e2e deps"
+(cd web/e2e && bun install --frozen-lockfile >/dev/null)
```
FranDias: **NIT** — `bun install --frozen-lockfile` runs unconditionally on every invocation, even when `node_modules` is already populated. Bun is fast on the no-op path, but this is wall-clock noise on a script devs run repeatedly. Consider gating on `[ ! -d node_modules ]` like you do for `web/dist` and `./wuphf`:

```bash
if [ ! -d web/e2e/node_modules ]; then
  echo "[run-local] installing e2e deps"
  (cd web/e2e && bun install --frozen-lockfile >/dev/null)
fi
```

Same for the `bunx playwright install chromium` line below — that one is even slower on a cold cache.

web/e2e/run-local.sh
```diff
@@ -0,0 +1,131 @@
...
+
+stop_wuphf() {
+  if [ -n "${pid:-}" ] && kill -0 "$pid" 2>/dev/null; then
+    kill "$pid" 2>/dev/null || true
+    wait "$pid" 2>/dev/null || true
+    pid=""
+  fi
+  # Wait for the port to free up so the next phase can rebind.
+  for _ in $(seq 1 10); do
+    if ! (exec 3<>/dev/tcp/127.0.0.1/"$web_port") 2>/dev/null; then break; fi
```
FranDias, line 122: **NIT** — The `(exec 3<>/dev/tcp/127.0.0.1/...)` trick works on Linux and on macOS bash, but `/dev/tcp` is a *bash* feature, not a kernel device. Confirmed it works under `set -euo pipefail` because the `2>/dev/null` swallows the connect failure and the subshell short-circuits. Worth a one-line comment so the next reader doesn't reach for `nc -z` thinking it's broken — and so anyone tempted to switch the shebang to `#!/bin/sh` knows why they can't.

No functional fix needed; the shebang is `#!/usr/bin/env bash` and CI uses the same trick at `.github/workflows/ci.yml:308`.

web/e2e/run-local.sh
```diff
@@ -0,0 +1,131 @@
...
+# Sandbox: every wuphf state file lands under this throwaway dir, so a
+# developer running this script never has their real onboarded.json or
+# broker-state.json mutated.
+runtime_home="$(mktemp -d -t wuphf-e2e-runtime-XXXXXX)"
+echo "[run-local] sandboxed WUPHF_RUNTIME_HOME=${runtime_home}"
+
+cleanup() {
+  if [ -n "${pid:-}" ] && kill -0 "$pid" 2>/dev/null; then
+    kill "$pid" 2>/dev/null || true
+    wait "$pid" 2>/dev/null || true
```
FranDias: **SUGGESTION** — `kill "$pid"` sends SIGTERM. If wuphf hangs on shutdown (e.g. a stuck broker goroutine), `wait` blocks indefinitely and the trap never returns, leaving `runtime_home` undeleted and the next phase's port-free loop spinning. Consider a bounded fallback:

```bash
cleanup() {
  if [ -n "${pid:-}" ] && kill -0 "$pid" 2>/dev/null; then
    kill "$pid" 2>/dev/null || true
    for _ in $(seq 1 5); do
      kill -0 "$pid" 2>/dev/null || break
      sleep 1
    done
    kill -9 "$pid" 2>/dev/null || true
    wait "$pid" 2>/dev/null || true
  fi
  rm -rf "$runtime_home"
}
```

Not blocking — wuphf exits cleanly today — but cheap insurance for a script devs invoke after CI failures, where a flaky shutdown shouldn't compound.

internal/team/broker.go
```diff
@@ -3179,25 +3179,63 @@ func (b *Broker) saveLocked() error {
...
+// is shared across every test that doesn't override. Two saves landing on
+// the same path could interleave like A.WriteFile / B.WriteFile / A.Rename /
+// B.Rename — and B's Rename failed with "no such file or directory" because
+// A had already renamed the shared tmp out from under it. That was the CI
+// flake on PR #281's `test` job. See broker_save_race_test.go for the
+// regression repro.
+func atomicWriteFile(path string, data []byte, perm os.FileMode) error {
+	dir := filepath.Dir(path)
+	base := filepath.Base(path)
+	tmpf, err := os.CreateTemp(dir, base+".*.tmp")
```
FranDias, line 3214: **FYI / report-only (per review brief)** — Same `os.WriteFile(<path>.tmp) + os.Rename(...)` pattern still lives in two other call sites that did NOT adopt `atomicWriteFile`:

- `internal/team/scanner_manifest.go:85-90` — `WriteScanManifest` is mutex-guarded and currently single-writer per process, so no race in production. Tests don't appear to fan out across this path the way they do across `brokerStatePath`.
- `internal/team/human_identity.go:240-245` — `persist()` short-circuits on `os.Stat(path) == nil`, so concurrent calls for the same email skip the write entirely. Same single-writer-by-construction story.

Neither is at risk today, but if/when test isolation hits the same shape (multiple tests sharing an HTTP test server's runtime home, say) the same flake will reappear. Worth keeping `atomicWriteFile` exported within the package and reaching for it next time. (Other sites — `internal/team/headless_opencode.go`, `internal/onboarding/state.go`, `internal/operations/wiki_materialize.go` — already use unique tmp names.)

PR #290, opened 2026-04-25
ci(release): gate npm publish behind environment + scope per-job permissions

FranDias:
## Summary

Closes the audit's "release.yml workflow_dispatch no env gate" 🔴 from
[NEX_CRM_REPO_AUDIT_2026_04_24.md](../blob/main/NEX_CRM_REPO_AUDIT_2026_04_24.md) Table E row `wuphf`. Single file changed.

After staff-review (commit 4a22dd6f), publish-npm is **inlined into the `release` job** so the env gate produces an atomic outcome — either GitHub Release + npm both ship, or neither does. A separate `publish-npm` job would partial-fail (GH Release published, npm gate timed out) and require manual cleanup.

Final shape:
- Workflow default: `permissions: contents: read`.
- `concurrency: { group: release-${{ github.ref }}, cancel-in-progress: false }` — no half-finished release on a re-dispatch.
- `release` job: env-gated to `npm-publish`, `contents: write`, runs goreleaser + `npm publish` in sequence.
- `smoke-test-npx`: read-only, runs after `release` completes; verifies the binary's self-reported version matches the tag on Linux + macOS.

## Follow-up (admin steps, not in this PR — required for merge to be safe)

For the env gate to actually bind:

1. Create environment `npm-publish` in repo Settings → Environments.
2. Add a **required reviewer** rule (default to `@FranDias`).
3. Add a **deployment branch / tag policy**: protected refs → tag pattern `v*`. Without this, a `workflow_dispatch` against an arbitrary ref would still trigger the gate.
4. Add `NPM_TOKEN` as an **environment-scoped** secret on `npm-publish`.
5. **Delete the repo-scoped `NPM_TOKEN`** — otherwise any other workflow can still read it and the env gate is decorative.

The workflow references the env, so the gate is structural the moment this PR merges. The protection rules need a UI click before the next release fires.

## Future work (not in this PR)

- npm provenance: add `id-token: write` to the `release` job's `permissions:` block once we have a green tag through this gate. That gives the npm package a build attestation users can verify.

## Test plan

- [x] Lefthook pre-push (secretlint + commitlint + go test + build) green locally.
- [x] CI green on initial push.
- [ ] Re-verify CI green after the inline restructure (4a22dd6f).
- [ ] Post-merge: create env + reviewer rule + tag policy + move NPM_TOKEN + delete repo-scoped NPM_TOKEN. Next tagged release verifies the gate fires correctly and atomically.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,12 +9,17 @@ on:
   # GITHUB_TOKEN does not fire other workflows, but workflow_dispatch does.
   workflow_dispatch:
 
+# Default to least-privilege; per-job blocks below add only what each job
+# actually needs. The release job uploads artifacts (contents: write); npm
+# publish + npx smoke only need to read the repo.
 permissions:
-  contents: write
+  contents: read
 
 jobs:
   release:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     outputs:
       version: ${{ steps.version.outputs.version }}
     steps:
@@ -68,6 +73,16 @@ jobs:
   publish-npm:
     needs: release
     runs-on: ubuntu-latest
+    # GitHub Environment gate: protection rules (required reviewer +
+    # branch policy `v*` tags) live on the env, so any future change to
+    # who/what can publish to npm is configured in one place rather than
+    # scattered across the workflow. NPM_TOKEN is bound to this env, not
+    # the repo, so a fork PR cannot reach it even via workflow_dispatch.
+    environment:
+      name: npm-publish
+      url: https://www.npmjs.com/package/wuphf
+    permissions:
+      contents: read
     env:
       VERSION: ${{ needs.release.outputs.version }}
     steps:
@@ -95,6 +110,8 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
     env:
       VERSION: ${{ needs.release.outputs.version }}
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,9 +15,25 @@ on:
 permissions:
   contents: read
 
+# Don't cancel an in-flight release: a partially-finished goreleaser leaves
+# orphan release assets, and a partial npm publish leaves a half-shipped
+# version. Two concurrent releases for the same tag can't be valid anyway —
+# tags are immutable; the second invocation has nothing new to ship.
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: false
+
 jobs:
   release:
     runs-on: ubuntu-latest
+    # Env gate sits on the FIRST job that ships anything (goreleaser uploads
+    # GitHub Release artifacts). Putting it on `publish-npm` instead would
+    # leave a published GitHub Release with no corresponding npm package on
+    # rejection, requiring manual cleanup. Pausing here means the human
+    # approves once and both surfaces ship together — or nothing ships.
+    environment:
+      name: npm-publish
+      url: https://www.npmjs.com/package/wuphf
     permissions:
       contents: write
     outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -68,6 +73,16 @@ jobs:
   publish-npm:
     needs: release
     runs-on: ubuntu-latest
+    # GitHub Environment gate: protection rules (required reviewer +
+    # branch policy `v*` tags) live on the env, so any future change to
+    # who/what can publish to npm is configured in one place rather than
+    # scattered across the workflow. NPM_TOKEN is bound to this env, not
+    # the repo, so a fork PR cannot reach it even via workflow_dispatch.
+    environment:
+      name: npm-publish
```
FranDias: **IMPORTANT — partial-release failure mode introduced by gating only `publish-npm`**

The `release` job (which uploads the GitHub Release via goreleaser) runs *before* the env approval gate on `publish-npm`. If the required reviewer rejects, or the approval times out, you end up with:

- A published GitHub Release at the tag (assets uploaded), and
- No corresponding `wuphf@<version>` on npm.

`auto-release.yml` already pushed the tag and the GitHub Release exists, so re-running won't auto-recover — recovery requires manually re-dispatching `release.yml` and re-approving. Worse, the second `release` job will hit goreleaser's "release already exists" path unless it's run as `--clean` against the same tag.

Two options worth considering:

1. **Move the env gate up** to also cover the `release` job, so a rejected approval = nothing ships at all. Drawback: every release stalls on a human even before goreleaser. Probably what you want for a public CLI.
2. **Split publish-npm into a separately re-runnable workflow** (or job) keyed off `workflow_run` of release.yml so you can re-trigger just the npm publish without rebuilding goreleaser artifacts.

Not blocking — the current setup is still strictly safer than today — but worth choosing intentionally now rather than discovering it the first time someone rejects.

.github/workflows/release.yml
```diff
@@ -68,6 +73,16 @@ jobs:
   publish-npm:
     needs: release
     runs-on: ubuntu-latest
+    # GitHub Environment gate: protection rules (required reviewer +
+    # branch policy `v*` tags) live on the env, so any future change to
+    # who/what can publish to npm is configured in one place rather than
+    # scattered across the workflow. NPM_TOKEN is bound to this env, not
+    # the repo, so a fork PR cannot reach it even via workflow_dispatch.
```
FranDias: **SUGGESTION — env must exist before next merged auto-release, or auto-release breaks**

The PR body acknowledges this ("protection rules need a UI click"), but it's worth flagging the merge ordering: as soon as this lands on `main`, the very next `auto-release.yml` run will dispatch `release.yml`, the `release` job will succeed and tag a GitHub Release, then `publish-npm` will fail with "environment `npm-publish` not found".

Cheapest fix is to create the `npm-publish` env in repo settings *before* merging this PR (env can exist with no protection rules and no secrets — it just needs to resolve). Then layer on:

1. Required reviewer
2. Deployment branch policy: tags matching `v*`
3. Move `NPM_TOKEN` to env-scoped secret **and delete the repo-scoped one** (otherwise a future workflow that doesn't reference the env can still read it, defeating the gate).

Worth adding the "delete repo-scoped NPM_TOKEN" step to the PR body checklist explicitly — it's the part that actually closes the supply-chain hole, not the env binding.

.github/workflows/release.yml
```diff
@@ -9,12 +9,17 @@ on:
   # GITHUB_TOKEN does not fire other workflows, but workflow_dispatch does.
   workflow_dispatch:
 
+# Default to least-privilege; per-job blocks below add only what each job
+# actually needs. The release job uploads artifacts (contents: write); npm
```
FranDias, line 13: **SUGGESTION — workflow-level `contents: read` is correct, but think about future OIDC**

The job-level `permissions:` block *replaces* (not merges with) the workflow-level — so the `release` job below gets exactly `{contents: write}`, which is what goreleaser needs to upload release artifacts. Good.

One forward-looking note: if you ever turn on **npm provenance** (which you should — it's free supply-chain attestation for `wuphf@<version>`), you'll need `id-token: write` on the `publish-npm` job. Add it there, not at the workflow level. The current pattern of declaring per-job permissions makes that easy.

Nothing to change here; just leaving a breadcrumb for the follow-up.

.github/workflows/release.yml
```diff
@@ -9,12 +9,17 @@ on:
   # GITHUB_TOKEN does not fire other workflows, but workflow_dispatch does.
   workflow_dispatch:
```
FranDias, line 10: **SUGGESTION — no `concurrency:` block; tag-push and dispatch can race**

Not introduced by this PR, but the env gate amplifies it: if a human pushes `v1.2.3` directly *and* `auto-release.yml` happens to dispatch the same tag (or a manual re-dispatch lands), two `release` runs queue and the second will fail at either goreleaser ("release already exists") or npm publish ("cannot republish version"). With the env gate added, the second one might also sit in approval queue indefinitely, blocking the queue.

Worth adding:

```yaml
concurrency:
  group: release-${{ github.ref }}
  cancel-in-progress: false
```

Keyed by ref (the tag), so two tags publish in parallel, but the same tag serializes.

FranDias:
## Review summary — PR #290 (release env gate)

Approving in spirit; would request a couple of edits before merge.

**What I verified**
- Workflow-level `permissions: contents: read` + per-job re-elevation is the correct GitHub pattern; job-level `permissions:` block *replaces* (not merges with) the workflow level, so the `release` job gets exactly `{contents: write}`. Goreleaser's release upload needs that — verified the goreleaser-action invocation already sets `GITHUB_TOKEN: secrets.GITHUB_TOKEN`.
- `auto-release.yml` dispatches `release.yml` with `--ref "$VERSION"` where `$VERSION` is `v$MAJOR.$MINOR.$PATCH`, so a deployment-branch policy of "tags matching `v*`" on the `npm-publish` env will permit the dispatch path. ✓
- `workflow_dispatch` from forks is **not** something GitHub permits — only from the base repo with `actions: write` — so the env binding for `NPM_TOKEN` is sufficient against fork-PR exfiltration. ✓
- `release` job no longer holds workflow-wide `contents: write`, so `publish-npm` and `smoke-test-npx` cannot inadvertently push or attach assets. ✓

**Inline issues** (4 — see comments above)
1. Partial-release failure mode: `release` job ships GitHub Release before env approval gate fires on `publish-npm`. Rejected approval = half-released version. (IMPORTANT, design-level — choose intentionally now.)
2. `npm-publish` env must exist *before* this PR merges, otherwise the next auto-release breaks. Also: must delete repo-scoped `NPM_TOKEN` after moving it to env, or the gate is decorative. (SUGGESTION — add to PR body checklist.)
3. Forward-looking: `id-token: write` on `publish-npm` if you ever turn on npm provenance. (SUGGESTION — no change needed now.)
4. No `concurrency:` block on `release.yml` — tag push and dispatch can race. (SUGGESTION.)

**Nits**
- Line 48 still has `# v2.2.0` (bare format). PR body acknowledges this is held for #291 follow-up — fine, just don't forget.
- Commit message body starts with `…issions` (truncation artifact from rebasing/squashing); will look odd in `git log`. Worth a `git commit --amend` to clean, since this PR is single-commit.

Net: I'd merge after fixing items 1 and 2 (or explicitly choosing to live with 1).

— staff-code-reviewer

PR #291, opened 2026-04-25
ci: standardize ratchet pin comments across workflows

FranDias:
## Summary

`ratchet update` only re-resolves pins whose trailing comment matches `# ratchet:<owner>/<repo>@<ref>`. Mixed-format files mean future bumps silently skip half the actions.

This PR converts every remaining bare-version comment so ratchet tooling treats every pin uniformly.

Files touched:
- `.github/workflows/approve-on-comment.yml` (1)
- `.github/workflows/commitlint.yml` (2)
- `.github/workflows/ratchet-check.yml` (2)
- `.github/workflows/secret-scan.yml` (3)
- `.github/workflows/ci.yml` (4)

`release.yml` has one remaining bare-version line — left for a follow-up to avoid conflict with PR #290 (release env gate), which is already in flight against the same file.

## Note: version-target precision changes (per staff review)

Three lines silently shift their `@<tag>` major/minor while standardizing the comment, all aligning with how `ci.yml` already pins the same actions in this repo:

| Workflow | Action | Before | After |
|---|---|---|---|
| `secret-scan.yml` | `oven-sh/setup-bun` | `@v2` | `@v2.2.0` (narrower) |
| `commitlint.yml` | `actions/checkout` | `@v6.0.2` | `@v6` (broader) |
| `ratchet-check.yml` | `actions/setup-go` | `@v6.0.0` | `@v6` (broader) |

Net effect is repo-wide normalization to the form `ci.yml` already uses. SHAs are unchanged — the trailing comment only documents what the SHA is *expected* to point at; the SHA is what GitHub Actions actually resolves.

## Test plan

- [x] `ratchet lint .github/workflows/*.yml` exits 0 locally.
- [x] CI green.

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -27,13 +27,13 @@ jobs:
   commitlint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # ratchet:actions/checkout@v6
         with:
           # Need history back to the merge base so commitlint can lint
           # every commit on the PR, not just the head.
           fetch-depth: 0
 
-      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # ratchet:oven-sh/setup-bun@v2.2.0
         with:
           bun-version: "1.3.10"
 
--- a/.github/workflows/ratchet-check.yml
+++ b/.github/workflows/ratchet-check.yml
@@ -33,9 +33,8 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-
-      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # ratchet:actions/checkout@v6
+      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # ratchet:actions/setup-go@v6
         with:
           # No go.mod-version mode: ratchet has its own go.mod, we just
           # need any modern Go to build it. Match the major used elsewhere
--- a/.github/workflows/secret-scan.yml
+++ b/.github/workflows/secret-scan.yml
@@ -38,17 +38,17 @@ jobs:
       contents: read
     steps:
       # Pin to full SHA: third-party action runs on every PR.
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # ratchet:actions/checkout@v6
         with:
           fetch-depth: 0
 
-      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # ratchet:oven-sh/setup-bun@v2.2.0
         with:
           bun-version: "1"
 
       - name: Cache Bun dependencies
         # Pin to full SHA: third-party action, same threat model as checkout.
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v4.2.3
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # ratchet:actions/cache@v4.2.3
         with:
           path: ~/.bun/install/cache
           key: bun-${{ runner.os }}-${{ hashFiles('bun.lock') }}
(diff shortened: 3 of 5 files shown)


.github/workflows/secret-scan.yml
```diff
@@ -38,17 +38,17 @@ jobs:
...
       # Pin to full SHA: third-party action runs on every PR.
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # ratchet:actions/checkout@v6
         with:
           fetch-depth: 0
 
-      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # ratchet:oven-sh/setup-bun@v2.2.0
         with:
           bun-version: "1"
```
FranDias, line 47: **SUGGESTION — this is a *precision* change, not just a comment-format change**

The PR description frames this as "converts every remaining bare-version comment so ratchet tooling treats every pin uniformly," but `# v2` → `# ratchet:oven-sh/setup-bun@v2.2.0` is **narrower than the original intent**. The previous comment said "this SHA is some v2.x of setup-bun" (matching whatever floats `v2`); the new comment says "this SHA is exactly v2.2.0."

`ratchet update` consequence:
- Before: ratchet skipped this line entirely (bare-version format).
- After: ratchet will refresh the SHA only if `v2.2.0` itself is retagged (which would actually be the supply-chain attack scenario you're guarding against).

This is probably fine — the other workflows (`ci.yml`, `commitlint.yml`) already pin setup-bun to `v2.2.0` explicitly, so this normalizes precision across the repo. But worth one extra sentence in the PR body that this also tightens the version target, not just the comment format. (Alternative: keep it floating with `# ratchet:oven-sh/setup-bun@v2` to preserve the original "any v2.x" intent — but that's inconsistent with the other files.)

.github/workflows/commitlint.yml
```diff
@@ -27,13 +27,13 @@ jobs:
   commitlint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # ratchet:actions/checkout@v6
```
FranDias, line 30: **SUGGESTION — `v6.0.2` → `v6` *broadens* the ratchet update target**

Mirror of the secret-scan.yml comment but in the opposite direction: the previous comment pinned to a specific patch (`v6.0.2`); the new comment uses the floating major (`v6`).

Mechanically: a future `ratchet update` will now move this SHA every time `actions/checkout` retags `v6` to a new minor/patch — that's the same upstream-trust posture used in `ci.yml` (which also uses `# ratchet:actions/checkout@v6`), so this PR is normalizing precision across files. Consistent with the rest of the repo, just worth flagging that the *update behavior* changed in addition to the comment format.

.github/workflows/ratchet-check.yml
```diff
@@ -33,9 +33,8 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-
-      - uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # ratchet:actions/checkout@v6
+      - uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # ratchet:actions/setup-go@v6
```
FranDias: **SUGGESTION — same precision broadening as commitlint.yml**

`# v6.0.0` → `# ratchet:actions/setup-go@v6` floats the update target from a specific patch to the v6 major. Consistent with how `ci.yml` pins setup-go (also `# ratchet:actions/setup-go@v6`), so net effect is normalization, but the change in `ratchet update` semantics is real and worth noting in the PR description for future archeology.

FranDias:
## Review summary — PR #291 (ratchet pin comment normalization)

Approving. Mechanically clean, no behavior risk for CI runs themselves; the only thing worth noting is a precision drift that the PR description doesn't fully capture.

**What I verified**
- Every `uses:` line under `.github/workflows/` (excluding `release.yml` line 48 — explicitly held for the #290 follow-up per PR body) is now in the canonical `# ratchet:<owner>/<repo>@<ref>` form.
- GitHub Actions runtime ignores the trailing comment entirely — pin resolution uses only the `@<sha>` portion. So this PR is **runtime-inert**: no workflow behavior changes for any current CI run. ✓
- `ratchet check` passes either form (the bare `# v6.0.2` is a valid tag-comment ratchet recognizes for *check* purposes). The benefit is `ratchet update`, which only re-resolves canonical-format pins. ✓

**Inline issues** (3 — see comments above; all SUGGESTION-level)
- The PR mixes two distinct kinds of normalization in one diff:
  1. **Format-only** (e.g. `# v9.2.0` → `# ratchet:golangci/golangci-lint-action@v9.2.0`) — same precision, just canonical form. Most lines.
  2. **Precision change** — three lines also shift the version target:
     - `secret-scan.yml` setup-bun: `v2` → `v2.2.0` (narrower).
     - `commitlint.yml` checkout: `v6.0.2` → `v6` (broader).
     - `ratchet-check.yml` setup-go: `v6.0.0` → `v6` (broader).
  All three precision changes match the existing pattern in `ci.yml`, so the net effect is normalization across files. But the PR body ("converts every remaining bare-version comment so ratchet tooling treats every pin uniformly") undersells it — future archeology will be easier with one sentence in the description acknowledging that `ratchet update` semantics shift on those three lines.

**Nits**
- None blocking. Commit is clean, single-purpose, reversible.

Net: I'd merge as-is; consider amending the PR body to call out the three precision changes for the audit trail.

— staff-code-reviewer

PR #293, opened 2026-04-25
fix(team): make skill creation deterministic for all agents

najmuzzaman-mohammad:
## Summary

- Add structured `team_skill_create` MCP tool in 1:1, DM, and office server paths.
- Allow any registered agent to propose skills with `action=propose`; keep immediate active creation restricted to CEO with `action=create`.
- Require `action` explicitly and validate proposal authors so requests cannot silently fall back to the wrong path.
- Remove the legacy `[SKILL PROPOSAL]` parser and related prompt/test coverage.
- Isolate broker state in tests so persisted `broker-state.json` cannot leak across cases.

## Validation

- [x] `go test ./internal/team/...` — passed
- [x] `go test ./internal/teammcp/...` — passed
- [x] `go build ./...` — passed
- [x] `npm --prefix web run build` — passed; Vite emitted only the existing large-chunk warning
- [x] Browser smoke on `http://localhost:8000` — app rendered from built assets, Skills app opened, `/api/skills` returned 200, no console errors or failed browser requests


FranDias (approved):

PR #294, opened 2026-04-25
[codex] Add self-healing incident tasks

najmuzzaman-mohammad:
## Summary

Adds self-healing for both runtime failures and real capability gaps.

- creates or reuses a `Self-heal @agent on task` incident when an agent escalates, exhausts timeout/error recovery, or blocks a task because a required capability is missing
- detects capability-gap blockers from the actual task-block workflow (`team_task action=block` / broker `/tasks`) such as missing tools, integrations, skills, channels, providers, credentials, or access
- adds `team_task action=resume` so the repair lane can unblock and retry the original workflow after the missing capability is fixed
- records repeated failure signatures on the existing incident task instead of dropping later context
- prevents nested self-healing tasks and keeps repairs routed through office task state

## Why

The goal is not just recovering from errors. The system needs to notice “I cannot do this because a capability is missing,” make that missing capability the next real work item, repair it, and then resume the original workflow. The deeper Hermes and Browser Harness review pointed to a bounded approach: recover locally when possible, preserve failure context, and only turn a repair into durable procedure after the workaround proves reusable.

## Real Scenario Tested

Ran the capability-gap workflow through a live broker HTTP server, not a mocked function call:

1. Created a real office task: “Post client launch update to Slack.”
2. Simulated the owning agent blocking it with: “Unable to continue: missing Slack integration tool path...”
3. Verified the original task became `blocked`.
4. Verified the system created `Self-heal @eng on <task>` with trigger `capability_gap`.
5. Simulated the repair lane resuming the original task after the Slack capability became available.
6. Verified the original task returned to `in_progress` and retained both the original blocker and repair note.

## Validation

- `go test ./internal/team -run 'TestBrokerHandlePostTaskCapabilityGapCreatesSelfHealingTask|TestBrokerHandlePostTaskResumeUnblocksAfterCapabilityRepair' -count=1 -v`
- `go test ./internal/team -run 'TestBrokerHandlePostTaskCapabilityGapCreatesSelfHealingTask|TestBrokerHandlePostTaskNonCapabilityBlockDoesNotCreateSelfHealingTask|TestBrokerHandlePostTaskResumeUnblocksAfterCapabilityRepair|TestPostEscalation|TestRecoverTimedOutHeadlessTurnBlocksLocalWorktreeAfterRetryExhausted|TestRecoverFailedHeadlessTurnBlocksLocalWorktreeAfterRetryExhausted' -count=1`
- `go test ./internal/team -count=1`
- `go test ./internal/teammcp -count=1`
- Applied the exact staged follow-up diff to a clean temporary worktree and ran `go test ./internal/team ./internal/teammcp -count=1`
- `git diff --cached --check`

### DIFF
--- a/internal/agent/types.go
+++ b/internal/agent/types.go
@@ -101,8 +101,9 @@ type StreamFn func(msgs []Message, tools []AgentTool) <-chan StreamChunk
 type EscalationReason string
 
 const (
-	EscalationStuck      EscalationReason = "stuck"       // no phase change for too many ticks
-	EscalationMaxRetries EscalationReason = "max_retries" // PhaseError seen too many times for one task
+	EscalationStuck         EscalationReason = "stuck"          // no phase change for too many ticks
+	EscalationMaxRetries    EscalationReason = "max_retries"    // PhaseError seen too many times for one task
+	EscalationCapabilityGap EscalationReason = "capability_gap" // agent identified missing capability needed to proceed
 )
 
 // Escalator is a callback invoked when an agent can't make forward progress.
--- a/internal/team/broker.go
+++ b/internal/team/broker.go
@@ -4072,7 +4072,7 @@ func (b *Broker) canAccessChannelLocked(slug, channel string) bool {
 		}
 		return slug == b.oneOnOneAgent
 	}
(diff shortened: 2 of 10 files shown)


FranDias (approved):

PR #297, opened 2026-04-25
test(team): address Track A staff-review nits — sleep-free + accurate doc

FranDias:
## Summary

Three follow-ups from the Track A staff review on #289. None blocked the main refactor; landing them before they accumulate.

| Item | Where | Fix |
|---|---|---|
| Misattributed PR ref | \`broker_save_race_test.go:3\` | "PR #281's flaky test" → "flaky test surfaced on #281's CI; fix landed in #282" |
| \`time.Sleep(250ms)\` in test | \`broker_state_path_test.go\` | Replaced \`TestBrokerStop_NoWriteAfterReturn\` with \`TestBrokerStop_ClosesStopChannelAndPreservesState\` — sleep-free, asserts \`b.stopCh\` closed + state file bytes unchanged |
| Aspirational docstring | file header | Trimmed to match the test's actual assertion shape |

## Why the Stop test was rewritten, not just trimmed

The previous test claimed "no goroutine writes after Stop" but never started the broker — there were no goroutines to drain. The 250ms sleep also violated this repo's \`feedback_no_sleeps_in_tests\` rule and would pass for the wrong reason if the offending goroutine happened to be quiescent during the window.

The new test:
- Starts the broker via \`StartOnPort(0)\` so the HTTP listener goroutine is actually present.
- Asserts \`select { case <-b.stopCh: default: t.Fatal(...) }\` — deterministic, the read happens-after Stop's synchronous close.
- Compares \`os.ReadFile\` bytes (not modtime) immediately before and after Stop. Catches content drift AND phantom mid-Stop saves.

Honest about what it verifies: Stop's observable contract, not "no late writes ever." A stronger guarantee would need a \`sync.WaitGroup\` on the broker's goroutine set; that's broker instrumentation worth doing separately if/when the goroutine surface grows beyond the (test-disabled) activity watchdog.

## Stack

Based on \`refactor/broker-statepath-ctor\` (PR #289). When #289 merges, this auto-retargets to main.

## Test plan

- [x] All 9 affected tests pass: \`TestBrokerStop_ClosesStopChannelAndPreservesState\`, \`TestNewBrokerAt_PanicsOnEmptyPath\`, \`TestNewBrokerAt_PathSnapshottedAtConstruction\`, \`TestNewBroker_SkipStateLoadGateRespected\`, \`TestBrokerStateSnapshotPathIsLastGoodSibling\`, three \`TestDefaultBrokerStatePath_*\` cases, and \`TestSaveLocked_ConcurrentBrokersSamePathDoNotRace\` — total 0.79s.
- [x] Pre-push hook green, no \`--no-verify\`.


internal/team/broker_state_path_test.go
```diff
@@ -230,30 +236,34 @@ func TestBrokerStop_NoWriteAfterReturn(t *testing.T) {
...
-	// the watchdog ticker, if enabled, fires once per minute; shorter
-	// loops like wiki-index reconcile use 100ms+ intervals. 250ms is
-	// long enough to catch short-interval leaks without slowing the
-	// test suite meaningfully.
-	time.Sleep(250 * time.Millisecond)
+	// Deterministic check: stopCh must be closed when Stop returns.
+	// Any background goroutine that selects on it has observed the
+	// signal — the read here cannot happen before Stop closes the
+	// channel, since Stop is synchronous.
+	select {
```
FranDias, line 252: **NIT — process-wide-gate mutation, cross-reference to line 136.**

Unrelated to this hunk but in this same file: `TestNewBroker_SkipStateLoadGateRespected` (lines 135-137, outside this PR's diff) toggles `skipBrokerStateLoadOnConstruct = false` mid-body and restores via `t.Cleanup`. That's safe today because nothing in `internal/team` calls `t.Parallel()`. But this gate is process-wide. If a future contributor decides to parallel-mark sibling tests for speed, they'll observe the wrong value at NewBrokerAt time and silently load (or skip-load) contrary to expectation.

Not blocking — the package-wide convention is sequential — but a one-line comment on the existing test noting "do not parallel-mark this test or its siblings while this gate is mutated mid-body" would save someone half an hour later. Could fold into a follow-up cleanup.

FranDias (commented):
Three review-driven nits, all addressed correctly. The new `TestBrokerStop_ClosesStopChannelAndPreservesState` actually starts the broker via `StartOnPort(0)`, asserts `stopCh` is closed via select+default (deterministic, no sleep), and asserts file bytes are unchanged. Test name now matches the assertions. One latent foot-gun on the gate-mutation pattern noted inline.

PR #298, opened 2026-04-25
refactor(team): invert Pam → WikiWorker dep to a narrow interface (Track B start)

FranDias:
## Summary

Track B's first move: introduce a narrow \`pamWiki\` interface inside pam.go and switch \`PamDispatcher\` from \`*WikiWorker\` to that interface. Sets up the dependency-inversion seam needed to extract pam into its own package without an import cycle on \`team\`.

## What changed (14 lines)

\`\`\`go
type pamWiki interface {
    Enqueue(ctx context.Context, slug, path, content, mode, commitMsg string) (string, int, error)
    Repo() *Repo
}

type PamDispatcher struct {
    worker pamWiki  // was: *WikiWorker
    ...
}

func NewPamDispatcher(worker pamWiki, ...) *PamDispatcher { ... }
\`\`\`

\`*WikiWorker\` satisfies the interface via duck typing — every existing call site passes \`*WikiWorker\` unchanged.

## Why this matters

The original team-split plan ([planning notes](#)) assumed each leaf cluster (pam, operation, launcher+headless) would need a neutral \`teamcore\` package to break a circular type dependency on \`WikiWorker\`/\`Repo\`/\`ArchivistAuthor\`. The Phase-1 exploration showed pam actually only touches WikiWorker through 2 methods — narrow enough to invert in-place. This PR is the proof: 14 lines, no caller changes, all tests pass.

The next two leaves (operation, launcher+headless) have richer broker coupling and will need different shapes — but pam at least gives us the pattern to validate before tackling them.

## Stack

Stacked on \`chore/track-a-review-followups\` (#297) → \`refactor/broker-statepath-ctor\` (#289). When those land, this auto-retargets to main.

## Test plan

- [x] \`go test -count=1 ./internal/team/\` green (133s, includes 14 \`TestPam*\` tests)
- [x] Pre-push hook green (smoke + build + vhs in 12s), no \`--no-verify\`

## Out of scope

- Actual file move to \`internal/pam/\` — separate PR once the interface is settled.
- Eliminating the \`ArchivistAuthor\` constant + \`readArticle\` helper coupling pam still has on team — both are used elsewhere in team and can stay until the file move.


internal/team/pam.go
```diff
@@ -145,11 +145,25 @@ type PamDispatcherConfig struct {
...
+// commit a replacement article body via Enqueue, and read the existing
+// article body via Repo (which reads through to the on-disk markdown).
+//
+// Stating it as an interface here is the dependency-inversion seam for
+// the planned `internal/pam` extraction (Track B): once pam.go lives
+// in its own package, this interface stays with pam (the consumer) and
+// `*team.WikiWorker` continues to satisfy it via duck typing — no
+// import-cycle, no shared "core" package needed.
+type pamWiki interface {
+	Enqueue(ctx context.Context, slug, path, content, mode, commitMsg string) (string, int, error)
```
FranDias, line 153: **MEDIUM — docstring overstates the seam.**

The comment says once pam moves to `internal/pam`, this interface will let `*team.WikiWorker` satisfy it via duck typing with "no import-cycle, no shared 'core' package needed."

That's not quite true as written: `Repo() *Repo` returns `*team.Repo` (defined in `internal/team/wiki_git.go:79`). For pam to declare this interface in its own package, it must still import `internal/team` to name `*team.Repo` in the return type. Duck typing only saves you from importing the *implementation* — it can't elide the *type* the method returns.

The real fix when pam moves out is one of:

1. Hoist `Repo` into a neutral package (e.g. `internal/wikirepo`) that both `team` and `pam` import.
2. Replace `Repo() *Repo` in this interface with a narrower interface (e.g. `Repo() interface { Read(...) ... }`) that captures only the methods pam actually calls.

Not blocking this PR — the interface is still useful as-is — but the comment is currently a promise the next PR can't keep without extra work. Suggest tightening it to acknowledge the `*Repo` return is still a `team`-side type until follow-up.

FranDias (commented):
Interface seam looks right and the signatures match `*WikiWorker` byte-identically. One concern with the docstring claim — see inline.

PR #299, opened 2026-04-25
refactor(team): drop unused *Broker receiver from operation handler (Track B leaf 2)

FranDias:
## Summary

Track B's second leaf: the operation cluster's only \`(b *Broker)\` method body never references \`b\`. Convert it to a free function and update the two mux registrations.

## What changed (3 files, +11 / −5)

- \`internal/team/operation_bootstrap.go:346\` — \`func (b *Broker) handleOperationBootstrapPackage\` → \`func handleOperationBootstrapPackage\`. Body unchanged; the receiver was cosmetic. Added a header comment explaining the inversion seam.
- \`internal/team/broker.go:1553-1554\` — \`b.handleOperationBootstrapPackage\` → \`handleOperationBootstrapPackage\` in the two mux \`HandleFunc\` calls (\`/studio/bootstrap-package\` and \`/operations/bootstrap-package\`).
- \`internal/team/operation_matrix_test.go:219-222\` — test was constructing a broker only to invoke the handler off it; broker construct removed, handler called directly.

## Why this is the smallest possible inversion

Phase-1 exploration flagged operation as "deferred — broker handler methods need to factor out." The actual surface turned out to be even narrower than pam: not "factor out" but "remove cosmetic receiver." The operation cluster has ZERO real coupling to \`*Broker\` after this PR.

Subsequent move into \`internal/operation/\` is now a pure \`git mv\` + package rename — no interface, no constructor, no callers to retype. Lands as a separate PR once the inversion seams across all leaves are settled.

## Stack

Stacked on \`refactor/extract-pam\` (#298) → \`chore/track-a-review-followups\` (#297) → \`refactor/broker-statepath-ctor\` (#289). Each retargets to main as predecessors merge.

## Test plan

- [x] \`go test -count=1 ./internal/team/\` green (223s, 16 \`TestOperation*\` + \`TestBuildOperation*\` cases)
- [x] vhs visual regression PASS (after reverting a stale local \`testdata/vhs/help.txt\` from a previous failed run — golden was 153 lines, current vhs output is 48; reverted to origin/main's version)
- [x] Pre-push hook green, no \`--no-verify\`


internal/team/operation_routes_test.go
```diff
@@ -0,0 +1,95 @@
...
+				t.Fatalf("NewRequest %s: %v", path, err)
+			}
+			// No Authorization header — must be rejected.
+			resp, err := http.DefaultClient.Do(req)
+			if err != nil {
+				t.Fatalf("GET %s: %v", path, err)
+			}
+			defer resp.Body.Close()
+			if resp.StatusCode != http.StatusUnauthorized {
+				t.Fatalf("GET %s without token: expected 401, got %d", path, resp.StatusCode)
```
FranDias, line 91: **NIT — consider also covering wrong-token rejection.**

This test sends no `Authorization` header at all. It would catch a future regression that drops the wrapper entirely, but it would not catch a regression that swaps `requireAuth` for a permissive variant that accepts any non-empty token. A second case sending `Authorization: Bearer wrong-token` and asserting 401 closes that gap cheaply.

Not blocking — the current 0-token vs. valid-token pair is good enough to catch the realistic regression class (someone reorganizes the registration block and drops the wrapper). Defer-able.

internal/team/operation_routes_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"testing"
+)
+
+func TestOperationBootstrapPackageRoutesReachHandler(t *testing.T) {
+	statePath := filepath.Join(t.TempDir(), "broker-state.json")
+	b := NewBrokerAt(statePath)
+	if err := b.StartOnPort(0); err != nil {
+		t.Fatalf("StartOnPort: %v", err)
+	}
+	defer b.Stop()
```
FranDias, line 28: **NIT — defer ordering.**

`defer b.Stop()` runs after `t.Run` subtests complete, which is correct. But each call to `b.Stop()` also calls `b.server.Close()` — fine here since each subtest re-uses the same broker via the closure, not a fresh one per subtest. Just flagging that if someone later refactors this to construct a broker per subtest, they need to move `defer b.Stop()` inside the loop body. Current shape is correct.

FranDias (commented):
Receiver drop is safe — verified the body of `handleOperationBootstrapPackage` references no `b.*`. `requireAuth` accepts `http.HandlerFunc`, so the registration `b.requireAuth(handleOperationBootstrapPackage)` binds correctly. New route tests are useful (cover the mux wiring + the requireAuth wrapper, which the direct-call test in `operation_matrix_test.go` does not). Two small comments inline.

PR #312, opened 2026-04-25
[codex] fix slow DM and wiki loading

najmuzzaman-mohammad:
## Summary

- Open direct-message views optimistically instead of waiting for `/channels/dm` persistence to return.
- Replace per-article wiki catalog `git log` calls with a single batched commit index.
- Reuse the batched commit index for wiki section timestamps and backlink author metadata.

## Root Cause

Opening a DM was gated on `POST /channels/dm`, whose handler can persist the full broker state before responding. Wiki loads were doing repeated per-file `git log` subprocess calls while building catalog, section, and backlink metadata, so latency scaled with article count and backlink count.

## Validation

- `go test ./internal/team ./internal/channel -count=1`
- `npm run typecheck` from the temporary worktree with `web/node_modules` symlinked to the existing checkout

Synthetic timing during investigation: 400 per-file `git log` calls took about 1.7s, while one batched all-files `git log --name-only` call took about 9ms.

### DIFF
--- a/internal/team/wiki_article.go
+++ b/internal/team/wiki_article.go
@@ -41,6 +41,7 @@ import (
 	"regexp"
 	"sort"
 	"strings"
+	"time"
 )
 
 // ArticleMeta is the rich view sent to the UI for an article.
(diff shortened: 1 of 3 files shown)


FranDias (approved):

PR #316, opened 2026-04-26
test(team): migrate remaining NewBroker() test sites to newTestBroker(t)

FranDias:
## Summary

Closes the structural-isolation gap left open by [#289](https://github.com/nex-crm/wuphf/pull/289). That PR migrated tests on the persistence path to `NewBrokerAt(path)` but explicitly deferred ~59 `NewBroker()` call sites in tests that were only protected behaviorally (`worktree_guard_test.go` pinning `WUPHF_RUNTIME_HOME` + `skipBrokerStateLoadOnConstruct=true`, all sharing one tempdir).

This PR migrates **all 90 remaining `NewBroker()` call sites** in `internal/team` tests across 17 files to `newTestBroker(t)`, which calls `NewBrokerAt(filepath.Join(t.TempDir(), "broker-state.json"))`. Each broker now gets its own bound `statePath` under a per-test tempdir — structural isolation, not behavioral.

## Files touched (17)

| Originally listed in #289 | Scope expansion |
| --- | --- |
| `broker_entity_test.go` | `broker_entity_graph_test.go` |
| `broker_nex_register_test.go` | `broker_providers_test.go` |
| `broker_notebook_test.go` | `broker_test.go` |
| `broker_review_test.go` | `broker_wiki_dlq_test.go` |
| `broker_web_test.go` | `config_endpoint_test.go` |
| `dispatch_test.go` | `launcher_test.go` |
| `headless_codex_test.go` | `openclaw_e2e_test.go` |
| `openclaw_test.go` | `pam_test.go` |
|  | `wiki_worker_test.go` |

The 9 expanded-scope files apply the same mechanical swap — they were also constructing brokers via the package-var path and getting the same shared-tempdir behavioral isolation. Treating them in the same PR keeps the migration atomic.

## What this buys

After this PR, no `internal/team` test relies on the `worktree_guard_test.go` shared-tempdir + `skipBrokerStateLoadOnConstruct` shim for state isolation. Every test broker has its own bound `statePath`. The structural invariant from #289 (`b.statePath` cannot be retargeted mid-flight, leaked goroutines write to per-broker paths) now holds for the entire test corpus.

The only remaining `NewBroker()` callers are non-test code (`broker.go` default constructor, `launcher.go`, `headless_codex.go`) and one comment in `openclaw_test.go`.

## Test plan

- [x] `go vet ./internal/team/...` clean
- [x] `go test ./internal/team/...` green locally (~96s)
- [x] `go test -race -count=1 ./internal/team/...` green locally (~124s)
- [ ] CI green

### DIFF
--- a/internal/team/broker_web_test.go
+++ b/internal/team/broker_web_test.go
@@ -25,7 +25,7 @@ func TestWebUIProxyHandlerForwardsOnboardingRoutes(t *testing.T) {
 	}))
 	defer upstream.Close()
 
-	b := NewBroker()
+	b := newTestBroker(t)
 	req := httptest.NewRequest(http.MethodGet, "/onboarding/state?step=providers", nil)
 	rec := httptest.NewRecorder()
 
@@ -60,7 +60,7 @@ func TestWorkspaceShredRouteResetsBrokerWithoutShutdown(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	b := NewBroker()
+	b := newTestBroker(t)
 	b.mu.Lock()
 	b.messages = []channelMessage{{
 		ID:        "stale-message",
--- a/internal/team/headless_codex_test.go
+++ b/internal/team/headless_codex_test.go
@@ -91,7 +91,7 @@ func TestBuildCodexOfficeConfigOverridesIncludesOfficeMCPEnv(t *testing.T) {
 
 	t.Setenv("WUPHF_NO_NEX", "1")
 
-	broker := NewBroker()
+	broker := newTestBroker(t)
 	if err := broker.SetSessionMode(SessionModeOneOnOne, "pm"); err != nil {
 		t.Fatalf("SetSessionMode: %v", err)
 	}
@@ -163,7 +163,7 @@ func TestRunHeadlessCodexTurnUsesHeadlessOfficeRuntime(t *testing.T) {
 	l := &Launcher{
 		pack:        agent.GetPack("founding-team"),
 		cwd:         t.TempDir(),
-		broker:      NewBroker(),
+		broker:      newTestBroker(t),
 		headlessCtx: context.Background(),
 	}
 
@@ -297,7 +297,7 @@ func TestRunHeadlessCodexTurnUsesAssignedWorktreeForCodingAgents(t *testing.T) {
 	t.Setenv("CODEX_TUI_RECORD_SESSION", "1")
 	t.Setenv("CODEX_TUI_SESSION_LOG_PATH", "/tmp/controller-session.jsonl")
 
-	broker := NewBroker()
+	broker := newTestBroker(t)
 	ensureTestMemberAccess(broker, "general", "builder", "Builder")
 	ensureTestMemberAccess(broker, "general", "operator", "Operator")
 	task, _, err := broker.EnsurePlannedTask(plannedTaskInput{
@@ -416,7 +416,7 @@ func TestRunHeadlessCodexTurnUsesAssignedWorktreeForLocalWorktreeBuilder(t *test
 	t.Setenv("OPENAI_API_KEY", "test-key")
 	t.Setenv("PWD", repoRoot)
 
-	broker := NewBroker()
+	broker := newTestBroker(t)
 	ensureTestMemberAccess(broker, "general", "builder", "Builder")
 	task, _, err := broker.EnsurePlannedTask(plannedTaskInput{
 		Channel:       "general",
@@ -497,7 +497,7 @@ func TestRunHeadlessCodexTurnPassesScopedChannelEnv(t *testing.T) {
 	l := &Launcher{
 		pack:        agent.GetPack("founding-team"),
 		cwd:         t.TempDir(),
-		broker:      NewBroker(),
+		broker:      newTestBroker(t),
 		headlessCtx: context.Background(),
 	}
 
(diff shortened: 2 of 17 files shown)


internal/team/openclaw_test.go
```diff
@@ -185,7 +185,7 @@ func TestHandleClientEventForwardsAssistantMessage(t *testing.T) {
 func TestHandleClientEventSkipsUserRole(t *testing.T) {
 	t.Setenv("HOME", t.TempDir())
 	fake := newFakeOC()
-	broker := NewBroker()
+	broker := newTestBroker(t)
```
FranDias, line 189: Nit / stale comment: the rationale block at lines 178-184 (just above this hunk) explains the before/after-delta approach as needed `because NewBroker() rehydrates from persisted state that may contain messages from prior runs or tests`. After this migration the broker is constructed via `newTestBroker(t)` against a per-test `t.TempDir()` AND `skipBrokerStateLoadOnConstruct` is true, so neither rehydration nor cross-test bleed is possible here. The delta check is harmless but the comment is now misleading — worth either dropping the comment or simplifying the test to assert an absolute count. Same opportunity in `TestHandleClientEventForwardsAssistantMessage` just above (line 145).

internal/team/headless_codex_test.go
```diff
@@ -91,7 +91,7 @@ func TestBuildCodexOfficeConfigOverridesIncludesOfficeMCPEnv(t *testing.T) {
 
 	t.Setenv("WUPHF_NO_NEX", "1")
 
-	broker := NewBroker()
+	broker := newTestBroker(t)
```
FranDias, line 94: Verified the `broker.SetSessionMode(SessionModeOneOnOne, "pm")` call here is fine post-migration: SetSessionMode acquires `b.mu`, mutates `sessionMode`/`oneOnOneAgent`, and calls `saveLocked()`. The save now lands in `<t.TempDir>/broker-state.json` rather than the worktree_guard's leaked-tempdir default — strictly more isolated. No-op note, just double-checking the constructor swap before `SetSessionMode` doesn't change behavior.

internal/team/broker_web_test.go
```diff
@@ -60,7 +60,7 @@ func TestWorkspaceShredRouteResetsBrokerWithoutShutdown(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	b := NewBroker()
+	b := newTestBroker(t)
```
FranDias, line 63: Worth verifying explicitly: pre-PR, `NewBroker()` here resolved its state path under the test's `WUPHF_RUNTIME_HOME=home` (set on line 53), so a prior test's `b.Reset()`-driven `saveLocked` would have written to `<home>/.wuphf/team/broker-state.json`. Post-PR it writes to a separate `t.TempDir()`. The shred handler scrubs paths under `config.RuntimeHomeDir()` — i.e. `<home>/.wuphf/...` — so the broker state file is now outside the shred radius. The test's assertions (in-memory `messageCount == 0`, log file removal at `<home>/.wuphf/logs/channel-stderr.log`) are unaffected, but if a future test is added asserting `broker-state.json` is removed by shred, this isolation will silently make it pass for the wrong reason. Worth a one-line comment in the test noting the broker state is intentionally outside the runtime-home for this test.

internal/team/openclaw_e2e_test.go
```diff
@@ -37,7 +37,7 @@ func TestOpenclawBridgeFullPipeline_E2E(t *testing.T) {
 		t.Fatalf("identity: %v", err)
 	}
 
-	broker := NewBroker()
+	broker := newTestBroker(t)
```
FranDias, line 40: Out-of-scope follow-up to surface: `internal/teammcp/*_test.go` still has 33+ `team.NewBroker()` callsites (server_test.go, actions_test.go, skills_test.go) that hit the same shared-tempdir hazard #289 set out to fix — `team.DisableRealTaskWorktreeForTests()` only sets `WUPHF_RUNTIME_HOME` if it's empty, so every teammcp test ends up sharing one `<leaked-tempdir>/.wuphf/team/broker-state.json`. Not for this PR (scope is `internal/team/` per the description), but worth a tracking note so it doesn't get forgotten — could be a 30-line follow-up that exports `team.NewTestBroker(t)` from `test_support.go` and runs the same swap there.

FranDias (commented):
Mechanical migration looks correct across all 90 callsites I checked. The struct-literal and embedded-arg variants (`NewOpenclawBridge(newTestBroker(t), ...)`, `&Launcher{broker: newTestBroker(t)}`) all swap cleanly; no leftover `setBrokerStatePathForTest` / `withIsolatedBrokerState` / `brokerStatePath` monkey-patches remain in the team package; and the production `NewBroker()` callers (`broker.go:908`, `launcher.go:316/4725`, `headless_codex.go:132`) are correctly left untouched per scope.

A few non-blocking notes inline. The worktree_guard's `WUPHF_RUNTIME_HOME` pin and `skipBrokerStateLoadOnConstruct=true` flag are both still load-bearing — the env pin is consumed by `config.RuntimeHomeDir()` for manifest/log/launcher paths that this PR doesn't touch, and the load-skip gate is what makes `reloadedBroker(t, b)`'s explicit `loadState()` opt-in meaningful (without it, `NewBrokerAt(b.statePath)` would auto-load and the helper would be a no-op).

PR #319, opened 2026-04-26
test(teammcp): migrate NewBroker() test sites + characterization

FranDias:
## Summary

Closes the structural-isolation gap left open by [#316](https://github.com/nex-crm/wuphf/pull/316). That PR migrated the team package's own tests; this PR migrates the **33 `team.NewBroker()` call sites in `internal/teammcp` tests** that #316 explicitly flagged as out-of-scope follow-up, and lands a **characterization test floor first** (analogous to PR [#288](https://github.com/nex-crm/wuphf/pull/288)'s role for the team package's Track A refactor).

### 1. `8d09f4d2` — characterize broker-construction isolation invariants

Adds `internal/teammcp/broker_construction_test.go` pinning the two invariants every other teammcp test silently relies on:

- **`TestBrokerConstructionFreshHasZeroState`** — a freshly-constructed broker starts with zero actions, zero skill/action messages, zero requests, zero skills via `/skills`. No leakage from prior tests.
- **`TestBrokerConstructionTwoBrokersDoNotShareState`** — two brokers in the same test process do not see each other's state. b1 seeds `iso-marker-skill`; b2 must not observe it.

### 2. `d7ce95e0` — migrate `team.NewBroker()` → `newTestBroker(t)` across 33 sites

| File | Sites |
| --- | --- |
| `actions_test.go` | 4 |
| `skills_test.go` | 5 |
| `server_test.go` | 24 |
| **Total** | **33** |

Helper added to `testmain_test.go` mirrors `internal/team`'s unexported helper:
```go
func newTestBroker(t *testing.T) *team.Broker {
    t.Helper()
    return team.NewBrokerAt(filepath.Join(t.TempDir(), "broker-state.json"))
}
```

`b.statePath` now bound at construction — structural isolation, no reliance on `$HOME` resolution. The existing `t.Setenv("HOME", ...)` shims **stay** because they're load-bearing for non-broker paths (`RuntimeHomeDir`, agent log dirs, manifest paths).

The 3 `team.NewBroker()` references in `broker_construction_test.go` are intentional — the characterization needs the OLD pattern to remain meaningful as a regression detector.

### 3. `598f5757` — table-drive characterization over both broker constructors

Addresses staff-review nit: original characterization only exercised the legacy `team.NewBroker()` + HOME-shim pattern. Now both characterization tests table-drive over a `brokerCtors()` slice with `legacy` (pre-#316 pattern) and `helper` (`newTestBroker(t)`) strategies. Both must pass — directly asserts `newTestBroker(t)` preserves the contract instead of inferring it from migrated tests.

Also drive-by hardens `fetchSkillNames`: propagates `http.NewRequestWithContext` error and `url.QueryEscape`s the channel name.

### 4. `20108de3` — add skill-invocation lifecycle end-to-end test

New cross-cutting test exercising the broker's skill flow across HTTP, in-memory accessors, action log, and disk persistence — the same surface `team_skill_run` and the MCP skill tools flow through.

### 5. `e27f4443` — move skill-invocation test into `internal/team/`

Addresses staff-review IMPORTANT finding on commit 4: the original location in `tests/e2e/` worked only because that package doesn't trigger `internal/team/worktree_guard_test.go:init()` flipping `skipBrokerStateLoadOnConstruct = true`. The team package already has the documented helper for this seam (`reloadedBroker(t, b)` at `internal/team/broker_test.go:104`). The test now lives next to its sibling `TestBrokerStatePersistsAcrossReload_ChannelAndMember` and uses the same opt-in-to-disk-load pattern.

Drive-by review nits also applied: `http.NewRequestWithContext` for parity, explicit `json.Marshal` error check, pin `Source == "office"` on action-log entries, filter the failure diagnostic to relevant entries.

## Verification

| Check | After commit 5 |
| --- | --- |
| `go build ./internal/teammcp/...` | ✅ |
| `go vet ./internal/teammcp/...` | ✅ |
| `go test ./internal/teammcp/...` | ✅ |
| `go test ./internal/team/...` | ✅ |
| `go test -race -count=1 ./internal/team/...` | ✅ (~119s) |
| `go test -race -count=1 ./internal/teammcp/...` | ✅ |
| `go test -race -count=1 ./tests/e2e/...` | ✅ |

Both characterization tests pass under both constructors. The skill-invocation integration test exercises HTTP → save-to-disk → restart → rehydrate via the documented `reloadedBroker(t, b)` opt-in.

### DIFF
--- a/internal/team/broker_skill_invocation_integration_test.go
+++ b/internal/team/broker_skill_invocation_integration_test.go
@@ -1,41 +1,31 @@
-package e2e
+package team
+
+// End-to-end broker skill-invocation integration test. Exercises the
+// full lifecycle a team_skill_run / MCP skill tool flows through:
+// HTTP /skills/{name}/invoke, in-memory accessors (AllMessages,
+// Actions), action-log linkage via RelatedID, and disk persistence
+// across a broker restart.
+//
+// Sibling to TestBrokerStatePersistsAcrossReload_ChannelAndMember:
+// covers a third saveLocked call site (handleInvokeSkill), with the
+// added wrinkle of asserting per-invoker From accounting and the
+// usage_count rehydration after restart. Uses reloadedBroker(t, b)
+// for the restart leg — same opt-in-to-disk-load pattern the team
+// package's other persistence tests rely on.
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"net/http"
-	"path/filepath"
 	"strings"
 	"testing"
 
 	"github.com/nex-crm/wuphf/internal/agent"
-	"github.com/nex-crm/wuphf/internal/team"
 )
 
-// TestSkillInvocationE2E exercises the broker's skill lifecycle
-// end-to-end across the HTTP surface, the in-memory accessors, and
-// disk persistence — the same surface every team_skill_run / MCP
-// invocation flows through.
-//
-// The flow:
-//  1. Construct a broker with a bound state path (post-#316 helper
-//     pattern), seed it with a skill via SeedDefaultSkills.
-//  2. Invoke the skill twice via POST /skills/{name}/invoke,
-//     attributed to two different agent slugs.
-//  3. Assert the broker's observable state: usage_count, recorded
-//     skill_invocation messages (correct From, Channel, Kind),
-//     and the action log entries.
-//  4. Stop the broker and reconstruct a fresh one against the same
-//     state path — the persisted skill, usage count, messages, and
-//     actions must all rehydrate.
-//
-// This pins the contract every other test in the repo only exercises
-// in pieces, and asserts that the structural-isolation work (#289 +
-// #316 + this branch) didn't break the persistence path.
-func TestSkillInvocationE2E(t *testing.T) {
-	statePath := filepath.Join(t.TempDir(), "broker-state.json")
-
-	b := team.NewBrokerAt(statePath)
+func TestBrokerSkillInvocationE2E_PersistsAcrossReload(t *testing.T) {
+	b := newTestBroker(t)
 	if err := b.StartOnPort(0); err != nil {
 		t.Fatalf("start broker: %v", err)
 	}
@@ -57,11 +47,14 @@ func TestSkillInvocationE2E(t *testing.T) {
 
 	invoke := func(t *testing.T, invokedBy string) (skillID string, usageCount int) {
 		t.Helper()
-		body, _ := json.Marshal(map[string]string{
+		body, err := json.Marshal(map[string]string{
 			"invoked_by": invokedBy,
 			"channel":    channel,
 		})
-		req, err := http.NewRequest(http.MethodPost,
+		if err != nil {
+			t.Fatalf("marshal invoke body: %v", err)
+		}
+		req, err := http.NewRequestWithContext(context.Background(), http.MethodPost,
 			"http://"+b.Addr()+"/skills/"+skillName+"/invoke", bytes.NewReader(body))
 		if err != nil {
 			t.Fatalf("build invoke request: %v", err)
@@ -123,28 +116,27 @@ func TestSkillInvocationE2E(t *testing.T) {
 	}
 
 	// Action log should also show two skill_invocation entries linked
-	// to the seeded skill ID via RelatedID.
-	var actionsForSkill int
+	// to the seeded skill ID via RelatedID, sourced from "office".
+	var actionsForSkill []officeActionLog
 	for _, a := range b.Actions() {
 		if a.Kind == "skill_invocation" && a.RelatedID == skillID {
-			actionsForSkill++
+			actionsForSkill = append(actionsForSkill, a)
 		}
 	}
-	if actionsForSkill != 2 {
-		t.Errorf("expected 2 action-log entries for skill %s, got %d (all=%+v)",
-			skillID, actionsForSkill, b.Actions())
+	if len(actionsForSkill) != 2 {
+		t.Errorf("expected 2 action-log entries for skill %s, got %d (%+v)",
+			skillID, len(actionsForSkill), actionsForSkill)
+	}
+	for _, a := range actionsForSkill {
+		if a.Source != "office" {
+			t.Errorf("skill_invocation action source=%q, want %q", a.Source, "office")
+		}
 	}
 
-	// Stop the broker before reconstructing — closes the listener and
-	// flushes any in-flight saves. Persistence must survive the
-	// restart with the same bound state path.
-	b.Stop()
-
-	b2 := team.NewBrokerAt(statePath)
-	if err := b2.StartOnPort(0); err != nil {
-		t.Fatalf("restart broker: %v", err)
-	}
-	defer b2.Stop()
+	// Restart the broker against the same state path via the documented
+	// helper — opts back in to loadState(), which test-mode NewBrokerAt
+	// otherwise skips to prevent cross-test state leakage.
+	b2 := reloadedBroker(t, b)
 
 	var rehydrated int
 	for _, m := range b2.AllMessages() {
@@ -157,7 +149,12 @@ func TestSkillInvocationE2E(t *testing.T) {
 	}
 
 	// Confirm the skill itself survived and the usage count persisted.
-	req, err := http.NewRequest(http.MethodGet,
+	if err := b2.StartOnPort(0); err != nil {
+		t.Fatalf("restart broker: %v", err)
+	}
+	defer b2.Stop()
+
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet,
 		"http://"+b2.Addr()+"/skills?channel="+channel, nil)
 	if err != nil {
 		t.Fatalf("build skills request: %v", err)
--- a/internal/teammcp/broker_construction_test.go
+++ b/internal/teammcp/broker_construction_test.go
@@ -1,0 +1,130 @@
+package teammcp
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"testing"
+
+	"github.com/nex-crm/wuphf/internal/agent"
+	"github.com/nex-crm/wuphf/internal/team"
+)
+
+// These tests pin the broker-construction invariants that every other
+// teammcp test implicitly depends on:
+//
+//  1. A freshly-constructed broker, in a teammcp test scoped by
+//     t.Setenv("HOME", t.TempDir()), starts with no skills, no actions,
+//     and no skill-invocation messages — i.e. nothing leaks in from any
+//     prior test in the package binary.
+//
+//  2. Two brokers constructed in the same test process with disjoint
+//     HOME tempdirs do not see each other's mutations. State is per-
+//     broker, not shared across constructions.
+//
+// These hold under the current `team.NewBroker()` + `t.Setenv("HOME",...)`
+// pattern. Any future migration of this package's broker construction
+// (e.g. to NewBrokerAt or a newTestBroker helper) must keep them green.
+// If they regress, the package's other tests can silently corrupt each
+// other through state leakage and the symptoms will be flaky/wrong.
+
+func TestBrokerConstructionFreshHasZeroState(t *testing.T) {
+	t.Setenv("HOME", t.TempDir())
+	b := team.NewBroker()
+	if err := b.StartOnPort(0); err != nil {
+		t.Fatalf("start broker: %v", err)
+	}
+	defer b.Stop()
+
+	if got := len(b.Actions()); got != 0 {
+		t.Fatalf("fresh broker should have zero actions, got %d (%+v)", got, b.Actions())
+	}
+	for _, m := range b.AllMessages() {
+		switch m.Kind {
+		case "skill_invocation", "external_action_planned":
+			t.Fatalf("fresh broker should have no skill/action messages, found %+v", m)
+		}
+	}
+	if got := len(b.Requests("general", true)); got != 0 {
+		t.Fatalf("fresh broker should have zero requests in general, got %d", got)
+	}
+	if skills := fetchSkillNames(t, b, "general"); len(skills) != 0 {
+		t.Fatalf("fresh broker should expose zero skills in general, got %v", skills)
+	}
+}
+
+func TestBrokerConstructionTwoBrokersDoNotShareState(t *testing.T) {
+	// Broker 1: scoped to its own HOME, seed a uniquely-named skill.
+	t.Setenv("HOME", t.TempDir())
+	b1 := team.NewBroker()
+	if err := b1.StartOnPort(0); err != nil {
+		t.Fatalf("start b1: %v", err)
+	}
+	defer b1.Stop()
+	b1.SeedDefaultSkills([]agent.PackSkillSpec{{
+		Name:        "iso-marker-skill",
+		Title:       "Isolation Marker",
+		Description: "If b2 sees this name, the construction-isolation invariant has regressed.",
+		Trigger:     "n/a",
+		Tags:        []string{"characterization"},
+		Content:     "n/a",
+	}})
+
+	// Sanity: b1 itself sees its own seeded skill.
+	if got := fetchSkillNames(t, b1, "general"); !sliceContains(got, "iso-marker-skill") {
+		t.Fatalf("b1 should see its own seeded skill, got %v", got)
+	}
+
+	// Broker 2: distinct HOME, freshly constructed.
+	t.Setenv("HOME", t.TempDir())
+	b2 := team.NewBroker()
+	if err := b2.StartOnPort(0); err != nil {
+		t.Fatalf("start b2: %v", err)
+	}
+	defer b2.Stop()
+
+	// b2 must NOT see b1's marker — that would mean state leaked across
+	// constructions through a shared default state path.
+	if got := fetchSkillNames(t, b2, "general"); sliceContains(got, "iso-marker-skill") {
+		t.Fatalf("b2 leaked b1's marker skill (state-isolation regression); b2 skills=%v", got)
+	}
+	for _, m := range b2.AllMessages() {
+		if m.Kind == "skill_invocation" || m.Kind == "external_action_planned" {
+			t.Fatalf("b2 should be empty but observed leaked message %+v", m)
+		}
+	}
+}
+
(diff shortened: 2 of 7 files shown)


internal/teammcp/broker_construction_test.go
```diff
@@ -0,0 +1,130 @@
...
+//
+// These hold under the current `team.NewBroker()` + `t.Setenv("HOME",...)`
+// pattern. Any future migration of this package's broker construction
+// (e.g. to NewBrokerAt or a newTestBroker helper) must keep them green.
+// If they regress, the package's other tests can silently corrupt each
+// other through state leakage and the symptoms will be flaky/wrong.
+
+func TestBrokerConstructionFreshHasZeroState(t *testing.T) {
+	t.Setenv("HOME", t.TempDir())
+	b := team.NewBroker()
```
FranDias: **SUGGESTION** — Asymmetric characterization.

Both tests construct via the OLD `team.NewBroker()` pattern, so they pin the pre-migration contract. They never exercise `newTestBroker(t)` itself. If a future change to `team.NewBrokerAt` (or our local helper) silently bound to a process-shared path, neither test here would catch it — it would only surface as flakiness in the 33 migrated tests.

Low priority because the helper is a one-line wrapper around a function that panics on empty input. But cheap to close: add a parallel pair (`...FreshHasZeroState_NewHelper`, `...TwoBrokersDoNotShareState_NewHelper`) using `newTestBroker(t)` — or pull the broker constructor into a closure and table-drive both tests over `{old, new}`. Either turns this into a true bidirectional invariant pin.

internal/teammcp/broker_construction_test.go
```diff
@@ -0,0 +1,130 @@
...
+	for _, m := range b2.AllMessages() {
+		if m.Kind == "skill_invocation" || m.Kind == "external_action_planned" {
+			t.Fatalf("b2 should be empty but observed leaked message %+v", m)
+		}
+	}
+}
+
+func fetchSkillNames(t *testing.T, b *team.Broker, channel string) []string {
+	t.Helper()
+	req, _ := http.NewRequestWithContext(context.Background(), http.MethodGet,
```
FranDias: **SUGGESTION** — Two minor hygiene nits in this helper:

1. The `http.NewRequestWithContext` error is dropped via `_`. If `b.Addr()` ever returned empty (e.g., a future regression in `StartOnPort(0)` that didn't error but also didn't bind), the next line `req.Header.Set(...)` would nil-panic instead of producing a helpful test failure. Cheap to handle: `req, err := ...; if err != nil { t.Fatalf("new request: %v", err) }`.
2. `channel` is interpolated into the URL without `url.QueryEscape`. Currently always called with the literal `"general"`, so safe — but the helper signature invites a future caller to pass an arbitrary string. Either escape it or document the assumption inline.

Neither is blocking; both are seconds to fix.

internal/teammcp/server_test.go
```diff
@@ -1117,7 +1117,7 @@ func TestHandleTeamRuntimeStateIncludesRecoveryAndCapabilities(t *testing.T) {
 func TestHandleTeamRequestDefaultsApprovalOptions(t *testing.T) {
 	t.Setenv("HOME", t.TempDir())
 
-	b := team.NewBroker()
+	b := newTestBroker(t)
```
FranDias, line 1120: **SUGGESTION** (representative of a small pattern across the file) — `t.Setenv("HOME", t.TempDir())` looks like it may be dead code in this test post-migration. `handleTeamRequest` → `brokerPostJSON` → broker `/requests` handler doesn't appear to traverse `RuntimeHomeDir()` / `os.UserHomeDir()`; the broker's only remaining `HOME`-resolving site (`defaultBrokerStatePath`) is no longer invoked because `newTestBroker(t)` binds the path at construction.

The commit message claims the shim is "load-bearing for non-broker paths." That's true for the memory tests (which call `team.QuerySharedMemory` / `team.WriteSharedMemory` → `os.UserHomeDir`), the runtime-state test, and the worktree-touching tests. It looks dead for the pure-broker-CRUD tests like this one, `TestHandleTeamPlanCreatesDependentBlockedTasks`, `TestHandleTeamPlanPreservesTaskMetadata`, `TestHandleTeamTaskCreatePreservesTaskMetadata`, `TestHandleTeamMemberCreateTriggersReconfigure`, etc.

Not worth holding the PR for — defense-in-depth shim is fine and removing it is a separate cleanup. Just flagging that the commit message's blanket justification doesn't strictly hold for every callsite.

FranDias (commented):
Reviewed both commits in order. Migration is mechanically sound: 33 callsites swapped, exactly 3 OLD-pattern `team.NewBroker()` calls remain (intentionally, in the characterization file), the dropped `team` imports check out (still required in `server_test.go` for `RuntimeSnapshot` / `CapabilityKey*` / `SessionModeOneOnOne`, dropped from `actions_test.go` and `skills_test.go` with no residual references), and the `newTestBroker(t)` helper exactly mirrors `internal/team`'s unexported helper. Verified the characterization tests pass on commit 1 (under the OLD pattern) and on commit 2 (after the migration), as the PR description claims.

No CRITICAL or IMPORTANT findings. A handful of SUGGESTION-tier items inline — none are blocking. The biggest-shaped one is structural: the characterization holds the OLD construction green but doesn't itself exercise the NEW construction under the same assertion surface, so a hypothetical bug in `NewBrokerAt`'s path binding would slip past these two tests (it would only surface as flakiness in the migrated suite). Worth considering a parallel test using `newTestBroker(t)` to close that loop, but not blocking — the migration's safety case is already strong.

**Recommendation: ready to mark-ready.**

tests/e2e/skill_invocation_test.go
```diff
@@ -0,0 +1,198 @@
...
+//     state path — the persisted skill, usage count, messages, and
+//     actions must all rehydrate.
+//
+// This pins the contract every other test in the repo only exercises
+// in pieces, and asserts that the structural-isolation work (#289 +
+// #316 + this branch) didn't break the persistence path.
+func TestSkillInvocationE2E(t *testing.T) {
+	statePath := filepath.Join(t.TempDir(), "broker-state.json")
+
+	b := team.NewBrokerAt(statePath)
```
FranDias: **IMPORTANT — fragility on the rehydrate leg.** This test depends on `team.NewBrokerAt(statePath)` auto-loading from disk during construction. In production that is true (`skipBrokerStateLoadOnConstruct == false`), but inside the team package's tests it is flipped to `true` by `internal/team/worktree_guard_test.go` `init()`. The reason your test works is that you live in `package e2e`, so that `_test.go` `init()` does not run — `loadState()` fires here, but that is incidental to where you placed the file.

The codebase's documented idiom for restart-from-disk testing is `reloadedBroker(t, b)` (see `internal/team/broker.go:894-901` and the comment at `internal/team/broker_restart_integration_test.go:96-99`: "the only way to opt back into disk read under the test-mode gate"). That helper is in-package, so unreachable from `tests/e2e/`.

Two options:
1. Move this test to `internal/team/skill_invocation_restart_test.go` and use `reloadedBroker(t, b)` — that's the closest sibling to `TestBrokerStatePersistsAcrossReload_ChannelAndMember` and the rehydrate assertion becomes load-bearing rather than coincidental.
2. Keep it here but add a one-liner comment naming the dependency, so the next person who flips that gate's default (or factors out the test_only init) doesn't silently regress this test into a no-op rehydrate.

Option 1 is what I'd do.

internal/team/broker_skill_invocation_integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/nex-crm/wuphf/internal/agent"
+	"github.com/nex-crm/wuphf/internal/team"
+)
```
FranDias, line 25: **SUGGESTION — placement.** `tests/e2e/delegation_test.go` exercises agent runtime + orchestration via fake LLM resolvers (`agent.StreamFnResolver`); this test exercises only the broker's HTTP surface + persistence. The closer sibling is `internal/team/broker_restart_integration_test.go::TestBrokerStatePersistsAcrossReload_ChannelAndMember` — same construction style, same Stop+reload shape. Either move this next to it as `broker_skill_invocation_restart_test.go` (and use the in-package `reloadedBroker` helper), or — if you want it to stay in `tests/e2e/` — also seed an agent and route the invoke through the runtime to make it actually "end to end" rather than a broker integration test.

Not blocking, but the misplacement compounds the gate-fragility comment on line 38.

tests/e2e/skill_invocation_test.go
```diff
@@ -0,0 +1,198 @@
...
+		Content:     "Step 1: Reproduce. Step 2: Isolate. Step 3: Root cause. Step 4: Fix.",
+	}})
+
+	invoke := func(t *testing.T, invokedBy string) (skillID string, usageCount int) {
+		t.Helper()
+		body, _ := json.Marshal(map[string]string{
+			"invoked_by": invokedBy,
+			"channel":    channel,
+		})
+		req, err := http.NewRequest(http.MethodPost,
```
FranDias: **SUGGESTION — context parity.** Your characterization test in this same PR (`internal/teammcp/broker_construction_test.go:134`, `fetchSkillNames`) uses `http.NewRequestWithContext(context.Background(), ...)`. The two `http.NewRequest(...)` calls in this file (here and line 160) bypass the context — drop in `http.NewRequestWithContext(context.Background(), ...)` for parity and to keep the codebase clean against `noctx` linters.

internal/team/broker_skill_invocation_integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+			Skill struct {
+				ID         string `json:"id"`
+				UsageCount int    `json:"usage_count"`
+			} `json:"skill"`
+		}
+		if err := json.NewDecoder(resp.Body).Decode(&out); err != nil {
+			t.Fatalf("decode invoke response: %v", err)
+		}
+		return out.Skill.ID, out.Skill.UsageCount
+	}
```
FranDias, line 82: **SUGGESTION — the `invoke` closure could simplify.** It already takes `t *testing.T` (good — preserves stack frames via `t.Helper()`), but the channel/skillName closure-capture is fine. Two minor things:

- `body, _ := json.Marshal(...)` swallows the marshal error. Marshalling `map[string]string` cannot fail in practice but the no-context-error pattern from `delegation_test.go` is to `t.Fatalf` even on the impossible cases for grep-ability. Compare to the explicit `if err != nil` on the request build below — same call site, two different error policies.
- `req.Header.Set("Authorization", "Bearer "+b.Token())` is duplicated between this closure and the GET on line 165. If you stay in `tests/e2e/`, a tiny `authedRequest(t, b, method, url, body)` would tighten this up; if you move to `internal/team/`, follow the `post := func(...)` shape that `broker_restart_integration_test.go` already uses.

tests/e2e/skill_invocation_test.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	if actionsForSkill != 2 {
+		t.Errorf("expected 2 action-log entries for skill %s, got %d (all=%+v)",
+			skillID, actionsForSkill, b.Actions())
+	}
+
+	// Stop the broker before reconstructing — closes the listener and
+	// flushes any in-flight saves. Persistence must survive the
+	// restart with the same bound state path.
+	b.Stop()
```
FranDias: **SUGGESTION — race risk on Stop is theoretical here.** You called this out in the review prompt. Worth confirming for the record: `handleInvokeSkill` at `internal/team/broker.go:10796` calls `b.saveLocked()` synchronously under `b.mu` and the HTTP response is sent after that returns, so the on-disk rename is complete before this `b.Stop()` call. `Stop()` does not wait on background goroutines, but none of the goroutines started by `StartOnPort(0)` write to broker state for this test's exercised paths (the watchdog only mutates if it observes a stuck "active" agent — there are no agents). Race-clean today, but if someone later adds a background skill-stat aggregator that calls `saveLocked()`, the rehydrate could read pre-flush state.

No action required, but consider adding a one-line comment near `b.Stop()` documenting that this test relies on save being synchronous in the invoke handler — otherwise the next person who refactors save to be async will have a flake mystery.

tests/e2e/skill_invocation_test.go
```diff
@@ -0,0 +1,198 @@
...
+	// Action log should also show two skill_invocation entries linked
+	// to the seeded skill ID via RelatedID.
+	var actionsForSkill int
+	for _, a := range b.Actions() {
+		if a.Kind == "skill_invocation" && a.RelatedID == skillID {
+			actionsForSkill++
+		}
+	}
+	if actionsForSkill != 2 {
+		t.Errorf("expected 2 action-log entries for skill %s, got %d (all=%+v)",
```
FranDias: **SUGGESTION — error message verbosity.** `t.Errorf(... b.Actions())` on a failure dumps the entire `officeActionLog` slice, which can be noisy if other action kinds get added in unrelated work (channel creation, member add, etc.). Filtering to just `skill_invocation` actions in the dump would keep the failure message focused: `actionsForSkill, all := filterByKind(b.Actions(), "skill_invocation"); ...; b.Errorf(..., actionsForSkill, all)`. Pure ergonomics.

internal/team/broker_skill_invocation_integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+	if invocations["qa"] != 1 {
+		t.Errorf("expected exactly 1 invocation by qa, got %d", invocations["qa"])
+	}
+
+	// Action log should also show two skill_invocation entries linked
+	// to the seeded skill ID via RelatedID.
+	var actionsForSkill int
+	for _, a := range b.Actions() {
+		if a.Kind == "skill_invocation" && a.RelatedID == skillID {
```
FranDias, line 122: **Note on the `Source` vs `RelatedID` bug you flagged in the prompt** — verified. The skill invoke handler at `internal/team/broker.go:10794` passes `sk.ID` as the `relatedID` arg (the 6th positional) to `appendActionLocked(kind, source, channel, actor, summary, relatedID)`, with `"office"` as `source`. So `RelatedID == sk.ID` and `Source == "office"` is correct. Spot-checked the other 9 `appendActionLocked` call sites in `broker.go` — all of them pass IDs into the `relatedID` slot, none confuse it with `source`. So that was a one-off thinko on the test side; no broader codebase bug.

If you want a stronger assertion, pin both: `if a.Kind == "skill_invocation" && a.RelatedID == skillID && a.Source == "office" { actionsForSkill++ }` — that catches a future refactor that retargets `source` to `"skill"` or similar.

internal/team/broker_skill_invocation_integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	if _, count2 := invoke(t, "qa"); count2 != 2 {
+		t.Fatalf("second invoke: expected usage_count=2, got %d", count2)
+	}
+
+	// Validate observable state: two skill_invocation messages with
+	// distinct From slugs, attributed to the requested channel.
+	invocations := map[string]int{}
+	for _, m := range b.AllMessages() {
```
FranDias, line 99: **Brittleness call-out (asked in the prompt).** This test pins five surface-area contracts simultaneously: `b.AllMessages()` returning unexported `channelMessage` (with `Kind`/`From`/`Channel`/`Title` fields), `b.Actions()` returning unexported `officeActionLog` (with `Kind`/`RelatedID`), the `/skills/{name}/invoke` JSON shape (`{skill: {id, usage_count}}`), and the `/skills?channel=` JSON shape. None of these are actively under refactor that I saw, but two are exposed via *unexported types* (`channelMessage`, `officeActionLog`) — Go lets you call methods + access exported fields from another package, but you cannot name the type, which means a future refactor to typed accessors (`b.SkillInvocations() []SkillInvocation`) would break this file silently in an awkward place. Flagging for awareness; not a blocker.

internal/team/broker_skill_invocation_integration_test.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	}
+	if !found {
+		names := make([]string, 0, len(skillsBody.Skills))
+		for _, s := range skillsBody.Skills {
+			names = append(names, s.Name)
+		}
+		t.Errorf("seeded skill %q not found after restart; skills=%s",
+			skillName, strings.Join(names, ","))
+	}
```
FranDias, line 194: **Time-stability (asked in the prompt).** No test assertion in this file reads `LastExecutionAt`, `UpdatedAt`, or any other RFC3339 timestamp — the `usage_count` and message/action counts are the only persisted state checked. Time-flake-clean.

**One unrelated nit on the PR itself**: the PR description (`gh pr view 319`) doesn't mention this new e2e file at all — title is still "migrate NewBroker() test sites + characterization" and the body lists "Two commits, in order" while the branch now has four commits. Worth either updating the description before merge or splitting this e2e into a follow-up PR. Per the prompt you noted this scope-crept beyond the original PR title; I lean toward updating the description and merging together rather than spinning a separate PR for ~200 lines.

FranDias (commented):
Re-review focused on `20108de3` (the new e2e test added on top of the migration).

**Summary**: the e2e adds genuine value over `TestHandleTeamSkillRunBumpsUsageAndLogsInvocation` in two specific places — (a) the per-invoker `From` accounting across two distinct slugs in the same channel, and (b) the stop-then-rehydrate-from-disk leg. Otherwise the message/usage assertions overlap heavily with the teammcp test. The placement is debatable: it lives in `tests/e2e/` next to `delegation_test.go` (which exercises agent runtime + orchestration with fake LLM resolvers), but exercises only the broker's HTTP + persistence surface — same shape as `internal/team/broker_restart_integration_test.go::TestBrokerStatePersistsAcrossReload_ChannelAndMember`. That existing test uses the in-package `reloadedBroker(t, b)` helper, which is the documented "only way to opt back into disk read under the test-mode gate". Your test bypasses that helper and relies on the fact that the team package's `worktree_guard_test.go` `init()` (which flips `skipBrokerStateLoadOnConstruct = true`) does not run when the test lives in `package e2e`. That works today but is coincidental — see the inline comment.

**Bottom line**: ready to merge after addressing the rehydrate-via-test-mode-gate fragility (one-line comment is fine), and updating the PR description to mention the new e2e file. The other inline notes are SUGGESTION-grade.

PR #320, opened 2026-04-26
fix(team): drain headless codex queue workers via stop channel

FranDias:
## Why

The `runHeadlessCodexQueue` goroutine spawned by `enqueueHeadlessCodexTurnRecord` had no per-test cleanup channel, so under `go test ./internal/team/` it outlived the spawning test and:

- **wrote to `t.TempDir()` files while RemoveAll cleanup ran**, surfacing as `directory not empty` failures on the **release-build job** (which runs `go test ./...` *without* `-race` AND *without* the `internal/team` carve-out the CI matrix uses);
- tripped DATA RACE on subsequent tests' setup of `headlessActive` / `headlessQueues` maps under `-race` (already gated by CI carve-out).

This was the audit's 🔴 #1 blocker — auto-release was failing intermittently because of this leak. Closes that path.

## Fix shape

- Add `headlessStopCh chan struct{}` and `headlessWorkerWg sync.WaitGroup` to `Launcher`. Lazily allocated under `headlessMu` so a zero-value `Launcher` (used by ~95 test sites) stays valid.
- New `spawnHeadlessWorker(slug)` helper does `wg.Add(1)` + `go`-spawn under-lock; replaces the 3 `go runHeadlessCodexQueue(slug)` sites in `enqueueHeadlessCodexTurnRecord`.
- `runHeadlessCodexQueue` now takes a `<-chan struct{}` stop param, `defer wg.Done()`, and selects on stop in its outer loop.
- New `stopHeadlessWorkers()` method closes the stop channel and waits for all workers to drain. Idempotent.
- `newHeadlessLauncherForTest` now takes `*testing.T` and registers `t.Cleanup(l.stopHeadlessWorkers)` so any test that spawns workers drains them before the test's `t.TempDir` cleanup fires (LIFO order on `t.Cleanup` ensures this — `t.TempDir()` registers cleanup first when called earlier in the test, helper registers cleanup second, helper fires first).

## Verification

- `go test ./internal/team/` (mirrors release-build job): **green in 167s**.
- `golangci-lint run ./internal/team/...`: **0 issues**.
- v0.79.0 successfully re-cut at 02:59 UTC after the orphan tag was deleted — auto-release now landing cleanly.

## Out of scope (deliberate)

`go test -race ./internal/team/` still has a **separate, pre-existing race** on package-level test override globals (`prepareTaskWorktree`, `cleanupTaskWorktree`) where defer-style restorers race the worker. That's the same architectural shape #304 fixed for `headlessCodexRunTurn` (atomic.Pointer); same fix should be applied to those globals in a follow-up. Gated by CI's `-race` carve-out for `internal/team`.

## Test plan

- [x] `go test ./internal/team/` clean.
- [x] `golangci-lint run` clean.
- [x] Local pre-push hooks (build + smoke + vhs) green.
- [ ] CI green.

### DIFF
--- a/internal/team/headless_codex.go
+++ b/internal/team/headless_codex.go
@@ -228,7 +228,7 @@ func (l *Launcher) enqueueHeadlessCodexTurnRecord(slug string, turn headlessCode
 				appendHeadlessCodexLog(slug, "queue-replace: refreshed pending turn for same task")
 			}
 			if startWorker {
-				go l.runHeadlessCodexQueue(slug)
+				l.spawnHeadlessWorker(slug)
 			}
 			return
 		}
@@ -277,7 +277,7 @@ func (l *Launcher) enqueueHeadlessCodexTurnRecord(slug string, turn headlessCode
 			l.headlessMu.Unlock()
 			appendHeadlessCodexLog(slug, "queue-replace: lead queue at cap, replacing pending turn with urgent task notification")
 			if startWorker {
-				go l.runHeadlessCodexQueue(slug)
+				l.spawnHeadlessWorker(slug)
 			}
 			return
 		}
@@ -309,27 +309,7 @@ func (l *Launcher) enqueueHeadlessCodexTurnRecord(slug string, turn headlessCode
 		cancel()
 	}
 	if startWorker {
-		// TODO(test-isolation): this goroutine has no per-test cleanup channel,
-		// so under `go test -race ./internal/team/` it outlives the spawning
-		// test and trips DATA RACE on subsequent tests' setup of headlessActive
-		// / headlessQueues maps. Reproducible:
-		//
-		//   bash scripts/test-go.sh ./internal/team   # passes (-race carved out)
-		//   go test -race ./internal/team             # flakes
-		//
-		// The race detector trace (representative — different tests on each run):
-		//
-		//   Write at ... by goroutine N (this gowrap, after test exit):
-		//     (*Launcher).beginHeadlessCodexTurn  ← l.headlessActive[slug] = …
-		//   Previous read at ... by goroutine N-1 (test cleanup, finished):
-		//     TestRecoverFailedHeadlessTurnRequeues...  ← delete(l.headlessActive, slug)
-		//
-		// Fix shape: thread a per-test stop channel through Launcher (or
-		// reuse l.broker.stopCh) and have runHeadlessCodexQueue select on
-		// it inside its outer for-loop, so test cleanup can drain in-flight
-		// workers before the next TestMain leg starts. Tracked alongside
-		// the carve-out in .github/workflows/ci.yml (go-test-list).
-		go l.runHeadlessCodexQueue(slug)
+		l.spawnHeadlessWorker(slug)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


internal/team/headless_codex.go
```diff
@@ -371,8 +351,57 @@ func (l *Launcher) headlessLeadTurnNeedsImmediateWakeLocked(slug, prompt string)
...
+		// stop while the worker is in steady state.
+		select {
+		case <-stop:
+			l.headlessMu.Lock()
+			delete(l.headlessWorkers, slug)
+			l.headlessMu.Unlock()
+			return
+		default:
+		}
 		func() {
```
FranDias, line 416: **IMPORTANT**: the stop check only runs at the top of the outer `for` loop, but a worker that's mid-turn is parked inside `headlessCodexRunTurn(l, turnCtx, ...)` until the provider call returns or `turnCtx` deadlines (line 1032: `context.WithTimeout(baseCtx, timeout)` where `baseCtx = l.headlessCtx`). `stopHeadlessWorkers()` closes `headlessStopCh` but does **not** cancel `l.headlessCancel`, so an in-flight turn will block `wg.Wait()` for up to the full `timeout` (minutes for `headlessCodexLocalWorktreeTurnTimeout`).

In tests this is hidden because `headlessCodexRunTurnOverride` returns immediately, but in any path where a real provider call is reached the t.Cleanup will hang past the test budget. Two options:

1. Have `stopHeadlessWorkers` also call `l.headlessCancel()` before `wg.Wait()` so any in-flight turn observes ctx cancellation. Cheap and surgical.
2. Build per-turn ctx as `WithCancel(stop-derived)` so the stop close propagates into the turn ctx automatically.

Option 1 is the smaller change and matches the existing `headlessCancel` plumbing. Worth landing in this PR rather than deferring — the comment in the PR description claims the stop is an outer-loop tick, but the real cost (parked turn) isn't acknowledged.

internal/team/headless_codex.go
```diff
@@ -371,8 +351,57 @@ func (l *Launcher) headlessLeadTurnNeedsImmediateWakeLocked(slug, prompt string)
...
+// sites must funnel through here so no worker escapes the WaitGroup.
+func (l *Launcher) spawnHeadlessWorker(slug string) {
+	l.headlessMu.Lock()
+	if l.headlessStopCh == nil {
+		l.headlessStopCh = make(chan struct{})
+	}
+	stop := l.headlessStopCh
+	l.headlessWorkerWg.Add(1)
+	l.headlessMu.Unlock()
+	go l.runHeadlessCodexQueue(slug, stop)
```
FranDias, line 367: **SUGGESTION**: lazy allocation of `headlessStopCh` is split between `spawnHeadlessWorker` and `stopHeadlessWorkers`, both under `headlessMu` — fine for correctness, but if `stopHeadlessWorkers` is called *before* any worker has been spawned, you create+close a fresh channel that no one ever reads. That's a no-op so it's harmless, but it also means subsequent `spawnHeadlessWorker` calls will reuse the already-closed `headlessStopCh` and immediately return on first iteration. That breaks any test sequence that does `stopHeadlessWorkers` → spawn-new-work (e.g. a hypothetical reuse-launcher test). Worth a one-line comment that the Launcher is not reusable after stop, or reset `headlessStopCh = nil` after `wg.Wait()` returns.

internal/team/headless_codex_test.go
```diff
@@ -861,14 +861,32 @@ func canonicalPath(path string) string {
...
+	// "directory not empty" failure that's been red-listing
+	// release-build runs.
+	//
+	// LIMIT: tests that restore package-level overrides via classic
+	// `defer` (prepareTaskWorktree, etc.) still race the worker —
+	// defers fire BEFORE t.Cleanup, so the worker is still alive when
+	// the override restorer runs. That's pre-existing and gated by
+	// CI's `-race` carve-out for internal/team. Convert those tests'
+	// `defer restore()` to `t.Cleanup(restore)` to fix the residual
+	// race; not done here to keep this PR scoped.
```
FranDias, line 887: **SUGGESTION**: the comment notes the residual `defer restore()` race in tests like `prepareTaskWorktree` callers, gated by the `-race` carve-out. Worth filing a follow-up issue and linking it here so the carve-out lifts cleanly when those `defer`s convert to `t.Cleanup`. As written the limitation lives only in this comment and will get lost.

PR #321, opened 2026-04-26
fix(web): guard setTheme localStorage write against private-browsing throw

FranDias:
## Why

Surfaced by staff-reviewer audit on the dark-mode landing (#306, commit fe4d6f91). Safari private browsing and sandboxed-iframe contexts throw on `localStorage` writes. The **read path** in `web/src/stores/app.ts` already wraps `localStorage.getItem` in try/catch (line 6-9); the **write path** in `setTheme` did not, so any toggle of the dark-mode header switch in those contexts threw uncaught.

## Fix

Same try/catch shape, same intent: persistence failure shouldn't prevent the in-session DOM + store update from landing. User still gets the visible toggle for the current session — the preference just isn't remembered across reloads in storage-blocked contexts.

```diff
   setTheme: (t) => {
+    try {
       localStorage.setItem("wuphf-theme", t);
+    } catch {}
     document.documentElement.setAttribute("data-theme", t);
     set({ theme: t });
   },
```

## Test plan

- [x] `bun run typecheck` clean.
- [x] `bun run build` clean (1.28s).
- [ ] CI green.
- [ ] Manual: open Safari private window, toggle dark mode → no thrown errors, theme applies.

### DIFF
--- a/web/src/stores/app.ts
+++ b/web/src/stores/app.ts
@@ -174,7 +174,13 @@ export const useAppStore = create<AppStore>((set, get) => ({
 
   theme: _storedTheme,
   setTheme: (t) => {
-    localStorage.setItem("wuphf-theme", t);
+    // Same try/catch shape as the read path above. Safari private browsing
+    // and sandboxed-iframe contexts both throw on localStorage writes; the
+    // toggle should still update the DOM + store even if persistence fails,
+    // so the user gets the visible state change for the current session.
+    try {
+      localStorage.setItem("wuphf-theme", t);
+    } catch {}
     document.documentElement.setAttribute("data-theme", t);
     set({ theme: t });
   },
--- a/web/src/stores/app.ts
+++ b/web/src/stores/app.ts
@@ -178,9 +178,16 @@ export const useAppStore = create<AppStore>((set, get) => ({
     // and sandboxed-iframe contexts both throw on localStorage writes; the
     // toggle should still update the DOM + store even if persistence fails,
     // so the user gets the visible state change for the current session.
+    // console.warn keeps a breadcrumb so a user reporting "theme doesn't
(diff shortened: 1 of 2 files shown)


web/src/stores/app.ts
```diff
@@ -174,7 +174,13 @@ export const useAppStore = create<AppStore>((set, get) => ({
...
   setTheme: (t) => {
-    localStorage.setItem("wuphf-theme", t);
+    // Same try/catch shape as the read path above. Safari private browsing
+    // and sandboxed-iframe contexts both throw on localStorage writes; the
+    // toggle should still update the DOM + store even if persistence fails,
+    // so the user gets the visible state change for the current session.
+    try {
+      localStorage.setItem("wuphf-theme", t);
+    } catch {}
     document.documentElement.setAttribute("data-theme", t);
```
FranDias, line 191: **SUGGESTION**: fix is correct — symmetrical with the read path at line 7. No existing Playwright/vitest covers this path (searched `web/e2e/tests/` + `tests/e2e/` for `setTheme`, `wuphf-theme`, `data-theme`, `QuotaExceeded`, `SecurityError` — zero hits). 

Low-cost regression test would be a vitest unit on the store: stub `localStorage.setItem` to throw, call `setTheme('dark')`, assert `document.documentElement.getAttribute('data-theme')` is `'dark'` and `useAppStore.getState().theme === 'dark'`. ~10 LOC. Worth landing alongside this PR rather than as a follow-up — a future refactor could regress this without anyone noticing because the throw only fires for users in private browsing.

Also: the empty `catch {}` swallows the original cause silently. Even a single-line `console.warn("theme persistence failed:", e)` keeps DevTools breadcrumbs for users who report "my theme doesn't stick" — same shape as the Go-side `log.Printf` change in #322.

PR #322, opened 2026-04-26
fix(team): surface commitBoundsByPath errors instead of swallowing

FranDias:
## Why

Surfaced by staff-reviewer audit on the slow-DM-and-wiki-loading PR (#312, commit 8135e2e2).

`wiki_article.go:305` (`BuildBacklinks`) and `wiki_sections.go:156` (`materializeSection`) both swallowed `commitBoundsByPath` errors with the bare-`_` pattern:

```go
commitBounds, _ := repo.commitBoundsByPath(ctx)
```

On a corrupt wiki repo or transient `git log` failure, the catalog/sections rendered with no author + no timestamps and no signal anywhere — operators got a blank-looking UI and had to guess.

## Fix

Both sites now log at warn level via the stdlib `log` package (matches `wiki_sections.go`'s existing import). The empty-map fallback is preserved so the request still completes — visible behavior on the happy path is unchanged, only error visibility improves.

```diff
-	commitBounds, _ := repo.commitBoundsByPath(ctx)
+	commitBounds, err := repo.commitBoundsByPath(ctx)
+	if err != nil {
+		log.Printf("...: commitBoundsByPath failed, ... will be empty: %v", err)
+	}
```

A third call site at `wiki_article.go:137` inside an `if bounds, err := r.commitBoundsByPath(ctx); err == nil { ... }` already handles the error correctly (falls through to the no-bounds branch). Not changed.

## Test plan

- [x] `go test -run "Wiki|Backlink|Section" ./internal/team/` clean (8.6s).
- [x] `go build ./...` clean.
- [ ] CI green.
- [ ] Manual: introduce a transient `git log` failure on a wiki repo and confirm the warn line surfaces in logs (deferred — needs corrupt-repo fixture; out of scope for this one-line surface fix).

### DIFF
--- a/internal/team/wiki_article.go
+++ b/internal/team/wiki_article.go
@@ -36,6 +36,7 @@ package team
 import (
 	"context"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -301,8 +302,13 @@ func (r *Repo) backlinksFor(ctx context.Context, target string) ([]Backlink, err
 	sort.Slice(hits, func(i, j int) bool { return hits[i].relPath < hits[j].relPath })
 
 	// Fill author_slug per article from git log. Best-effort: if log fails,
-	// leave author_slug empty rather than abort.
-	commitBounds, _ := r.commitBoundsByPath(ctx)
+	// leave author_slug empty rather than abort. Surface the underlying
+	// error at warn level — silent swallow used to mask corrupt-repo and
+	// `git log` failures so backlinks rendered with no author + no signal.
+	commitBounds, err := r.commitBoundsByPath(ctx)
+	if err != nil {
+		log.Printf("wiki backlinks: commitBoundsByPath failed, author_slug fields will be empty: %v", err)
+	}
 	backs := make([]Backlink, 0, len(hits))
 	for _, h := range hits {
 		author := ""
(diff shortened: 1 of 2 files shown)


internal/team/wiki_article.go
```diff
@@ -301,8 +302,13 @@ func (r *Repo) backlinksFor(ctx context.Context, target string) ([]Backlink, err
...
 
 	// Fill author_slug per article from git log. Best-effort: if log fails,
-	// leave author_slug empty rather than abort.
-	commitBounds, _ := r.commitBoundsByPath(ctx)
+	// leave author_slug empty rather than abort. Surface the underlying
+	// error at warn level — silent swallow used to mask corrupt-repo and
+	// `git log` failures so backlinks rendered with no author + no signal.
+	commitBounds, err := r.commitBoundsByPath(ctx)
+	if err != nil {
+		log.Printf("wiki backlinks: commitBoundsByPath failed, author_slug fields will be empty: %v", err)
```
FranDias, line 310: Confirmed `log.Printf` matches the `internal/team/` convention (entity_facts, broker_pam, pam, broker_playbook, wiki_lookup, wiki_dlq, entity_synthesizer all use stdlib `log`). No slog/zap in this package. Import is stdlib `log` — verified.

**SUGGESTION**: prefix is inconsistent across the package — some lines use `"pam:"`, `"entity facts:"`, etc. Yours uses `"wiki backlinks:"` which is fine, but `wiki_sections.go:462` already exists with `"wiki sections: discover failed"` so the new line at 162 follows that prefix. Consistent. No action needed.

PR #324, opened 2026-04-26
test(team): atomic.Pointer worktree seams + re-enable -race

FranDias:
## Summary

Worktree dispatch seams (`prepareTaskWorktree`, `cleanupTaskWorktree`, `taskWorktreeRootDir`, `verifyTaskWorktreeWritable`, `allowRealTaskWorktree`) were package-level vars that tests bare-assigned with defer-restore. Background goroutines spawned by the headless codex queue worker read those vars after a test's deferred restore had already run, racing on the function-pointer write. `go test -race` catches it on `TestRunHeadlessCodexQueueRetriesLocalWorktreeAfterGenericError`.

This PR mirrors the existing `headlessCodexRunTurnOverride` pattern (already in `headless_codex.go`):
- Each seam becomes `atomic.Pointer[fnType]` accessed via a thin wrapper function in `worktree.go`.
- Tests migrate to `setXForTest(t, fn)` helpers in `test_support.go`. Each registers `t.Cleanup` to restore the prior override.
- `allowRealTaskWorktree` becomes `atomic.Bool`.
- CI matrix's race-flag exclusion for `internal/team` is dropped — only `internal/teammcp` remains opted-out.

## Verification

- `go test -race -timeout 5m -count=1 ./internal/team/` → **pass** (149.8s).
- `go test -race -timeout 5m -run TestRunHeadlessCodexQueueRetriesLocalWorktreeAfterGenericError ./internal/team/` → **pass** (the original race target).
- `go vet ./...`, `gofmt -l`, `golangci-lint`, lefthook pre-commit + pre-push all clean.

## Followups (separate PRs)

- `internal/teammcp` still excluded from `-race`; it has its own shared-state path (different vars).
- Other function-pointer seams in `internal/team` (`graphRecordFactRefs`, `studioPackageGenerator`, `headlessCodexWorkspaceStatusSnapshot`, `launcherSendNotificationToPane`, `listHeadlessTaskRunnerProcesses`, `killHeadlessTaskRunnerProcess`) follow the same recipe but were not raced by any current test. Incremental sweep.

## Test plan

- [ ] CI passes including new `internal/team` race leg
- [ ] Spot-check a couple of converted tests still exercise the intended override (e.g. `TestBrokerReleaseTaskCleansWorktree` still observes the cleaned path/branch via the closure)

## Release Notes

* **Tests**
  * Refactored test infrastructure to enhance thread-safety and prevent race conditions during concurrent test execution through improved helper functions.
  * Updated worktree and workspace status testing with safer, concurrent-safe override mechanisms.

* **Chores**
  * Refined CI workflow test matrix configuration to optimize race condition detection coverage.

### DIFF
--- a/internal/team/headless_codex_test.go
+++ b/internal/team/headless_codex_test.go
@@ -262,7 +262,6 @@ func TestRunHeadlessCodexTurnUsesAssignedWorktreeForCodingAgents(t *testing.T) {
 	oldLookPath := headlessCodexLookPath
 	oldExecutablePath := headlessCodexExecutablePath
 	oldCommandContext := headlessCodexCommandContext
-	oldPrepareTaskWorktree := prepareTaskWorktree
 	headlessCodexLookPath = func(file string) (string, error) {
 		switch file {
 		case "codex":
@@ -277,14 +276,13 @@ func TestRunHeadlessCodexTurnUsesAssignedWorktreeForCodingAgents(t *testing.T) {
 		cmdArgs = append(cmdArgs, args...)
 		return exec.CommandContext(ctx, os.Args[0], cmdArgs...)
 	}
-	prepareTaskWorktree = func(taskID string) (string, string, error) {
+	setPrepareTaskWorktreeForTest(t, func(taskID string) (string, string, error) {
 		return worktreeDir, worktreeBranchName(taskID), nil
-	}
+	})
 	defer func() {
 		headlessCodexLookPath = oldLookPath
 		headlessCodexExecutablePath = oldExecutablePath
 		headlessCodexCommandContext = oldCommandContext
-		prepareTaskWorktree = oldPrepareTaskWorktree
 	}()
 
 	t.Setenv("GO_WANT_HEADLESS_CODEX_HELPER_PROCESS", "1")
@@ -385,7 +383,6 @@ func TestRunHeadlessCodexTurnUsesAssignedWorktreeForLocalWorktreeBuilder(t *test
 	oldLookPath := headlessCodexLookPath
 	oldExecutablePath := headlessCodexExecutablePath
 	oldCommandContext := headlessCodexCommandContext
-	oldPrepareTaskWorktree := prepareTaskWorktree
 	headlessCodexLookPath = func(file string) (string, error) {
 		switch file {
 		case "codex":
@@ -400,14 +397,13 @@ func TestRunHeadlessCodexTurnUsesAssignedWorktreeForLocalWorktreeBuilder(t *test
 		cmdArgs = append(cmdArgs, args...)
 		return exec.CommandContext(ctx, os.Args[0], cmdArgs...)
 	}
-	prepareTaskWorktree = func(taskID string) (string, string, error) {
+	setPrepareTaskWorktreeForTest(t, func(taskID string) (string, string, error) {
 		return worktreeDir, worktreeBranchName(taskID), nil
-	}
+	})
 	defer func() {
 		headlessCodexLookPath = oldLookPath
 		headlessCodexExecutablePath = oldExecutablePath
 		headlessCodexCommandContext = oldCommandContext
-		prepareTaskWorktree = oldPrepareTaskWorktree
 	}()
 
 	t.Setenv("GO_WANT_HEADLESS_CODEX_HELPER_PROCESS", "1")
@@ -1902,11 +1898,10 @@ func TestBeginHeadlessCodexTurnCapturesWorktreeForLocalWorktreeBuilder(t *testin
 	t.Setenv("HOME", t.TempDir())
 
 	worktreeDir := t.TempDir()
-	oldPrepareTaskWorktree := prepareTaskWorktree
 	oldSnapshot := headlessCodexWorkspaceStatusSnapshot
-	prepareTaskWorktree = func(taskID string) (string, string, error) {
+	setPrepareTaskWorktreeForTest(t, func(taskID string) (string, string, error) {
 		return worktreeDir, worktreeBranchName(taskID), nil
-	}
+	})
 	headlessCodexWorkspaceStatusSnapshot = func(path string) string {
 		if !samePath(path, worktreeDir) {
 			t.Fatalf("expected workspace snapshot to target %q, got %q", worktreeDir, path)
@@ -1914,7 +1909,6 @@ func TestBeginHeadlessCodexTurnCapturesWorktreeForLocalWorktreeBuilder(t *testin
 		return "snapshot"
 	}
 	defer func() {
-		prepareTaskWorktree = oldPrepareTaskWorktree
 		headlessCodexWorkspaceStatusSnapshot = oldSnapshot
 	}()
 
@@ -1956,17 +1950,10 @@ func TestRunHeadlessCodexQueueRetriesLocalWorktreeAfterGenericError(t *testing.T
 	t.Setenv("HOME", t.TempDir())
 	tmpDir := t.TempDir()
 
-	oldPrepare := prepareTaskWorktree
-	oldCleanup := cleanupTaskWorktree
-	prepareTaskWorktree = func(taskID string) (string, string, error) {
+	setPrepareTaskWorktreeForTest(t, func(taskID string) (string, string, error) {
 		return filepath.Join(tmpDir, "wuphf-task-"+taskID), "wuphf-" + taskID, nil
-	}
-	cleanupTaskWorktree = func(string, string) error { return nil }
-	defer func() {
-		prepareTaskWorktree = oldPrepare
-		cleanupTaskWorktree = oldCleanup
-	}()
-
+	})
+	setCleanupTaskWorktreeForTest(t, func(string, string) error { return nil })
 	setHeadlessWakeLeadFn(t, func(_ *Launcher, _ string) {})
 
 	b := NewBrokerAt(filepath.Join(tmpDir, "broker-state.json"))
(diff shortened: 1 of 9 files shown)


internal/team/headless_codex_test.go
```diff
@@ -1902,19 +1898,17 @@ func TestBeginHeadlessCodexTurnCapturesWorktreeForLocalWorktreeBuilder(t *testin
...
 
 	worktreeDir := t.TempDir()
-	oldPrepareTaskWorktree := prepareTaskWorktree
 	oldSnapshot := headlessCodexWorkspaceStatusSnapshot
-	prepareTaskWorktree = func(taskID string) (string, string, error) {
+	setPrepareTaskWorktreeForTest(t, func(taskID string) (string, string, error) {
 		return worktreeDir, worktreeBranchName(taskID), nil
-	}
+	})
 	headlessCodexWorkspaceStatusSnapshot = func(path string) string {
```
FranDias: **MEDIUM — leftover bare-assign on a goroutine-read seam**

`headlessCodexWorkspaceStatusSnapshot` is read at `headless_codex.go:537` inside `headlessTurnCompletedDurably`, which is called from `runHeadlessCodexQueue` (line 433) — i.e. the exact queue-worker goroutine path this PR is hardening against. Yet this test still does `oldSnapshot := headlessCodexWorkspaceStatusSnapshot` + `defer ... = oldSnapshot` (also at lines 1651, 1700, 1743 in `TestHeadlessTurnCompletedDurably*` tests).

The PR description acknowledges these as "not raced by any current test" follow-ups, but with `internal/team` re-entering the `-race` matrix, any leaked queue worker from a prior test in the same package process can now race the assignment/restore. `TestRunHeadlessCodexQueueRetriesLocalWorktreeAfterGenericError` does call `spawnHeadlessWorker("eng")` (line 2011), so a leaked goroutine reaching this codepath is plausible.

The headline test (`TestRunHeadlessCodexQueueRetriesLocalWorktreeAfterGenericError`) is the one the PR description names as the original race trigger — re-introducing -race for this package while leaving an adjacent seam unmigrated risks chasing the same flake under a different name in CI within a week.

**Fix:** Migrate `headlessCodexWorkspaceStatusSnapshot` in this PR (it's a one-liner identical to the four already converted: declare `headlessCodexWorkspaceStatusSnapshotFn` type + `atomic.Pointer` override + wrapper + `setHeadlessCodexWorkspaceStatusSnapshotForTest` helper). The recipe is right next to it in `test_support.go:98-105` — adding it now is cheaper than handling the flake later. If you'd rather scope-limit, at minimum add a `// TODO(team-race): migrate after PR #324 lands` and link a tracking issue so it doesn't slip.

internal/team/worktree_guard_test.go
```diff
@@ -62,26 +65,25 @@ func stubPrepareTaskWorktree(taskID string) (string, string, error) {
...
-	prevCleanup := cleanupTaskWorktree
-	allowRealTaskWorktree = true
-	prepareTaskWorktree = defaultPrepareTaskWorktree
-	cleanupTaskWorktree = defaultCleanupTaskWorktree
+	prevAllow := allowRealTaskWorktree.Load()
+	prevPrepare := prepareTaskWorktreeOverride.Load()
+	prevCleanup := cleanupTaskWorktreeOverride.Load()
+	allowRealTaskWorktree.Store(true)
+	prepareTaskWorktreeOverride.Store(nil)
+	cleanupTaskWorktreeOverride.Store(nil)
```
FranDias, line 89: **MEDIUM — race window between `allowRealTaskWorktree.Store(true)` and the override-`Store(nil)` calls**

The sequence here is:
```
allowRealTaskWorktree.Store(true)        // line 81
prepareTaskWorktreeOverride.Store(nil)   // line 82
cleanupTaskWorktreeOverride.Store(nil)   // line 83
```

Any goroutine leaked from a prior test that's still calling `prepareTaskWorktree(...)` will, in the window between line 81 and line 82, observe `allowRealTaskWorktree=true` AND `override=stub` — that's harmless (stub wins). But the equally narrow window between line 82 and line 83 lets a goroutine call `prepareTaskWorktree` (override=nil → `defaultPrepareTaskWorktree` → checks `allowRealTaskWorktree=true` → proceeds to real `git worktree add`) **before** `cleanupTaskWorktreeOverride` is nilled. In practice this is fine because (a) `worktree_test.go` callers don't have a Broker spinning queue workers, and (b) the comment above already documents that parallel/leaked-worker tests must not opt in. But the new comment on lines 70-75 oversells the safety: it says "the mutate-and-restore is race-safe" — which is true for the atomic ops themselves but not for the *composite* state (`allow` + override pair) that callers actually depend on.

Not a blocker, but the wording is misleading enough that a future contributor may add `t.Parallel()` to a `worktree_test.go` test trusting the new safety claim. 

**Fix:** Tighten the comment to something like: "Each store is atomic, but the (allowRealTaskWorktree, prepare/cleanup overrides) triple is not coherent across stores — callers must still avoid `t.Parallel()` and ensure no background brokers/workers can read these seams during the test." Same stricter wording on cleanup ordering.

internal/team/test_support.go
```diff
@@ -87,3 +90,49 @@ func setHeadlessCodexRunTurnForTest(t *testing.T, fn func(l *Launcher, ctx conte
...
+// setHeadlessCodexRunTurnForTest: the headless queue worker can read the
+// dispatcher after the test's deferred restore has already run.
+func setPrepareTaskWorktreeForTest(t *testing.T, fn prepareTaskWorktreeFn) {
+	t.Helper()
+	prior := prepareTaskWorktreeOverride.Load()
+	prepareTaskWorktreeOverride.Store(&fn)
+	t.Cleanup(func() {
+		prepareTaskWorktreeOverride.Store(prior)
+	})
+}
```
FranDias, line 113: **SUGGESTION — undocumented constraint: not safe under `t.Parallel()`**

`prior := X.Load()` followed by `X.Store(&fn)` is not atomic as a pair. If two parallel tests both call `setPrepareTaskWorktreeForTest`, the snapshots can interleave so the cleanup of the second test restores the first test's stub *over* the first test's still-active stub — and worse, the cleanup of the first test then restores the original stub even though no test owns the override anymore.

Currently this is fine because no test in `internal/team` calls `t.Parallel()` (verified across `broker_test.go`, `headless_codex_test.go`, `launcher_test.go`, `worktree_test.go` — zero hits). But there's nothing in the helper telling a future author this. The same applies to the other three setters and to the pre-existing `setHeadlessCodexRunTurnForTest`.

**Fix:** Add a one-line constraint comment on each setter (or one shared comment above the cluster):
```go
// NOT safe under t.Parallel() within the same package: the Load/Store
// pair is not atomic, so concurrent setters can scramble the cleanup chain.
```
If you want belt-and-braces, gate it: `if t != nil && hasParent && testingInternalParallel(t) { t.Fatal(...) }` — but a comment is sufficient given the team's small size.

internal/team/worktree.go
```diff
@@ -10,22 +10,71 @@ import (
...
 // codepath (always against a tempdir-rooted repo) opt in via
 // allowRealTaskWorktreeForTest(t), which scopes the re-enable to one test.
-var allowRealTaskWorktree = true
+//
+// atomic.Bool because the production read in defaultPrepareTaskWorktree
+// happens on goroutines spawned by the headless queue worker, which can
+// outlive a test's deferred restore.
+var allowRealTaskWorktree atomic.Bool
+
+func init() { allowRealTaskWorktree.Store(true) }
```
FranDias, line 77: **NIT — init() vs. file-scope assignment**

Replacing `var allowRealTaskWorktree = true` with `var allowRealTaskWorktree atomic.Bool` + `func init() { allowRealTaskWorktree.Store(true) }` is correct, but the init() introduces an init-order dependency: any other init() in the package that reads `allowRealTaskWorktree.Load()` is now sensitive to source-file ordering (alphabetical: `worktree.go` sorts after most files, so it would run *after* their inits and they'd see `false`).

A quick grep shows no in-package init reads it, and `worktree_guard_test.go`'s test-only init explicitly calls `Store(false)` so it's fine. Just flagging because this is an easy footgun: if anyone adds a package-level init in (say) `broker.go` that constructs a Broker which transitively touches `defaultPrepareTaskWorktree`, they'd hit `errRealTaskWorktreeDisabled` in production with no obvious cause.

**Fix:** Optional — declare the atomic + initializer together in a single statement (Go 1.20+ doesn't permit `atomic.Bool{v:1}` literal, so init() is the cleanest option you have). Alternatively, add a `// init order: must run before any package-level Broker construction` comment so the constraint is documented.

internal/team/worktree.go
```diff
@@ -10,22 +10,71 @@ import (
...
 
-var prepareTaskWorktree = defaultPrepareTaskWorktree
-var cleanupTaskWorktree = defaultCleanupTaskWorktree
-var taskWorktreeRootDir = defaultTaskWorktreeRootDir
-var verifyTaskWorktreeWritable = defaultVerifyTaskWorktreeWritable
+// Test seams for the worktree dispatch path. These read from background
+// goroutines (headless codex queue worker, launcher recovery) while tests
+// install per-test stubs, so the seams must be race-safe. Tests must use
+// the setXForTest helpers in test_support.go — never assign to the
+// override pointers directly. Same shape as headlessCodexRunTurnOverride.
```
FranDias, line 22: **NIT — comment overstates which seams are migrated**

The comment says "These read from background goroutines (headless codex queue worker, launcher recovery) while tests install per-test stubs, so the seams must be race-safe" — but `headlessCodexWorkspaceStatusSnapshot` is in the same boat (read from `runHeadlessCodexQueue` via `headlessTurnCompletedDurably`) and is left bare-assigned (see comment on `headless_codex_test.go:1905`).

If you choose not to migrate the snapshot seam in this PR, this comment should call out that the migration is partial and link a tracking issue.

FranDias (commented):
Solid migration — atomic.Pointer wrappers correctly mirror the existing `headlessCodexRunTurnOverride` shape, and the mechanical rewrite preserves closure-captured state (`cleanedPath`/`cleanedBranch`/`prepareCalls`/`cleaned`/`cleanupCalls` all still close over the right vars). CI matrix regex `(test("/internal/teammcp($|/)") | not)` correctly distinguishes `internal/team` (race=true) from `internal/teammcp` (race=false) since `team` doesn't contain the substring `teammcp`. `Store(nil)` on `atomic.Pointer[T]` is well-defined and the wrapper's nil-check fall-through is correct. init() ordering is also fine — non-test `worktree.go` init runs before `worktree_guard_test.go` init under `go test`, and only `worktree.go` runs in production.

A few things worth addressing before merge — most importantly that `headlessCodexWorkspaceStatusSnapshot` is read from the same queue-worker goroutine the PR is hardening against, but is left bare-assigned in 4+ tests under a package that now runs with `-race`. With the matrix change live, that seam is the next likely flake. Inline comments below.

PR #325, opened 2026-04-26
chore(lint): tighten golangci-lint coverage

FranDias:
## Summary

Adds `bodyclose`, `errorlint`, and `govet` shadow to the lint suite. Every finding is fixed as a genuine code change — no `//nolint:` suppressions.

## Per-fix

| File | Fix |
| --- | --- |
| `cmd/wuphf/channel_integration.go` | `errors.As` replacing type assertion against wrapped `*api.AuthError` |
| `cmd/wuphf/channel_member_draft.go` | drain + close response body on the unused-response branch |
| `internal/action/one.go` | `errors.As` + `errors.Is(execErr.Err, exec.ErrNotFound)` |
| `internal/agent/service.go` | rename shadowed `current`/`ok` to `cur`/`stillRegistered` |
| `internal/commands/dispatch.go` | `errors.Is(err, ErrQuit)` instead of `==` |
| `internal/openclaw/client.go` | drain + close handshake response body |
| `internal/operations/loader.go` | `%w` in `fmt.Errorf` to keep the cause chain unwrappable |
| `internal/team/wiki_index_bleve.go` | `%w` |
| `internal/team/wiki_index_sqlite.go` | `errors.Is(err, sql.ErrNoRows)` |
| `internal/team/wiki_lint.go` | `errors.Is(err, filepath.SkipAll)` |
| `internal/teammcp/server_test.go` | rename inner `got` to `found` |

## Linters deliberately NOT added

Each has too many existing findings to land here. Followups, one PR each:

- `contextcheck` — 26 findings (ctx threading)
- `gocritic` — 40 findings (mostly diagnostic-tag style nits)
- `nilerr` — 29 findings (returning nil after non-nil err)
- `noctx` — 95 findings (`net/http.{Get,Post}` and `exec.Command` without ctx)

`govet` shadow has an exclusion for `err` (idiomatic `if err := ...` always shadows). `errorlint`, `bodyclose` excluded on `_test.go`.

## Verification

- `golangci-lint run --timeout=5m ./...` → **0 issues**
- `go build ./...`, `go vet ./...`, `gofmt -l .` → all clean
- `go test -count=1 -timeout 4m` on touched packages (team, agent, teammcp, cmd/wuphf, action, commands, openclaw, operations) → all pass

## Test plan

- [ ] CI lint job stays green
- [ ] Touched packages still pass tests in CI

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error detection to correctly handle wrapped errors in authentication checks, quit conditions, and database operations, ensuring more robust error recognition across the application.
  * Fixed resource cleanup by explicitly closing HTTP response bodies, preventing potential resource leaks.

* **Chores**
  * Enhanced linter configuration to enforce additional code quality checks.
  * Refined test variable naming for improved readability.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,16 +8,26 @@ issues:
 linters:
   default: none
   enable:
+    - bodyclose
     - errcheck
+    - errorlint
     - govet
     - ineffassign
-    - staticcheck
     - misspell
+    - staticcheck
     - unconvert
+    # Followups (own PR each — too many existing findings to land here):
+    #   - contextcheck (26 findings: ctx threading)
+    #   - gocritic (40 findings: mostly diagnostic-tag style nits)
+    #   - nilerr (29 findings: returning nil after non-nil err)
+    #   - noctx (95 findings: net/http and exec.Command without context)
   settings:
     errcheck:
       exclude-functions:
         - (io.Closer).Close
+    govet:
+      enable:
+        - shadow
     staticcheck:
       # Disable the QF* quickfix suggestions (pure style hints) and the
       # ST1000/ST1003/ST1021 naming/comment rules (pedantic, massive scope).
@@ -37,6 +47,14 @@ linters:
       - path: _test\.go
         linters:
           - errcheck
+          - bodyclose
+          - errorlint
+      # govet's shadow check on `err` is famously noisy in idiomatic Go
+      # (every nested `if err := ...; err != nil` shadows). Keep shadow on
+      # for non-err identifiers (real bugs hide there) but mute the err case.
+      - linters:
+          - govet
+        text: 'shadow: declaration of "err"'
 formatters:
   enable:
     - gofmt
--- a/cmd/wuphf/channel_member_draft.go
+++ b/cmd/wuphf/channel_member_draft.go
@@ -267,7 +267,9 @@ func mutateOfficeMemberSpec(draft channelMemberDraft, activeChannel string) tea.
 			addReq, err := newBrokerRequest(http.MethodPost, "http://127.0.0.1:7890/channel-members", bytes.NewReader(addBody))
 			if err == nil {
 				addReq.Header.Set("Content-Type", "application/json")
-				_, _ = client.Do(addReq)
+				if resp, doErr := client.Do(addReq); doErr == nil {
+					_ = resp.Body.Close()
+				}
 			}
 		}
 		l, err := team.NewLauncher("")
--- a/cmd/wuphf/channel_member_draft.go
+++ b/cmd/wuphf/channel_member_draft.go
@@ -267,8 +267,8 @@ func mutateOfficeMemberSpec(draft channelMemberDraft, activeChannel string) tea.
 			addReq, err := newBrokerRequest(http.MethodPost, "http://127.0.0.1:7890/channel-members", bytes.NewReader(addBody))
 			if err == nil {
 				addReq.Header.Set("Content-Type", "application/json")
-				if resp, doErr := client.Do(addReq); doErr == nil {
-					_ = resp.Body.Close()
+				if addResp, doErr := client.Do(addReq); doErr == nil {
+					_ = addResp.Body.Close()
 				}
 			}
 		}
--- a/internal/action/one.go
+++ b/internal/action/one.go
@@ -635,7 +635,8 @@ func (o *OneCLI) run(ctx context.Context, args []string) ([]byte, error) {
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 	if err := cmd.Run(); err != nil {
-		if ee, ok := err.(*exec.Error); ok && ee.Err == exec.ErrNotFound {
+		var execErr *exec.Error
+		if errors.As(err, &execErr) && errors.Is(execErr.Err, exec.ErrNotFound) {
 			return nil, fmt.Errorf("one CLI not found. Install One, make npx available, or set WUPHF_ONE_BIN")
 		}
 		msg := strings.TrimSpace(stderr.String())
--- a/internal/agent/service.go
+++ b/internal/agent/service.go
@@ -373,7 +373,7 @@ func (s *AgentService) runAgentWorker(slug string, ma *ManagedAgent, stopCh <-ch
 
 		if !running {
 			s.mu.Lock()
-			if current, ok := s.agents[slug]; ok && current == ma {
+			if cur, stillRegistered := s.agents[slug]; stillRegistered && cur == ma {
 				delete(s.tickTimers, slug)
 			}
 			s.mu.Unlock()
--- a/internal/agent/service.go
+++ b/internal/agent/service.go
@@ -348,7 +348,7 @@ func (s *AgentService) runAgentWorker(slug string, ma *ManagedAgent, stopCh <-ch
 
 		if shouldStop {
 			s.mu.Lock()
-			if current, ok := s.agents[slug]; ok && current == ma {
+			if cur, stillRegistered := s.agents[slug]; stillRegistered && cur == ma {
 				delete(s.tickTimers, slug)
 			}
 			s.mu.Unlock()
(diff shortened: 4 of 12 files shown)


FranDias (commented):
Tightening lint coverage with bodyclose / errorlint / govet-shadow plus genuine code fixes (no `//nolint:` suppressions). Verified the PR's claim of `golangci-lint run ./... → 0 issues` locally on the head SHA. Each fix is correct on the merits — the `errors.As` widening (`channel_integration.go`, `action/one.go`) is safe because the upstream stdlib paths don't wrap today, the `bodyclose` placements work against all three `gorilla/websocket.Dialer.DialContext` return shapes (verified against v1.5.3 source), and the regex-based `err` shadow exclusion is precise (govet only emits `shadow: declaration of "err"` when the literal identifier is `err`, so non-err names won't be silently masked).

No blocking issues. A few consistency / readability nits inline. Approving conceptually but leaving as comment so author can decide on the nits.

internal/team/wiki_index_sqlite.go
```diff
@@ -285,7 +286,7 @@ func (s *SQLiteFactStore) GetFact(ctx context.Context, id string) (TypedFact, bo
 		        created_at, created_by, reinforced_at
 		 FROM facts WHERE id = ?`, id)
 	f, err := scanFact(row)
-	if err == sql.ErrNoRows {
+	if errors.Is(err, sql.ErrNoRows) {
```
FranDias, line 289: **SUGGESTION (consistency)**: this `errors.Is(err, sql.ErrNoRows)` rewrite is correct, but the analogous comparison at **line 510** in the same file (`if err == sql.ErrNoRows` after `.Scan(&to)` in `ResolveRedirect`) is unchanged.

`errorlint` doesn't flag line 510 because go-errorlint's `allowed.go:44` explicitly allowlists `database/sql.ErrNoRows` against `(*sql.Row).Scan` — so technically both are fine. But applying the rewrite to one and not the other invites the next reader to wonder which is intentional. Either rewrite line 510 too, or leave both as `==` since the linter tolerates the pattern.

```go
// line 510
if err == sql.ErrNoRows {
```

internal/agent/service.go
```diff
@@ -348,7 +348,7 @@ func (s *AgentService) runAgentWorker(slug string, ma *ManagedAgent, stopCh <-ch
 
 		if shouldStop {
 			s.mu.Lock()
-			if current, ok := s.agents[slug]; ok && current == ma {
+			if cur, stillRegistered := s.agents[slug]; stillRegistered && cur == ma {
```
FranDias, line 351: **SUGGESTION (consistency)**: the rename of `current`/`ok` → `cur`/`stillRegistered` here fixes the shadow govet flagged. The same shadow pattern still exists ~25 lines below at **line 376** (`if current, ok := s.agents[slug]; ok && current == ma`), which shadows the outer `current, ok` declared at line 338.

govet's shadow analyzer happens to be conservative about if-init scope and doesn't flag line 376 today, so the lint stays clean — but the rename only addresses one of two structurally identical shadows. For consistency (and to survive a future ratchet to a stricter shadow analyzer like gocritic's `paramTypeCombine` or `nolintlint`), I'd apply the same `cur`/`stillRegistered` rename at line 376 too.

cmd/wuphf/channel_member_draft.go
```diff
@@ -267,7 +267,9 @@ func mutateOfficeMemberSpec(draft channelMemberDraft, activeChannel string) tea.
 			addReq, err := newBrokerRequest(http.MethodPost, "http://127.0.0.1:7890/channel-members", bytes.NewReader(addBody))
 			if err == nil {
 				addReq.Header.Set("Content-Type", "application/json")
-				_, _ = client.Do(addReq)
+				if resp, doErr := client.Do(addReq); doErr == nil {
```
FranDias: **NIT**: `resp` here shadows the outer `resp` declared at line 252 (which is `defer`-closed). govet's shadow check doesn't fire because of conservative if-init scope handling, but a one-character rename to `addResp` would (a) make the shadow explicit, (b) future-proof against a stricter shadow ratchet, and (c) match the naming used for `addBody`/`addReq` two lines above.

```go
if addResp, doErr := client.Do(addReq); doErr == nil {
    _ = addResp.Body.Close()
}
```

internal/openclaw/client.go
```diff
@@ -61,7 +61,10 @@ func Dial(ctx context.Context, cfg Config) (*Client, error) {
 	dialer := websocket.Dialer{HandshakeTimeout: cfg.DialTimeout}
 	dctx, cancel := context.WithTimeout(ctx, cfg.DialTimeout)
 	defer cancel()
-	conn, _, err := dialer.DialContext(dctx, cfg.URL, nil)
+	conn, resp, err := dialer.DialContext(dctx, cfg.URL, nil)
+	if resp != nil && resp.Body != nil {
+		_ = resp.Body.Close()
+	}
```
FranDias, line 67: **NIT (doc)**: this close is correct against all three return shapes of `gorilla/websocket@v1.5.3`'s `Dialer.DialContext`:
- success: `resp.Body` is replaced with `ioutil.NopCloser(bytes.NewReader([]byte{}))` (client.go:421), so the close is a no-op on a 0-byte buffer — fine.
- `ErrBadHandshake`: `resp.Body` is replaced with `ioutil.NopCloser(bytes.NewReader(buf[:n]))` for up to 1024 bytes (client.go:401-403) — fine.
- early errors: `resp` is `nil`, the guard skips it.

Worth a one-line comment so the next reader doesn't think the success-path close is freeing real network state (it isn't — gorilla swaps the body out before returning):

```go
// gorilla replaces resp.Body with a NopCloser before returning, so this
// close is a no-op on success but drains the failure-path debug buffer.
if resp != nil && resp.Body != nil {
    _ = resp.Body.Close()
}
```

.golangci.yml
```diff
@@ -37,6 +47,14 @@ linters:
...
         linters:
           - errcheck
+          - bodyclose
+          - errorlint
+      # govet's shadow check on `err` is famously noisy in idiomatic Go
+      # (every nested `if err := ...; err != nil` shadows). Keep shadow on
+      # for non-err identifiers (real bugs hide there) but mute the err case.
+      - linters:
+          - govet
+        text: 'shadow: declaration of "err"'
```
FranDias, line 57: **NIT (verified)**: confirmed this regex doesn't over-mask. govet's shadow analyzer formats the message as `shadow: declaration of %q shadows declaration at line %d` where `%q` is the literal identifier, so a variable named `errors` or `errCode` would emit a different `text` payload. The exclusion is identifier-anchored and won't hide shadows on differently-named vars.

No change needed — leaving this as a verification note for future readers wondering whether the regex is too broad.

internal/action/one.go
```diff
@@ -635,7 +635,8 @@ func (o *OneCLI) run(ctx context.Context, args []string) ([]byte, error) {
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 	if err := cmd.Run(); err != nil {
-		if ee, ok := err.(*exec.Error); ok && ee.Err == exec.ErrNotFound {
+		var execErr *exec.Error
+		if errors.As(err, &execErr) && errors.Is(execErr.Err, exec.ErrNotFound) {
```
FranDias, line 639: **SUGGESTION (semantic widening, intentional)**: this changes the match from "`err` is exactly `*exec.Error`" to "`err` is or wraps `*exec.Error`". `exec.Cmd.Run` in stdlib doesn't currently wrap, so behavior is identical today — but if the wuphf One CLI invocation ever gets wrapped through a helper (e.g. an `exec.CommandContext` middleware), `errors.As` will still find it where the type assertion would have silently fallen through to the generic `one CLI failed:` branch and lost the actionable "install One" message.

The widening is the correct direction. Worth noting in a comment that the user-visible message hinges on this match succeeding, so the wrap-tolerance matters.

PR #326, opened 2026-04-26
fix: replace http.DefaultClient with timed clients in poll functions

hobostay:
## Summary
- Replace `http.DefaultClient` with `&http.Client{Timeout: 2 * time.Second}` in `pollTasks`, `pollActions`, and `pollScheduler` in `cmd/wuphf/channel_broker.go`

## Problem
Three poll functions used `http.DefaultClient.Do(req)` which has no timeout, while all other poll functions in the same file (`pollBroker`, `pollMembers`, `pollOfficeMembers`, `pollChannels`, `pollSkills`, `pollSignals`, `pollDecisions`, `pollWatchdogs`) already use a 2-second timeout. Without a timeout, these goroutines hang indefinitely if the broker becomes unresponsive, leading to resource exhaustion over time.

## Test plan
- [x] Existing tests pass (code change is consistent with the pattern used by all other poll functions in the same file)
- [ ] Manual: stop the broker and verify that pollTasks/pollActions/pollScheduler return within 2 seconds instead of hanging

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of polling operations by implementing request timeouts to prevent hanging connections.

### DIFF
--- a/cmd/wuphf/channel_broker.go
+++ b/cmd/wuphf/channel_broker.go
@@ -414,7 +414,8 @@ func pollTasks(channel string) tea.Cmd {
 		if err != nil {
 			return channelTasksMsg{}
 		}
-		resp, err := http.DefaultClient.Do(req)
+		client := &http.Client{Timeout: 2 * time.Second}
+		resp, err := client.Do(req)
 		if err != nil {
 			return channelTasksMsg{}
 		}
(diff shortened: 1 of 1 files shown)


FranDias (approved):

FranDias (approved):
LGTM. Pure pattern alignment with the 14 sibling poll fns in this file already using `&http.Client{Timeout: 2 * time.Second}` (lines 77, 105, 133, 161, 194, 251, 286, 388, 441, 500, 522, 544, 590, 630, 652). The 3 patched callsites (417, 476, 566) were the only `http.DefaultClient` users left. Bounded failure under broker hang is the right behavior for these tea.Tick-driven polls.

Optional follow-up (do not block): hoisting a package-level `var brokerHTTPClient = &http.Client{Timeout: 2 * time.Second}` would stop allocating a client per poll tick. Out of scope for this PR.

Thanks for the careful, surgical fix.

PR #327, opened 2026-04-26
fix: replace http.DefaultClient with timed clients in session mode handlers

hobostay:
## Summary
- Replace `http.DefaultClient` with `&http.Client{Timeout: 3 * time.Second}` in `switchSessionMode` and `switchFocusMode` in `cmd/wuphf/channel.go`

## Problem
Two session mode handler functions used `http.DefaultClient.Do(req)` without a timeout. This is inconsistent with all other HTTP calls in the codebase (`channel_broker.go`, and the rest of `channel.go`) which use properly timed clients. Without a timeout, session mode switches and focus mode toggles hang indefinitely if the broker is unresponsive.

## Test plan
- [x] Code change is consistent with the existing pattern used by all other HTTP calls in the codebase
- [ ] Manual: verify `/collab` and `/focus` commands still work correctly

### DIFF
--- a/cmd/wuphf/channel.go
+++ b/cmd/wuphf/channel.go
@@ -5597,7 +5597,8 @@ func switchSessionMode(mode, agent string) tea.Cmd {
 		if err != nil {
 			return channelResetDoneMsg{err: err}
 		}
-		resp, err := http.DefaultClient.Do(req)
+		client := &http.Client{Timeout: 3 * time.Second}
+		resp, err := client.Do(req)
 		if err != nil {
 			return channelResetDoneMsg{err: err}
 		}
@@ -5651,7 +5652,8 @@ func switchFocusMode(enabled bool) tea.Cmd {
 		if err != nil {
 			return nil
 		}
-		resp, err := http.DefaultClient.Do(req)
+		client := &http.Client{Timeout: 3 * time.Second}
+		resp, err := client.Do(req)
 		if err != nil {
 			return nil
 		}


FranDias (approved):

cmd/wuphf/channel.go
```diff
@@ -5597,7 +5597,8 @@ func switchSessionMode(mode, agent string) tea.Cmd {
 		if err != nil {
 			return channelResetDoneMsg{err: err}
 		}
-		resp, err := http.DefaultClient.Do(req)
+		client := &http.Client{Timeout: 3 * time.Second}
```
FranDias, line 5600: Use `5 * time.Second` here, not `3 * time.Second`. Every nearby POST/write helper in this file uses 5s — `createSkill` (L5514), `invokeSkill` (L5530), `resetDMSession` (L5550). `switchSessionMode` is also a POST and triggers a non-trivial broker-side `ResetSession` + `ReconfigureSession`, so 3s would be more aggressive than the local convention and risk spurious failures on a slow broker.

The 2s timeout in `channel_broker.go` (PR #326) is for read polls, not writes — don't carry that over here.

Suggested change:
```suggestion
		client := &http.Client{Timeout: 5 * time.Second}
```

cmd/wuphf/channel.go
```diff
@@ -5651,7 +5652,8 @@ func switchFocusMode(enabled bool) tea.Cmd {
 		if err != nil {
 			return nil
 		}
-		resp, err := http.DefaultClient.Do(req)
+		client := &http.Client{Timeout: 3 * time.Second}
```
FranDias, line 5655: Same as the comment above — use 5s to match the local write-path convention.

```suggestion
		client := &http.Client{Timeout: 5 * time.Second}
```

_Pre-existing nit (not a blocker for this PR): `resp.Body.Close()` is called without `io.Copy(io.Discard, resp.Body)` first, so the connection won't be reused on keepalive. Low impact for a single-shot user command — acceptable to leave for a follow-up._

FranDias (dismissed):
Diff is correct and follows the local pattern. One thing to fix before merge: the 3s timeout doesn't match this file's convention for write/POST helpers. Inline comments on both new client allocations.

Thanks for catching this. Once the timeout is bumped to 5s, happy to re-approve.

FranDias:
Merged. Thanks @hobostay! Followed up with the 3s→5s alignment in #335 to match this file's existing write-path convention (see review notes above). Co-authored back to you.

PR #328, opened 2026-04-26
fix: sanitize markdown link URLs to prevent javascript: scheme XSS

hobostay:
## Summary
- Add `isSafeUrl()` allowlist to `formatInline()` in `web/src/lib/markdown.ts`
- Only `http://`, `https://`, `mailto:`, `/`, and `#` URLs are rendered as clickable links
- Dangerous schemes (`javascript:`, `data:`, `vbscript:`) render as inert `<span>` elements

## Problem
`formatInline()` applies `escapeHtml()` before regex replacements, which prevents HTML injection. However, the markdown link regex `[text](url)` directly interpolates the URL into the `href` attribute of an `<a>` tag. A markdown link like `[click](javascript:alert(1))` passes through `escapeHtml()` unchanged (no HTML special characters) and produces a clickable XSS vector:

```html
<a class="msg-link" href="javascript:alert(1)" target="_blank" rel="noopener">click</a>
```

While agent messages are from a trusted source (the broker), defense-in-depth is important — a compromised or misconfigured agent could emit malicious markdown.

## Test plan
- [x] Verified `http://` and `https://` links still render as clickable `<a>` tags
- [x] Verified `javascript:`, `data:`, and `vbscript:` URLs render as non-clickable `<span>` elements
- [ ] Manual: post `[click](javascript:alert(1))` in the chat and verify it's not clickable
- [ ] Manual: verify normal links like `[docs](https://example.com)` still work

### DIFF
--- a/web/src/lib/markdown.ts
+++ b/web/src/lib/markdown.ts
@@ -147,6 +147,17 @@ export function formatMarkdown(raw: string): string {
   return result.join("");
 }
 
+function isSafeUrl(url: string): boolean {
+  const trimmed = url.trim().toLowerCase();
+  return (
+    trimmed.startsWith("http://") ||
+    trimmed.startsWith("https://") ||
+    trimmed.startsWith("mailto:") ||
+    trimmed.startsWith("/") ||
+    trimmed.startsWith("#")
+  );
+}
+
 function formatInline(text: string): string {
   let s = escapeHtml(text);
   // Bold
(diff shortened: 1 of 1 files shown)


FranDias (approved):

FranDias:
Merged. Thanks @hobostay! Followed up with #334: tightening the protocol-relative `//` case, adding `noreferrer`, and a 11-spec regression suite at `web/src/lib/markdown.test.ts` so a future regex refactor can't silently regress this fix. Co-authored back to you.

PR #330, opened 2026-04-26
chore(lint): fix noctx findings + enable linter

FranDias:
> [!WARNING]
> **Stacked on #325** — merge that first or rebase to main when reviewing.
> Base is `improve/golangci-tighten`, not `main`. The diff against `main`
> will look noisy until #325 lands.

## Summary

Enables the `noctx` golangci-lint rule and fixes all 96 findings across 12 files. `noctx` flags `http.*` / `(*http.Client).*` / `exec.Command` calls without a context — they can't be cancelled when the parent goroutine exits, which is the root of several "broker won't shut down" / "stuck git" classes of bug we've hit.

## Per-file summary (96 findings)

| File | Count | Approach |
|---|---|---|
| `internal/team/launcher.go` | 63 | Bulk-converted `exec.Command` → `exec.CommandContext(context.Background(), …)`. All 63 are short-lived tmux housekeeping (`new-session`, `set-option`, `select-pane`, `respawn-pane`, `kill-session`, `send-keys`, `capture-pane`). Threading ctx through ~20 internal pane-spawn helpers would have churned the file with no behavioural win — Launcher already owns a `headlessCtx` for the long-running headless dispatch path. Also: added `context.WithTimeout` to `resetBrokerState` and `loadRunningSessionMode` HTTP calls, and the `setup.InstallLatestCLI()` caller. |
| `internal/team/telegram.go` | 6 | Exported helpers (`VerifyBot`, `VerifyChat`, `DiscoverGroups`, `SendTypingAction`, `SendTelegramMessage`, `sendMessageWithMode`) now wrap their Telegram HTTP calls with a 30s `context.WithTimeout`. |
| `internal/team/broker.go` | 6 | `net.Listen` → `(&net.ListenConfig{}).Listen(ctx, …)`; proxy handler upgraded to `http.NewRequestWithContext(r.Context(), …)` so client disconnects propagate; `respawnAgentPane` and pane-snapshot loop swapped to `exec.CommandContext`. |
| `cmd/wuphf/channel.go` | 5 | `openBrowserURL` and `killTeamSession` use `context.Background()` (fire-and-forget OS handoff / process-exit cleanup respectively, with timeout on the latter). |
| `internal/team/worktree.go` | 3 | `runGit`, `runGitOutput`, `gitRefExists` each bound their git invocation with a 60s / 30s timeout — surfaces hung credential helpers instead of pinning the broker. |
| `cmd/wuphf-oc-probe/multi-provider-http/main.go` | 3 | `fetchMessages`, `brokerPOST`, `waitHealth` now use `http.NewRequestWithContext` with a 5–30s ctx. |
| `internal/onboarding/handlers.go` | 2 | `validateProviderKey` / `pingAnthropic` / `pingOpenAI` gained a `context.Context` parameter; `HandleValidateKey` passes `r.Context()` through. |
| `internal/api/client.go` | 2 | `request[T]` and `getRaw` use a per-request `context.WithTimeout(context.Background(), client.Timeout)`. The 64 generic-helper callers (`api.Get/Post/Put/Patch/Delete`, `client.GetRaw`) were left unchanged — threading ctx through all of them would have touched 20+ files; the deadline behaviour is identical because `client.Timeout` already bounded each request. |
| `cmd/wuphf/channel_broker.go` | 2 | `newBrokerRequest` gained a `context.Context` parameter (~30 callers across 5 files updated to pass `context.Background()` from inside their `tea.Cmd` closure); `pollHealth` rewritten to use `http.NewRequestWithContext` + `client.Do`. |
| `internal/team/human_identity.go` | 1 | `runGitConfig` bound with 5s timeout. |
| `internal/team/headless_codex.go` | 1 | `tmux kill-session` swapped to `exec.CommandContext`. |
| `internal/setup/install.go` | 1 | `InstallLatestCLI` gained a `context.Context` parameter. Three callers updated (`commands/cmd_system.go`, `cmd/wuphf/channel.go`, `internal/team/launcher.go`) plus the test. |

## Non-mechanical decisions

- **api package `context.Background()` at the boundary** — 64 call sites use the generic `api.Get/Post/...[T]` helpers; threading ctx through every one would have added an unrelated 20+ file diff. The per-request `context.WithTimeout(client.Timeout)` preserves the prior cancellation semantics exactly. Worth doing later as its own PR if/when callers gain real ctxs.
- **launcher.go bulk `context.Background()`** — same trade-off at larger scale. Threading 63 sites through layers of pane-spawn / pane-respawn helpers was disproportionate to the value (tmux ops are sub-second; the Launcher already cancels long-lived headless work via `headlessCtx`).
- **`telegram.go` exported helpers stay synchronous** — tempting to add `ctx` to their public signatures, but they're called from CLI / Bubble Tea closures that don't currently have a ctx, and changing public signatures would balloon the diff. The internal `WithTimeout(30s)` matches the de-facto behaviour from the previous default `http.Client`.

## Verification

- [x] `golangci-lint run --timeout=5m ./...` → 0 issues
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `gofmt -l .` clean
- [x] `go test -count=1 -timeout 10m ./...` → all packages pass (internal/team: 160s, full suite green)
- [x] Pre-commit hooks pass (gofmt, go-vet, golangci-lint, secretlint, no-secrets)
- [x] Pre-push smoke (build + vet + vhs) passes

## Punted

Nothing. No `//nolint:noctx` suppressions, no path exclusions, no skipped findings. The api package and launcher.go boundary decisions are documented above — they ARE the fix, just at a wider boundary than per-call.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,13 +14,13 @@ linters:
     - govet
     - ineffassign
     - misspell
+    - noctx
     - staticcheck
     - unconvert
     # Followups (own PR each — too many existing findings to land here):
     #   - contextcheck (26 findings: ctx threading)
     #   - gocritic (40 findings: mostly diagnostic-tag style nits)
     #   - nilerr (29 findings: returning nil after non-nil err)
-    #   - noctx (95 findings: net/http and exec.Command without context)
   settings:
     errcheck:
       exclude-functions:
@@ -49,6 +49,7 @@ linters:
           - errcheck
           - bodyclose
           - errorlint
+          - noctx
       # govet's shadow check on `err` is famously noisy in idiomatic Go
       # (every nested `if err := ...; err != nil` shadows). Keep shadow on
       # for non-err identifiers (real bugs hide there) but mute the err case.
--- a/cmd/wuphf-oc-probe/multi-provider-http/main.go
+++ b/cmd/wuphf-oc-probe/multi-provider-http/main.go
@@ -11,6 +11,7 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -241,7 +242,9 @@ type messageRow struct {
 }
 
 func fetchMessages() []messageRow {
-	req, _ := http.NewRequest(http.MethodGet, brokerURL+"/messages", nil)
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	req, _ := http.NewRequestWithContext(ctx, http.MethodGet, brokerURL+"/messages", nil)
 	req.Header.Set("Authorization", "Bearer "+brokerToken)
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
@@ -323,7 +326,9 @@ func mustPost(err error) {
 
 func brokerPOST(path string, body any) (string, error) {
 	data, _ := json.Marshal(body)
-	req, _ := http.NewRequest(http.MethodPost, brokerURL+path, bytes.NewReader(data))
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	req, _ := http.NewRequestWithContext(ctx, http.MethodPost, brokerURL+path, bytes.NewReader(data))
 	req.Header.Set("Content-Type", "application/json")
 	req.Header.Set("Authorization", "Bearer "+brokerToken)
 	client := &http.Client{Timeout: 30 * time.Second}
@@ -342,7 +347,10 @@ func brokerPOST(path string, body any) (string, error) {
 func waitHealth(timeout time.Duration) {
 	deadline := time.Now().Add(timeout)
 	for time.Now().Before(deadline) {
-		resp, err := http.Get(brokerURL + "/health")
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		req, _ := http.NewRequestWithContext(ctx, http.MethodGet, brokerURL+"/health", nil)
+		resp, err := http.DefaultClient.Do(req)
+		cancel()
 		if err == nil && resp.StatusCode == http.StatusOK {
 			resp.Body.Close()
 			return
--- a/cmd/wuphf/channel.go
+++ b/cmd/wuphf/channel.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -4263,7 +4264,7 @@ func postToChannel(text string, replyTo string, channel string) tea.Cmd {
 			"tagged":   extractTagsFromText(text),
 			"reply_to": strings.TrimSpace(replyTo),
 		})
-		req, err := newBrokerRequest(http.MethodPost, "http://127.0.0.1:7890/messages", bytes.NewReader(body))
+		req, err := newBrokerRequest(context.Background(), http.MethodPost, "http://127.0.0.1:7890/messages", bytes.NewReader(body))
 		if err != nil {
 			return channelPostDoneMsg{err: err}
 		}
@@ -5506,7 +5507,7 @@ func createSkill(description, channel string) tea.Cmd {
 			"channel":     channel,
 		}
 		body, _ := json.Marshal(payload)
-		req, err := newBrokerRequest(http.MethodPost, "http://127.0.0.1:7890/skills", bytes.NewReader(body))
+		req, err := newBrokerRequest(context.Background(), http.MethodPost, "http://127.0.0.1:7890/skills", bytes.NewReader(body))
 		if err != nil {
 			return channelSkillsMsg{}
 		}
@@ -5523,7 +5524,7 @@ func createSkill(description, channel string) tea.Cmd {
 
 func invokeSkill(name string) tea.Cmd {
 	return func() tea.Msg {
-		req, err := newBrokerRequest(http.MethodPost, "http://127.0.0.1:7890/skills/"+name+"/invoke", nil)
+		req, err := newBrokerRequest(context.Background(), http.MethodPost, "http://127.0.0.1:7890/skills/"+name+"/invoke", nil)
 		if err != nil {
 			return channelSkillsMsg{}
 		}
@@ -5543,7 +5544,7 @@ func resetDMSession(agent string, channel string) tea.Cmd {
 			"agent":   agent,
 			"channel": channel,
 		})
-		req, err := newBrokerRequest(http.MethodPost, "http://127.0.0.1:7890/reset-dm", bytes.NewReader(body))
+		req, err := newBrokerRequest(context.Background(), http.MethodPost, "http://127.0.0.1:7890/reset-dm", bytes.NewReader(body))
 		if err != nil {
 			return channelResetDMDoneMsg{err: err}
 		}
@@ -5593,7 +5594,7 @@ func switchSessionMode(mode, agent string) tea.Cmd {
 			"mode":  mode,
 			"agent": agent,
 		})
-		req, err := newBrokerRequest(http.MethodPost, "http://127.0.0.1:7890/session-mode", bytes.NewReader(body))
+		req, err := newBrokerRequest(context.Background(), http.MethodPost, "http://127.0.0.1:7890/session-mode", bytes.NewReader(body))
 		if err != nil {
 			return channelResetDoneMsg{err: err}
 		}
@@ -5647,7 +5648,7 @@ func switchFocusMode(enabled bool) tea.Cmd {
 		body, _ := json.Marshal(map[string]any{
 			"focus_mode": enabled,
 		})
-		req, err := newBrokerRequest(http.MethodPost, "http://127.0.0.1:7890/focus-mode", bytes.NewReader(body))
+		req, err := newBrokerRequest(context.Background(), http.MethodPost, "http://127.0.0.1:7890/focus-mode", bytes.NewReader(body))
 		if err != nil {
 			return nil
 		}
@@ -5662,7 +5663,7 @@ func switchFocusMode(enabled bool) tea.Cmd {
 
 func applyTeamSetup() tea.Cmd {
 	return func() tea.Msg {
-		notice, err := setup.InstallLatestCLI()
+		notice, err := setup.InstallLatestCLI(context.Background())
 		if err != nil {
 			return channelInitDoneMsg{err: err}
 		}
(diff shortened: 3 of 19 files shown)


internal/api/client.go
```diff
@@ -56,7 +57,19 @@ func request[T any](c *Client, method, path string, body any, timeout time.Durat
...
+	}
+	// The api.Client is invoked from many call sites that don't currently
+	// thread a context. Use a per-request background context with the
+	// configured timeout as the deadline — equivalent behaviour to the
+	// previous c.HTTPClient.Timeout but satisfies noctx and lets the request
+	// be cancelled when the deadline elapses.
+	ctx, cancel := context.WithTimeout(context.Background(), t)
+	defer cancel()
+
+	req, err := http.NewRequestWithContext(ctx, method, path, reqBody)
```
FranDias, line 75: **[nit]** `c.HTTPClient.Timeout = t` (L84 below) is now redundant with the ctx timeout — both fire at `t`. Worth deleting the assignment (and the duplicate at L124 in `getRaw`) to avoid the implicit shared-state mutation. Note: writing `c.HTTPClient.Timeout` on a shared client (e.g. `agent.AgentService.client`) is a data race when called concurrently. Pre-existing, but removing the assignment kills the race for free.

.golangci.yml
```diff
@@ -49,6 +49,7 @@ linters:
           - errcheck
           - bodyclose
           - errorlint
+          - noctx
```
FranDias, line 52: **[nit]** Confirms `_test.go` exclusion stays intact (preserved from #325). `httptest.NewRequest` is the canonical pattern in tests and noctx would false-flag every one. Good — no path-level suppressions added in this PR.

cmd/wuphf/channel.go
```diff
@@ -5784,10 +5792,18 @@ func isWindows() bool { return runtime.GOOS == "windows" }
 
 // killTeamSession kills the entire wuphf-team tmux session and all agent processes.
 func killTeamSession() {
+	// Best-effort cleanup at process exit; cap each step so a hung tmux or
+	// broker doesn't keep us alive forever.
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
FranDias, line 5797: **[nit]** Single 5s ctx is shared between the tmux kill (line 5800) and the broker `/health` call below. If `tmux kill-session` consumes most of the budget, the broker GET inherits an almost-expired (or fully-cancelled) ctx and fails immediately. Fine for best-effort process-exit cleanup, but a separate ctx per step would be clearer:

```go
func killTeamSession() {
    {
        ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
        _ = exec.CommandContext(ctx, "tmux", "-L", "wuphf", "kill-session", "-t", "wuphf-team").Run()
        cancel()
    }
    {
        ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
        defer cancel()
        // ...
    }
}
```

internal/team/launcher.go
```diff
@@ -2161,9 +2161,9 @@ func (l *Launcher) Attach() error {
 	if os.Getenv("TERM_PROGRAM") == "iTerm.app" {
 		// tmux -CC mode: iTerm2 takes over window management.
 		// Creates native iTerm2 tabs/splits for each tmux window/pane.
-		cmd = exec.Command("tmux", "-L", tmuxSocketName, "-CC", "attach-session", "-t", l.sessionName)
+		cmd = exec.CommandContext(context.Background(), "tmux", "-L", tmuxSocketName, "-CC", "attach-session", "-t", l.sessionName)
```
FranDias, line 2164: **[suggestion]** `tmux attach-session` is the long-running, foreground call here — the user keeps it open. Wrapping it in `exec.CommandContext(context.Background(), ...)` is fine (no cancellation needed), but it's also the one place in this file where a SIGKILL'd context *would* be harmful: if anything ever did cancel this ctx, tmux gets SIGKILL'd mid-detach and the session can be left in a corrupt state on the socket. Worth a one-line comment here saying "context.Background() is intentional — this call must outlive the parent ctx; tmux owns the foreground TTY." Same applies to L2166.

internal/team/broker.go
```diff
@@ -4758,12 +4759,13 @@ func respawnAgentPane(slug string) {
 			paneIdx := i + 1 // pane 0 is channel view
 			target := fmt.Sprintf("wuphf-team:team.%d", paneIdx)
 			// Send Ctrl+C to interrupt, then exit to terminate
-			_ = exec.Command("tmux", "-L", "wuphf", "send-keys", "-t", target, "C-c", "").Run()
+			ctx := context.Background()
```
FranDias, line 4762: **[medium]** This is `respawnAgentPane`, called from a goroutine spawned by the broker (line 4743 above). The `Launcher.headlessCtx` lives next door in launcher.go and is the natural lifecycle owner — `Launcher.Kill()` calls `headlessCancel()` to stop these goroutines. Same pattern shows up in `launcher.go` `watchChannelPaneLoop` (L1265-1273), `sendNotificationToPane` (L3122-3128), `clearAgentPanes` (L3245), `clearOverflowAgentWindows` (L3263), and the `notifyAgentsLoop` chain. All of them run *after* `l.headlessCtx` is initialised in `Launch()` (L434).

The PR doc says these were left as `context.Background()` because tmux ops are short-lived. That's defensible — but the symmetry should be: pre-`Launch()` paths use bg, post-init paths use `l.headlessCtx`. Without that, `Kill()` returns while in-flight tmux subprocesses run to completion against a dying tmux server. Not blocking — bounded leak — but worth a follow-up.

cmd/wuphf-oc-probe/multi-provider-http/main.go
```diff
@@ -342,7 +347,10 @@ func brokerPOST(path string, body any) (string, error) {
 func waitHealth(timeout time.Duration) {
 	deadline := time.Now().Add(timeout)
 	for time.Now().Before(deadline) {
-		resp, err := http.Get(brokerURL + "/health")
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		req, _ := http.NewRequestWithContext(ctx, http.MethodGet, brokerURL+"/health", nil)
+		resp, err := http.DefaultClient.Do(req)
+		cancel()
```
FranDias, line 353: **[nit]** `cancel()` runs *before* `resp.Body.Close()` and the goto/return paths below. After cancel, the connection is forcibly torn down — closing the body afterward can drop the connection from the keepalive pool ungracefully. Canonical order is `defer resp.Body.Close(); defer cancel()` so the body drains first.

Since this is a probe binary, low impact — but the loop runs in a tight retry until success, so each retry leaks a connection. Easy fix:

```go
ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
req, _ := http.NewRequestWithContext(ctx, http.MethodGet, brokerURL+"/health", nil)
resp, err := http.DefaultClient.Do(req)
if err == nil {
    if resp.StatusCode == http.StatusOK {
        resp.Body.Close()
        cancel()
        return
    }
    resp.Body.Close()
}
cancel()
```

internal/team/broker.go
```diff
@@ -2416,7 +2417,7 @@ func (b *Broker) webUIProxyHandler(brokerURL, stripPrefix string) http.Handler {
 			target += "?" + r.URL.RawQuery
 		}
 
-		proxyReq, err := http.NewRequest(r.Method, target, r.Body)
+		proxyReq, err := http.NewRequestWithContext(r.Context(), r.Method, target, r.Body)
```
FranDias, line 2420: **[verified — looks correct]** `r.Context()` flowing to the upstream proxy request is the right call: client disconnect → `r.Context().Done()` fires → upstream `proxyReq` ctx cancels → `client.Do` aborts and the SSE read loop at L2451 breaks on `readErr != nil`. Deferred `resp.Body.Close()` at L2438 still runs cleanly even when ctx fires mid-stream. No regression vs. the prior `http.NewRequest`.

internal/team/telegram.go
```diff
@@ -426,7 +426,14 @@ func (t *TelegramTransport) sendMessageWithMode(chatID, text, parseMode string)
 		return err
 	}
 
-	resp, err := t.client.Post(url, "application/json", bytes.NewReader(data))
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
FranDias, line 429: **[medium]** This is called from `drainOutbound(ctx)` (L169), which has the transport ctx in scope, but the new code creates a fresh `context.Background()` 30s timeout instead of accepting the parent. Net effect: `TelegramTransport.Start` shutdown won't cancel an in-flight Telegram send for up to 30s.

Same pattern in `SendTypingAction` (L469), `VerifyBot` (L511), `DiscoverGroups` (L556), `SendTelegramMessage` (L619), `VerifyChat` (L650). The exported helpers can keep the bg+timeout for now (no caller has ctx today), but `sendMessageWithMode` should accept `ctx context.Context` and the call site at L201 should pass `ctx`. Pre-existing gap, but the noctx fix is the natural moment to close it — per CLAUDE.md "own surfaced bugs".

FranDias (commented):
Mechanical noctx sweep across 96 sites — overall solid, but a few choices warrant a closer look before merge. Headline concerns:

1. **`launcher.go` `context.Background()` for tmux ops post-init.** A handful of the 63 conversions occur in long-running watch loops (`watchChannelPaneLoop`, `notifyAgentsLoop`, `respawnAgentPane`, etc.) that run *after* `l.headlessCtx` is initialised in `Launch()`. These could/should plug into `l.headlessCtx` so `Launcher.Kill()` actually aborts in-flight tmux subprocesses on shutdown. The author flagged this as deliberate; my read is that it's defensible for sub-second ops but a quality issue worth a follow-up.

2. **`internal/api/client.go` write to shared `c.HTTPClient.Timeout` is now redundant** with the per-request ctx timeout, and still races on shared clients (e.g. `agent.AgentService.client`). Pre-existing race, not introduced here, but the boundary fix highlights it.

3. **Telegram transport drops parent ctx.** `drainOutbound(ctx)` accepts ctx but calls `sendMessageWithMode` / `SendTelegramMessage` / `SendTypingAction`, all of which create their own `context.Background()` 30s timeouts — so transport shutdown won't cancel in-flight sends for up to 30s. Missed opportunity to thread ctx, since these are already on a ctx-aware code path.

4. **Body-close discipline is mostly fine** across the conversions — spot-checked `pollHealth`, `webUIProxyHandler`, `pingAnthropic/pingOpenAI`, and the probe binary; all close correctly.

5. **`webUIProxyHandler` ctx propagation looks correct** — `r.Context()` flows to upstream, SSE loop breaks on body read error, deferred close runs.

6. **No `//nolint:noctx` suppressions, no `any`, no path exclusions.** CLAUDE.md compliance ✅.

Nothing blocking, but the launcher.go decision and the telegram ctx drop deserve a response. Inline notes below.

PR #337, opened 2026-04-26
ci: collapse per-package go test matrix into single job

FranDias:
## Summary

CI's go-test fan-out was paying ~14 runner-minutes per run on pure overhead. Replacing the 28-leg matrix with a single \`go-test\` job collapses that to a single runner without affecting wall clock.

### Why the matrix was costing so much

From [run 24958782652](https://github.com/nex-crm/wuphf/actions/runs/24958782652) (most recent green main run before this PR):

| Slice | Compute |
|---|---|
| 28× \`go test <pkg>\` matrix legs | ~17.5 runner-minutes |
| Of that, actual \`go test\` work | ~3 minutes |
| Of that, per-leg overhead (checkout + setup-go + download-artifact) | ~14 minutes |

\`internal/team\` (151s with -race) is the long pole no matter how the suite is sliced — every other matrix leg finishes well within that window. So the matrix gave us per-package red x's at ~14 min/run of overhead with zero wall-clock benefit.

### After

One \`go-test\` job runs \`go test -race ./...\` (with \`internal/teammcp\` split off without -race, same as today). \`go test ./...\` parallelizes packages within one process up to GOMAXPROCS, so wall clock is still bounded by \`internal/team\`.

Expected: ~24 min compute → ~10 min compute, wall clock within ~30s of today.

### Branch protection

Required checks per \`gh api repos/nex-crm/wuphf/branches/main/protection\` are \`commitlint, scan, lint, vet, build, shellcheck\`. None of the matrix legs were required, so this PR doesn't change merge gating. The new \`go-test\` job inherits the same not-required posture; if we want it required, that's a separate branch-protection update.

## Test plan

- [ ] CI on this PR runs the new \`go-test\` job and it passes
- [ ] Wall clock is within ~30s of historical pre-PR baseline (\`internal/team\` long pole)
- [ ] Total compute drops materially (verify via Actions usage)

## Summary by CodeRabbit

* **Chores**
  * Improved CI caching for vulnerability tooling to reduce install times and reliably reuse cached artifacts.
* **Tests**
  * Simplified test workflow into consolidated test jobs with stricter failure behavior for empty suites, longer timeouts, captured test output, grouped failure summaries, and automatic upload of test logs as artifacts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,13 @@ jobs:
 
   lint:
     runs-on: ubuntu-latest
+    # Single source of truth for govulncheck binary location. `go install`
+    # writes to $GOBIN if set (else $GOPATH/bin), so pinning GOBIN here
+    # keeps the cache path, install destination, and PATH lookup below
+    # locked to one literal — eliminates the silent-cache-miss class of
+    # bug if a future setup-go release shifts the default GOPATH.
+    env:
+      GOBIN: ${{ github.workspace }}/.gobin
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # ratchet:actions/checkout@v6
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # ratchet:actions/setup-go@v6
@@ -81,13 +88,13 @@ jobs:
           args: --timeout=5m
       - name: Cache govulncheck binary
         # The version below is pinned, so the compiled binary at
-        # ~/go/bin/govulncheck is reusable across every CI run that
-        # doesn't bump GOVULNCHECK_VERSION. Saves ~10-15s of `go install`
+        # $GOBIN/govulncheck is reusable across every CI run that doesn't
+        # bump the govulncheck version. Saves ~10-15s of `go install`
         # (download + compile) per run on warm cache.
         id: cache-govulncheck
         uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # ratchet:actions/cache@v5.0.5
         with:
-          path: ~/go/bin/govulncheck
+          path: ${{ env.GOBIN }}/govulncheck
           key: govulncheck-${{ runner.os }}-v1.1.4
       - name: Install govulncheck
         if: steps.cache-govulncheck.outputs.cache-hit != 'true'
@@ -98,7 +105,7 @@ jobs:
         # advisory to make noise. Pinned to an explicit version so a new
         # govulncheck release cannot flip pass/fail on the same commit.
         run: |
-          export PATH="$(go env GOPATH)/bin:$PATH"
+          export PATH="${{ env.GOBIN }}:$PATH"
           govulncheck ./...
 
   # Single test job. We previously fanned out one matrix leg per Go package
@@ -124,14 +131,22 @@ jobs:
         with:
           name: web-dist
           path: web/dist
+      # `set -euo pipefail` + the empty-package guard close a fail-open
+      # path: bash does NOT propagate failures inside `$(...)`, so a
+      # transient `go list` error or a regex tweak that drops every match
+      # would otherwise degrade the step to `go test -race -timeout 15m`
+      # against the cwd package only — green CI, every subpackage silently
+      # skipped. `pipefail` surfaces the inner failure; the guard catches
+      # an empty match list.
+      #
+      # On failure: the matrix used to give one red x per failing package
+      # in the PR checks list; collapsed, both invocations swallow per-
+      # package status into a single red x. `tee` to a log file + a
+      # `grep '^FAIL'` summary at the end of the step reclaims the per-
+      # package signal in the step's own log without bringing back the
+      # matrix overhead. Logs are uploaded as an artifact on failure for
+      # offline triage.
       - name: Tests with race detector (everything except teammcp)
-        # `set -euo pipefail` + the empty-package guard close a fail-open
-        # path: bash does NOT propagate failures inside `$(...)`, so a
-        # transient `go list` error or a regex tweak that drops every
-        # match would otherwise degrade the step to `go test -race
-        # -timeout 15m` against the cwd package only — green CI, every
-        # subpackage silently skipped. `pipefail` surfaces the inner
-        # failure; the guard catches an empty match list.
         run: |
           set -euo pipefail
           pkgs=$(go list ./... | grep -vE '/internal/teammcp($|/)')
@@ -139,7 +154,16 @@ jobs:
             echo "no packages matched — refusing to run an empty go test suite" >&2
             exit 1
           fi
-          go test -race -timeout 15m $pkgs
+          set +e
+          go test -race -timeout 15m $pkgs 2>&1 | tee go-test-race.log
+          status=${PIPESTATUS[0]}
+          set -e
+          if [ "$status" -ne 0 ]; then
+            echo "::group::Failing packages (race suite)"
+            grep -E '^(FAIL|--- FAIL)' go-test-race.log || echo "(no FAIL lines parsed; see full log above)"
+            echo "::endgroup::"
+            exit "$status"
+          fi
       - name: Tests without race (internal/teammcp)
         run: |
           set -euo pipefail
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -145,16 +124,10 @@ jobs:
...
-        # wall-clock, so 5m is comfortable headroom for the slowest
-        # per-package legs (internal/team* hit ~1-2m locally without race).
-        run: |
-          if [ "${{ matrix.race }}" = "true" ]; then
-            go test -race -timeout 5m ${{ matrix.package }}
-          else
-            go test -timeout 5m ${{ matrix.package }}
-          fi
+      - name: Tests with race detector (everything except teammcp)
+        run: go test -race -timeout 15m $(go list ./... | grep -vE '/internal/teammcp($|/)')
```
FranDias: **IMPORTANT** — Bash failing-open: `$(go list ./... | grep -vE ...)` can silently degenerate to a no-op CI green.

This is a single-line `run:` so GitHub Actions wraps it in the default `bash --noprofile --norc -e -o pipefail`. Two ways this still bites:

1. `set -e` does NOT abort on command-substitution failures (long-standing bash gotcha — the `$(...)` exit code is not propagated to the outer command).
2. `pipefail` catches a failing `go list` inside the pipe — but only if the *pipeline* runs as the top-level command. Inside `$(...)` the pipeline's exit code populates the substitution's `$?`, which `go test` then ignores because it just sees its argv.

The failure mode: if `go list ./...` ever exits non-zero (network blip on a fresh `go env GOMODCACHE`, a transient module-proxy 5xx, future `GOTOOLCHAIN=local` mismatch, a single broken package's typecheck hanging the list) and prints nothing to stdout, `$()` resolves to empty and the step runs `go test -race -timeout 15m` with no args. That tests the cwd module's root package only (which exists at `embed.go` / `templates_embed_test.go` in the repo root) and exits 0 — green CI, ~all subpackage tests skipped.

**Fix:** convert both steps to a multi-line `run: |` block with `set -euo pipefail` and pre-compute the list with explicit non-empty validation:

```yaml
- name: Tests with race detector (everything except teammcp)
  run: |
    set -euo pipefail
    pkgs=$(go list ./... | grep -vE '/internal/teammcp($|/)')
    [ -n "$pkgs" ] || { echo 'go list returned no packages' >&2; exit 1; }
    go test -race -timeout 15m $pkgs
```

Same shape for the teammcp step (where empty would also be a regression — currently means "teammcp got renamed/moved without updating the workflow").

.github/workflows/ci.yml
```diff
@@ -145,16 +124,10 @@ jobs:
...
-        # `-timeout 5m` is a per-package margin. Race instrumentation 2-3x's
-        # wall-clock, so 5m is comfortable headroom for the slowest
-        # per-package legs (internal/team* hit ~1-2m locally without race).
-        run: |
-          if [ "${{ matrix.race }}" = "true" ]; then
-            go test -race -timeout 5m ${{ matrix.package }}
-          else
-            go test -timeout 5m ${{ matrix.package }}
-          fi
+      - name: Tests with race detector (everything except teammcp)
```
FranDias, line 149: **SUGGESTION** — Failure-attribution DX regression worth acknowledging in the comment block above.

The matrix gave one red x per failing package directly in the PR checks list (`go test internal/team` → red), so the failing package was identifiable from the GitHub UI without opening logs. Post-collapse, both `go test ./...` invocations swallow per-package status into a single red x, and a developer has to scroll the step log to find the `--- FAIL` / `FAIL\tgithub.com/.../foo` line. With ~25 packages and `-race` instrumentation noise, that's a real degradation for the on-call engineer triaging a flake on a Sunday.

Not a blocker — wall-clock and compute trade still favour the collapse — but two cheap mitigations worth considering:

1. Add `-v=false -json | tparse -all -follow` (or just `... | tee test.log` and a final `grep -E '^(FAIL|---)' test.log >&2 || true`) so a failure surfaces the failing package list at the bottom of the step log.
2. On failure, upload `test.log` as an artifact (mirrors the pattern web-e2e uses for Playwright + wuphf logs at lines 376-389).

Either is a 3-line addition and reclaims the per-package signal without bringing back the matrix.

.github/workflows/ci.yml
```diff
@@ -79,63 +79,42 @@ jobs:
...
           args: --timeout=5m
+      - name: Cache govulncheck binary
+        # The version below is pinned, so the compiled binary at
+        # ~/go/bin/govulncheck is reusable across every CI run that
+        # doesn't bump GOVULNCHECK_VERSION. Saves ~10-15s of `go install`
+        # (download + compile) per run on warm cache.
+        id: cache-govulncheck
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # ratchet:actions/cache@v5.0.5
+        with:
+          path: ~/go/bin/govulncheck
```
FranDias: **SUGGESTION** — Cache path / install path drift risk.

Cache stores `~/go/bin/govulncheck` (literal expansion of `$HOME/go/bin/govulncheck`). The `Install govulncheck` step on cache miss uses `go install`, which writes to `$GOBIN` if set, else `$GOPATH/bin`. The `Run govulncheck` step then resolves PATH via `$(go env GOPATH)/bin`.

Today on `ubuntu-latest` with default `setup-go`, all three resolve to `/home/runner/go/bin/` so it lines up. But the three references are independent — if a future setup-go release changes the default GOPATH, or someone adds a `GOBIN: /usr/local/bin` env var to this job, the cache will silently miss every run (binary saved at the wrong path → 0 cache hits → no observable failure, just wasted time).

**Fix (low cost):** anchor all three to one source of truth. Either:

```yaml
env:
  GOBIN: /home/runner/go/bin
```

at the job level and use `${{ env.GOBIN }}/govulncheck` for both the cache `path` and PATH munging. Or compute it once in a step and stash in `$GITHUB_ENV`.

Low severity — current behaviour is correct, just brittle in a way that fails silently rather than loudly.

.github/workflows/ci.yml
```diff
@@ -79,63 +79,42 @@ jobs:
...
+          key: govulncheck-${{ runner.os }}-v1.1.4
```
FranDias, line 98: **SUGGESTION** — Cache key precision.

`runner.os` resolves to `Linux` regardless of whether the underlying image is `ubuntu-22.04`, `ubuntu-24.04`, or whatever GitHub bumps `ubuntu-latest` to next. govulncheck is a pure-Go binary so glibc/CGO drift isn't an issue, but if the binary was compiled against a Go version present in image N and you later bump go.mod past what image N+1 ships, the cached binary keeps working anyway (govulncheck is statically linked). So the practical risk is near-zero.

That said, this matches the existing pattern (vhs, ttyd, playwright caches all key on `runner.os` only — see lines 183, 192, 275), so flagging for awareness rather than asking for a change. If consistency with the rest of the file is the goal, ship as-is.

FranDias:
**Verdict: approve-with-nits.** Compute reduction is real (~14 runner-min/CI run for unchanged wall clock), the govulncheck cache fills a genuine gap, and the regex anchoring (`/internal/teammcp($|/)`) correctly keeps `internal/team` in the race suite while excluding only the teammcp subtree.

Four inline comments posted, one of which is worth addressing before merge:

1. **(important)** `$(go list ./... | grep ...)` can fail-open to a no-op `go test` (see line 128 comment) — `set -e` doesn't propagate command-substitution failures, and the cwd has a Go package at the root, so an empty substitution would silently green CI. Three-line `set -euo pipefail` + non-empty guard fixes it.
2. **(suggestion)** Failure-attribution UX: the matrix surfaced per-package red x's in the checks list; the collapsed job hides which package failed until you open the log. Consider a `tee + grep '^FAIL'` and an artifact upload on failure (line 127).
3. **(suggestion)** govulncheck cache path / install path / PATH-resolution path are three independent references to the same logical location; a job-level `GOBIN` env var would prevent silent cache misses if setup-go's GOPATH default ever shifts (line 90).
4. **(suggestion, no-action)** `runner.os`-only cache key matches the existing pattern for vhs/ttyd/playwright caches — flagging for awareness only (line 91).

The race-detector memory pressure concern (~25 packages × 5-10x race overhead on a 16GB runner) is real in theory but the green run at 3:11 is empirical evidence it's fine; revisit only if OOMs surface.

Non-blocking; LGTM with the bash hardening on line 128.

FranDias:
@coderabbitai try again

.github/workflows/ci.yml
FranDias: Fixed in 56f6cf2c. Added job-level `env: GOBIN: ${{ github.workspace }}/.gobin` and reused it in the cache `path` and the PATH lookup. All three references now point at the same single literal — no more drift risk if setup-go changes its default GOPATH.

.github/workflows/ci.yml
FranDias: Fixed in 56f6cf2c. Both test steps now `tee` to a log file, capture `${PIPESTATUS[0]}`, and on failure surface `^(FAIL|--- FAIL)` matches inside a collapsible `::group::` at the end of the step — failing package is visible without scrolling. Added an `if: failure()` `upload-artifact` step (`go-test-logs`, 7-day retention) so the full logs are downloadable for offline triage, mirroring the web-e2e pattern.

.github/workflows/ci.yml
FranDias: Skipping per your own "ship as-is" note — this matches the existing pattern in vhs/ttyd/playwright caches, and the practical risk for a statically linked pure-Go binary is near-zero. Will revisit if we ever pin to a specific ubuntu image.

.github/workflows/ci.yml
FranDias: Already addressed in commit e70d1635 (pre-merge of main). Both steps now run as multi-line scripts with `set -euo pipefail`, capture the package list to a variable, and exit 1 with an error message if the list is empty. The 56f6cf2c follow-up adds the failure-attribution improvements on top of the same scaffold.

PR #343, opened 2026-04-26
feat(settings): make wuphf shred in restart banner a clickable inline button

FranDias:
## Summary

The General-section banner used to tell users to "Run `wuphf shred` then relaunch" but the shred command was a static `<code>` chip — they had to copy it into a terminal. Now the chip is a real inline button: click → existing type-the-phrase confirm modal → runs the wipe directly.

Also lifts two reusable ui primitives so future "user has to run X" banners can wire the same pattern:

- `WipeModal` — extracted out of `SettingsApp` so any feature can gate a destructive action behind the existing confirm-phrase friction. Danger Zone now imports the shared one instead of carrying its own copy.
- `InlineCommand` — a clickable inline `<code>`-styled chip with a small ▶ glyph. Pass `destructive={...}` to require a `WipeModal` confirm; otherwise click runs `onRun` immediately.

## Before / After

**Before** — static `wuphf shred` text chip, no affordance:

Before: static wuphf shred text chip

**After** — clickable inline button with play glyph + hover ring:

After: clickable wuphf shred inline button

**Click → existing WipeModal** (same type-the-phrase confirmation that Danger Zone already uses):

Click opens WipeModal with type-the-phrase gate

## Test plan

- [x] `npm run test` — 412/412 passed (incl. the existing SettingsApp coverage)
- [x] `npm run build` — production bundle builds cleanly
- [x] Visual verification via Playwright against a fresh wuphf instance — idle / hover / modal-open states all render correctly (see screenshots above)
- [ ] Manual smoke: hit Settings → General, click `wuphf shred`, type the phrase, confirm wipe runs and broker exits

> Screenshots live on a separate `assets/inline-command-screenshots` orphan branch so this branch's diff stays code-only.

## Summary by CodeRabbit

* **New Features**
  * Inline destructive "shred" control with confirmation, busy-state handling, and success/error notices; guidance to relaunch after completion.
  * Reusable destructive confirmation dialog that requires typing an exact phrase to confirm.
  * Shared shred action that surfaces success/error toasts, resets onboarding/navigation on success, and keeps modals open on failure.

* **Refactor**
  * Unified shred/reset confirmation UI and standardized warning, deletions, and preservation lists for consistent copy and layout.

### DIFF
--- a/web/src/components/apps/SettingsApp.tsx
+++ b/web/src/components/apps/SettingsApp.tsx
@@ -450,9 +450,9 @@ function GeneralSection({ cfg, save }: SectionProps) {
                 <>
                   This permanently deletes your team, company identity, office
                   task receipts, and saved workflows, plus local logs, sessions,
-                  provider state, calendar, and wiki memory. WUPHF will stop
-                  after the wipe; relaunch it to reopen onboarding. Task
-                  worktrees and config are kept.{" "}
+                  provider state (including codex-headless scratch), calendar,
+                  and wiki memory. WUPHF will stop after the wipe; relaunch it
+                  to reopen onboarding. Task worktrees and config are kept.{" "}
                   <strong>This cannot be undone.</strong>
                 </>
               ),
@@ -1641,7 +1641,8 @@ function DangerZoneSection() {
             <code>~/.wuphf/</code>
           </li>
           <li>
-            Logs, sessions, provider state, calendar, and local wiki memory
+            Logs, sessions, provider state (incl. <code>codex-headless</code>{" "}
+            scratch), calendar, and local wiki memory
           </li>
           <li>Broker runtime state (same as Reset)</li>
         </ul>
@@ -1693,8 +1694,9 @@ function DangerZoneSection() {
             <>
               This permanently deletes your team, company identity, office task
               receipts, and saved workflows, plus local logs, sessions, provider
-              state, calendar, and wiki memory. Onboarding will reopen
-              immediately. Task worktrees, config, and device identity are kept.{" "}
+              state (including codex-headless scratch), calendar, and wiki
+              memory. Onboarding will reopen immediately. Task worktrees,
+              config, and device identity are kept.{" "}
               <strong>This cannot be undone.</strong>
             </>
           }
--- a/web/src/components/apps/SettingsApp.tsx
+++ b/web/src/components/apps/SettingsApp.tsx
@@ -30,6 +30,11 @@ import {
 } from "../../api/client";
 import { useAppStore } from "../../stores/app";
 import { InlineCommand } from "../ui/InlineCommand";
+import {
+  ShredDeletionsList,
+  ShredPreservationList,
+  ShredWarningCopy,
+} from "../ui/ShredWarning";
 import { showNotice } from "../ui/Toast";
 import { WipeModal } from "../ui/WipeModal";
 
@@ -447,14 +452,7 @@ function GeneralSection({ cfg, save }: SectionProps) {
               severity: "critical",
               confirmLabel: "Shred workspace",
               intro: (
-                <>
-                  This permanently deletes your team, company identity, office
-                  task receipts, and saved workflows, plus local logs, sessions,
-                  provider state (including codex-headless scratch), calendar,
-                  and wiki memory. WUPHF will stop after the wipe; relaunch it
-                  to reopen onboarding. Task worktrees and config are kept.{" "}
-                  <strong>This cannot be undone.</strong>
-                </>
+                <ShredWarningCopy aftermath="WUPHF will stop after the wipe; relaunch it to reopen onboarding." />
               ),
             }}
           />{" "}
@@ -1629,33 +1627,11 @@ function DangerZoneSection() {
         </div>
         <div style={dangerStyles.listLabel}>Deletes</div>
         <ul style={dangerStyles.list}>
-          <li>
-            Onboarding flag (<code>~/.wuphf/onboarded.json</code>) so the wizard
-            reopens
-          </li>
-          <li>
-            Company identity (<code>~/.wuphf/company.json</code>)
-          </li>
-          <li>
-            Team runtime state, office, and workflows under{" "}
-            <code>~/.wuphf/</code>
-          </li>
-          <li>
-            Logs, sessions, provider state (incl. <code>codex-headless</code>{" "}
-            scratch), calendar, and local wiki memory
-          </li>
-          <li>Broker runtime state (same as Reset)</li>
+          <ShredDeletionsList />
         </ul>
         <div style={dangerStyles.listLabel}>Preserved</div>
         <ul style={dangerStyles.list}>
-          <li>
-            <strong>Task worktrees</strong> — uncommitted work on branches stays
-            on disk
-          </li>
-          <li>
-            Your global config (<code>config.json</code>) and API keys
-          </li>
-          <li>OpenClaw device identity used for gateway pairing</li>
+          <ShredPreservationList />
         </ul>
         <button
           type="button"
@@ -1691,14 +1667,7 @@ function DangerZoneSection() {
           title="Shred this workspace?"
           severity="critical"
           intro={
-            <>
-              This permanently deletes your team, company identity, office task
-              receipts, and saved workflows, plus local logs, sessions, provider
-              state (including codex-headless scratch), calendar, and wiki
-              memory. Onboarding will reopen immediately. Task worktrees,
-              config, and device identity are kept.{" "}
-              <strong>This cannot be undone.</strong>
-            </>
+            <ShredWarningCopy aftermath="Onboarding will reopen immediately." />
           }
           confirmLabel="Shred workspace"
           busy={busy}
--- a/web/src/components/apps/SettingsApp.tsx
+++ b/web/src/components/apps/SettingsApp.tsx
@@ -29,7 +29,9 @@ import {
   type WorkspaceWipeResult,
 } from "../../api/client";
 import { useAppStore } from "../../stores/app";
+import { InlineCommand } from "../ui/InlineCommand";
 import { showNotice } from "../ui/Toast";
+import { WipeModal } from "../ui/WipeModal";
 
 type SectionId =
   | "general"
@@ -387,6 +389,24 @@ function GeneralSection({ cfg, save }: SectionProps) {
   const [blueprint, setBlueprint] = useState(cfg.blueprint ?? "");
   const [email, setEmail] = useState(cfg.email ?? "");
   const [devUrl, setDevUrl] = useState(cfg.dev_url ?? "");
+  const queryClient = useQueryClient();
+
+  const runShred = async () => {
+    try {
+      const result: WorkspaceWipeResult = await shredWorkspace();
+      if (!result.ok) {
+        showNotice(result.error || "Shred failed", "error");
+        return;
+      }
+      queryClient.clear();
+      showNotice(
+        "Workspace shredded. Relaunch wuphf to start onboarding.",
+        "success",
+      );
+    } catch (err) {
+      showNotice(err instanceof Error ? err.message : "Shred failed", "error");
+    }
+  };
 
   const onSave = async () => {
     const patch: ConfigUpdate = {
@@ -419,17 +439,25 @@ function GeneralSection({ cfg, save }: SectionProps) {
           </strong>
           New values save immediately, but agents already running keep their
           launch-time settings. Run{" "}
-          <code
-            style={{
-              fontFamily: "var(--font-mono)",
-              padding: "1px 6px",
-              background: "var(--warning-200)",
-              color: "var(--warning-500)",
-              borderRadius: 3,
+          <InlineCommand
+            command="wuphf shred"
+            onRun={runShred}
+            destructive={{
+              title: "Shred this workspace?",
+              severity: "critical",
+              confirmLabel: "Shred workspace",
+              intro: (
+                <>
+                  This permanently deletes your team, company identity, office
+                  task receipts, and saved workflows, plus local logs, sessions,
+                  provider state, calendar, and wiki memory. WUPHF will stop
+                  after the wipe; relaunch it to reopen onboarding. Task
+                  worktrees and config are kept.{" "}
+                  <strong>This cannot be undone.</strong>
+                </>
+              ),
             }}
-          >
-            wuphf shred
-          </code>{" "}
+          />{" "}
           then relaunch to apply.
         </div>
       </div>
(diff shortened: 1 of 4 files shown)


FranDias (commented):
Solid refactor — lifting `WipeModal` and centralizing the shred copy is the right call. But the new `InlineCommand` chip has two real a11y gaps (no focus indicator, no keyboard equivalent for the backdrop dismiss it inherits via `WipeModal`), the dropped `autoFocus` is a non-trivial UX regression with a trivial fix, and `run()` swallows rejections in a way that will leave the modal stuck on `Working…` if any future caller doesn't internally catch.

Also: the lifted `WipeModal` is a good moment to add `role="dialog"` / `aria-modal` / Escape-to-close — this is a reusable primitive now, not an ad-hoc inline modal, and the bar should be higher.

No tests for any of the three new components — at least a smoke test for `InlineCommand` (destructive path opens modal, non-destructive runs immediately, busy gates re-entry) would catch most regressions.

Requesting changes on the focus-indicator, error-swallow, and `autoFocus` items; the rest are suggestions/nits.

web/src/components/ui/InlineCommand.tsx
```diff
@@ -0,0 +1,129 @@
...
+          gap: 5,
+          fontFamily: "var(--font-mono)",
+          fontSize: "inherit",
+          padding: "1px 8px",
+          background: hover && !busy ? palette.bgHover : palette.bg,
+          color: palette.fg,
+          border: "none",
+          borderRadius: 4,
+          cursor: busy ? "progress" : "pointer",
+          boxShadow: hover && !busy ? `0 0 0 2px ${palette.ring}` : "none",
```
FranDias: **[IMPORTANT] No keyboard focus indicator.**

Hover styling is wired via `onMouseEnter`/`onMouseLeave` (lines 83–84) and applied via `palette.bgHover` + `boxShadow` rings on lines 95 and 100, but there's no `:focus`/`:focus-visible` equivalent. A keyboard user tabbing onto the chip gets zero visual indication it's focused — no ring, no background change, nothing. Browser default outlines on `<button>` are also overridden by `border: none` (line 97).

Since this chip lives in the middle of a paragraph and is the primary CTA on the banner, missing focus styling is a real a11y issue and a regression vs. the inert `<code>` it replaces (which couldn't be focused at all, but also didn't pretend to be interactive).

**Fix:** track focus alongside hover and apply the same `bgHover` + ring on focus, e.g.
```tsx
const [focus, setFocus] = useState(false);
const active = (hover || focus) && !busy;
// ...
onFocus={() => setFocus(true)}
onBlur={() => setFocus(false)}
// ...
background: active ? palette.bgHover : palette.bg,
boxShadow: active ? `0 0 0 2px ${palette.ring}` : 'none',
```
Or drop the JS hover entirely and use a tiny CSS class with `:hover, :focus-visible` — the inline-style approach is what's forcing the JS workaround and still missing focus.

web/src/components/ui/InlineCommand.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const run = async () => {
+    if (busy) return;
+    setBusy(true);
+    try {
+      await onRun();
+      setOpen(false);
+    } finally {
+      setBusy(false);
+    }
+  };
```
FranDias, line 80: **[IMPORTANT] `run()` swallows rejections and leaves the modal stuck.**

If `onRun()` rejects, `setOpen(false)` on line 63 is skipped (it's after the `await`), the rejection escapes the try/finally, and the caller-site is `() => void run()` (line 121) which silently drops it. The modal stays open with the confirm button still rendering `Working…` (well — `busy` flips back to `false` in the `finally`, so actually it goes back to the enabled state, but the modal is still open and the user has no idea what happened). It also produces an unhandled-promise-rejection in the console.

This is currently masked because `GeneralSection.runShred` catches everything internally, but the public type is `onRun: () => void | Promise<void>` — any future caller that doesn't internally catch will trip this. The current `DestructiveConfirm` API doesn't even have a slot for surfacing failure to the parent, so this is a quiet correctness bomb.

**Fix:** catch and either keep modal open with a status, or close + let the caller's toast take over:
```ts
try {
  await onRun();
  setOpen(false);
} catch {
  // caller is expected to surface the error (toast, etc.). Keep modal
  // closed so we don't leave the user staring at a stale confirm dialog.
  setOpen(false);
} finally {
  setBusy(false);
}
```
Or — better — give `DestructiveConfirm` a `keepOpenOnError?: boolean` and let the caller decide.

web/src/components/ui/InlineCommand.tsx
```diff
@@ -0,0 +1,129 @@
...
+          boxShadow: hover && !busy ? `0 0 0 2px ${palette.ring}` : "none",
+          transition: "background 0.12s, box-shadow 0.12s",
+          verticalAlign: "baseline",
+          lineHeight: 1.45,
+          ...style,
+        }}
+      >
+        <PlaySolid
+          width={9}
+          height={9}
```
FranDias: **[SUGGESTION] 9 px is below the icon-legibility floor.**

A 9×9 raster icon at default 16 px root font is small enough that the play triangle effectively becomes a single dark pixel cluster. On a 1× display + 100 % zoom, it's hard to recognize as a play glyph at all. The chip already conveys "this is a command" via the monospace font + colored chip — the icon is the affordance for "this is also a button you can run."

**Fix:** bump to 10 or 11 px. The chip's line-height is 1.45 × inherited font-size, so an 11 px icon still fits inside a ~14 px line box without overflow.

web/src/components/ui/InlineCommand.tsx
```diff
@@ -0,0 +1,129 @@
...
+  };
+
+  const handleClick = () => {
+    if (busy) return;
+    if (destructive) {
+      setOpen(true);
+      return;
+    }
+    void run();
+  };
```
FranDias, line 89: **[SUGGESTION] Reusability gap for the planned non-destructive variant.**

Wiring this for a future "soft restart" agent-relaunch button (your stated follow-up) will hit a few rough edges:

1. **No busy UI for the non-destructive path.** When `destructive` is undefined, click → `run()` → `disabled={busy}` for ~hundreds of ms with `cursor: progress` and no other affordance. The destructive path gets the modal as a busy surface (`Working…` confirm button); the non-destructive path gets nothing. For a soft restart that takes a couple seconds, users will think the click didn't register.
2. **Two-tone palette.** `warning` and `neutral` only. A relaunch button probably wants `info`/`primary` (blue) — adding a third entry to `PALETTES` is fine, but the API will keep growing one tone at a time.
3. **`setOpen(false)` on success runs even when no modal was opened** (line 63). Harmless but wrong-shaped — implies state leakage between paths.

**Fix (incremental):** add a `busyLabel?: string` for the non-destructive path so the button can render `<Spinner /> {busyLabel}` while running, and make `run()` skip `setOpen(false)` when `!destructive`. Probably worth doing before the soft-restart PR rather than as part of it.

web/src/components/ui/WipeModal.tsx
```diff
@@ -0,0 +1,155 @@
...
+        <label style={styles.inputLabel}>
+          Type <code>{WIPE_CONFIRM_PHRASE}</code> to confirm
+        </label>
+        <input
+          type="text"
+          style={styles.input}
+          placeholder={WIPE_CONFIRM_PHRASE}
+          value={value}
+          onChange={(e) => setValue(e.target.value)}
+          disabled={busy}
```
FranDias, line 181: **[SUGGESTION] Dropping `autoFocus` is a real UX regression — biome's `noAutofocus` allows the imperative replacement.**

The PR description acknowledges this; flagging because the fix is two lines and the regression hits every shred/reset interaction. Pre-PR: open modal → start typing immediately. Post-PR: open modal → click into the input → start typing.

Biome's `noAutofocus` rule is specifically about the JSX `autoFocus` attribute — imperative `ref.current?.focus()` in a `useEffect` is the documented escape hatch and is what the rule's docs recommend for modal initial-focus. There's a real a11y rationale for the rule (AT users get teleported on page load), but a modal that has just been triggered by an explicit user action is exactly the case where moving focus *into* the modal is correct behavior.

**Fix:**
```tsx
const inputRef = useRef<HTMLInputElement>(null);
useEffect(() => { inputRef.current?.focus(); }, []);
// ...
<input ref={inputRef} ... />
```

web/src/components/ui/WipeModal.tsx
```diff
@@ -0,0 +1,155 @@
...
+  confirmLabel,
+  busy,
+  onConfirm,
+  onCancel,
+}: WipeModalProps) {
+  const [value, setValue] = useState("");
+  const enabled = !busy && value.trim().toLowerCase() === WIPE_CONFIRM_PHRASE;
+
+  return (
+    <div style={styles.backdrop} onClick={busy ? undefined : onCancel}>
```
FranDias: **[IMPORTANT] Modal is missing dialog semantics + Escape-to-close.**

Now that this is a reusable primitive (not an ad-hoc inline modal), the a11y bar should be higher. Issues:

1. **No `role="dialog"` / `aria-modal="true"`** on the panel (line 120). Screen readers don't announce that a modal opened.
2. **No `aria-labelledby`** linking to the title (line 121). SR users get no announcement of *what* dialog this is.
3. **No Escape handler.** Backdrop click cancels (line 119) but the keyboard equivalent (Escape) is missing — keyboard-only users can't dismiss without tabbing to the Cancel button.
4. **Backdrop is a `<div onClick>` with no role.** Biome's `useKeyWithClickEvents` should be flagging this once you re-run lint on the lifted file.
5. **No focus trap and no focus restoration.** Tab leaks back to the triggering chip / underlying page; on close, focus stays in the dead modal subtree.

**Fix (minimum viable):**
```tsx
useEffect(() => {
  const onKey = (e: KeyboardEvent) => {
    if (e.key === 'Escape' && !busy) onCancel();
  };
  window.addEventListener('keydown', onKey);
  return () => window.removeEventListener('keydown', onKey);
}, [busy, onCancel]);

// on the panel:
<div
  style={styles.panel}
  onClick={(e) => e.stopPropagation()}
  role="dialog"
  aria-modal="true"
  aria-labelledby="wipe-modal-title"
>
  <div id="wipe-modal-title" style={styles.title}>{title}</div>
```
Focus trap can be punted (most app modals here lack one) but the dialog role + Escape are table-stakes for a shared `WipeModal`.

web/src/components/ui/ShredWarning.tsx
```diff
@@ -0,0 +1,74 @@
+import type { ReactNode } from "react";
+
+// Single source of truth for "what does shred actually do?". Any UI that
+// warns about shred — banner buttons, danger-zone cards, confirm modals —
+// pulls copy from here so the description never drifts from
+// `internal/workspace/workspace.go:Shred`. When that function changes,
+// update this file (and only this file).
+
+const DELETIONS_PROSE =
+  "your team, company identity, office task receipts, and saved workflows, plus local logs, sessions, provider state (including codex-headless scratch), calendar, and wiki memory";
```
FranDias: **[SUGGESTION] "office task receipts" narrows what shred actually deletes.**

`internal/workspace/workspace.go:66` removes the entire `~/.wuphf/office` directory, not just task receipts. The `ShredDeletionsList` bullet on line 47 correctly says just "office"; it's only the prose here that narrows it. A user reading the modal might think other office state survives — it doesn't.

The other elision (no mention of `wiki.bak` from line 73 of `workspace.go`) is fine since it's a backup of the wiki and "wiki memory" reads as covering both.

**Fix:** s/office task receipts/office state/ here so the prose matches the bullet list.

web/src/components/apps/SettingsApp.tsx
```diff
@@ -383,6 +390,24 @@ function GeneralSection({ cfg, save }: SectionProps) {
...
+      }
+      queryClient.clear();
+      showNotice(
+        "Workspace shredded. Relaunch wuphf to start onboarding.",
+        "success",
+      );
+    } catch (err) {
+      showNotice(err instanceof Error ? err.message : "Shred failed", "error");
+    }
+  };
```
FranDias: **[SUGGESTION] `runShred` intentionally diverges from `DangerZoneSection.handleShred` — leave a comment.**

The two handlers look like duplicates but are intentionally different:
- `handleShred` (line 1094) calls `resetForOnboarding()` + `window.history.replaceState(...)` and shows "Onboarding reopened".
- `runShred` (here) skips both and tells the user to relaunch.

The rationale (the broker's `AfterShred` hook in `internal/team/broker.go` calls `requestShutdown()`, so post-shred the connection drops and the user has to relaunch the binary regardless) isn't obvious from either handler. A future maintainer reading just `SettingsApp.tsx` will see two near-identical handlers and "fix" the divergence — either by extracting a shared hook (and inadvertently re-introducing `resetForOnboarding()` here) or by deleting one of the unique branches.

**Fix:** one-line comment above `runShred` explaining why it doesn't `resetForOnboarding()` (broker shuts down after shred via `AfterShred` callback; user must relaunch). Same applies to the matching `handleShred` — note that the in-place onboarding it sets up is *also* short-lived because of the same shutdown.

web/src/components/apps/SettingsApp.tsx
```diff
@@ -415,17 +440,18 @@ function GeneralSection({ cfg, save }: SectionProps) {
...
-          <code
-            style={{
-              fontFamily: "var(--font-mono)",
-              padding: "1px 6px",
-              background: "var(--warning-200)",
-              color: "var(--warning-500)",
-              borderRadius: 3,
+          <InlineCommand
+            command="wuphf shred"
+            onRun={runShred}
```
FranDias: **[SUGGESTION] Concurrent shred via the two entry points has no coordination.**

The user can:
1. Open Danger Zone → click Shred → modal opens with `busy=false`
2. Switch sections to General (which unmounts `DangerZoneSection`, killing its modal but leaving its in-flight promise stale)
3. Click the inline `wuphf shred` chip → fires a second `shredWorkspace()` POST

Server-side `Shred()` is `os.RemoveAll` over a fixed list of paths, so two concurrent shreds aren't catastrophic — `RemoveAll` is best-effort and re-entrant. But the second call will likely hit `errors` on already-removed paths, the toasts will collide, and the `AfterShred → requestShutdown` hook fires twice (idempotent via `sync.Once`, fine).

Not worth a global lock, but a cheap fix is a tiny zustand flag (`workspaceWipeInFlight`) that both sites consult before firing. Optional — flagging because you asked.

FranDias:
Addressed in d5d647c9. Map to inline comments:

**InlineCommand.tsx**
- *No keyboard focus indicator (IMPORTANT)* — added `onFocus`/`onBlur` state; `ringActive` now fires on hover OR focus, plus `aria-busy` for SR users.
- *`run()` swallows rejections (IMPORTANT)* — wrapped `await onRun()` in try/catch; rejection now logs and closes the modal in `finally` instead of leaking via `() => void run()`.
- *9px icon below legibility floor (SUGGESTION)* — bumped to 11px.
- *Reusability gap for non-destructive busy UI (SUGGESTION)* — non-destructive path now dims to 0.65 opacity while busy + sets `aria-busy`. Destructive path still gets the WipeModal as a richer busy surface.

**WipeModal.tsx**
- *Missing dialog semantics + Escape (IMPORTANT)* — added `role="dialog"`, `aria-modal="true"`, `aria-labelledby`, `aria-describedby` (linked via `useId()`), plus a `useEffect` keydown listener that calls `onCancel` on Escape (gated on `!busy`).
- *autoFocus regression (SUGGESTION)* — restored via `useRef`+`useEffect`; biome's `noAutofocus` only forbids the JSX attribute.
- *Backdrop drag-out dismiss (MINOR)* — switched to `onMouseDown`/`onMouseUp` with a `backdropPressRef` so dismiss only fires when both press and release happened on the backdrop.

**ShredWarning.tsx**
- *"office task receipts" narrows what shred deletes (SUGGESTION)* — broadened to "office state" with a comment pointing back to `internal/workspace/workspace.go:Shred`.

**SettingsApp.tsx**
- *runShred leaves user on broken Settings page (MAJOR)* + *runShred vs handleShred divergence (SUGGESTION)* — both addressed in one move: extracted `useShredAction()` so GeneralSection's inline button and DangerZoneSection's full card share post-success cleanup (`queryClient.clear` + `history.replaceState` + `resetForOnboarding` + "Onboarding reopened" toast). The two paths are now identical.

**Not addressed**
- *Concurrent shred coordination (SUGGESTION)* — left as-is. Server-side `os.RemoveAll` is idempotent so two concurrent shreds aren't catastrophic, and adding a module-level lock for an interaction the user has to deliberately set up (open Danger Zone modal, switch sections, click inline chip) felt like pre-emptive complexity. Happy to add a guard if you'd rather.

Visual verification via Playwright confirmed the modal a11y attributes (role=dialog, aria-modal=true, labelledby/describedby linked), input autofocus on open, Escape closes the modal, and the new focus ring is visible on Tab. Tests still 412/412 passing, typecheck + biome clean.

FranDias:
Second-pass fixes in a26258db:

- **Aftermath copy diverges (Minor)** + **Docstring rationale wrong (Minor)** — fixed together. Dropped the `aftermath` prop from `ShredWarningCopy` entirely; baked `"Onboarding will reopen immediately."` into `AFTERMATH_PROSE` as the single canonical sentence. Both call sites now render `<ShredWarningCopy />` with no props. The earlier docstring rested on a stale reading of `broker.go` where `AfterShred` used to call `requestShutdown` — origin/main now wires `RouteOptions{ResetRuntime: b.Reset}` so the broker stays up in both paths and the prose convergence matches behavior.
- **cardSubtitle still narrows to "office task receipts" (Minor)** — added `ShredCardSubtitle` to `ShredWarning.tsx` and swapped the hand-written subtitle for `<ShredCardSubtitle />`. The card is no longer a fourth source of truth for the deletion list.
- **handleShred closes the modal on failure (Minor)** — applied the suggested diff: `const ok = await shred(); if (ok) setOpen(null);` so a failed wipe leaves the modal mounted for retry.

Verified end-to-end via Playwright: both modals (inline + Danger Zone) now render byte-identical body text, and the Danger Zone card subtitle uses the canonical `DELETIONS_PROSE` (no "office task receipts" left anywhere). Typecheck + biome clean, tests still 412/412.

PR #344, opened 2026-04-26
feat(upgrade): in-app upgrade banner + wuphf upgrade subcommand

FranDias:
## Summary

Surfaces an "upgrade available" prompt when the running wuphf is older than the latest published on npm. Two consumer surfaces, both fed by a shared `internal/upgradecheck` package, both routed through the broker (no direct browser → npm/GitHub calls):

- **Web banner** above every screen (login, wizard, office shell). Compact row → "Update available: vX → vY" + click-to-copy `npm install -g wuphf@latest` + GitHub-compare link + dismissible per-version. Expandable "What's new" drawer renders a conventional-commit-grouped changelog with PR links — **Breaking changes** group promoted to the top.
- **`wuphf upgrade`** CLI subcommand: always-fresh `Check()`, prints comparison + grouped changelog. `--json` for scripting (exits non-zero + `error` field on failure; dev-build aware).

### Banner preview

**Expanded — "What's new" with full conventional-commit grouping:**

upgrade banner expanded

**Collapsed:**

upgrade banner collapsed

Captured against the broker-proxied `/upgrade-changelog` endpoint with mocked commits covering every group:
- **Breaking changes** — `feat(api)!: drop legacy /v1 endpoints (#400)` (promoted to top)
- **New features** — wiki, team
- **Bug fixes** — broker, `api handle (#42) properly (#314)` (inline `(#42)` preserved as text, trailing `#314` linked)
- **Performance**, **Refactoring**, **Documentation**, **Other changes**

Reproduce locally: build wuphf, launch with broker running, visit any wuphf URL with `?upgrade-from=v0.79.10&upgrade-to=v0.79.15` to force the banner visible without a real version mismatch (URL params validated against `VERSION_RE` so they can't inject path segments into upstream API calls).

### Architecture

- `internal/upgradecheck` — npm registry fetch + GitHub compare API + conventional-commit parser + version comparator. Defensive 30s outer cap on `http.Client` so a future caller passing `context.Background()` can't hang on a half-open socket. `FetchChangelog` self-validates `from`/`to` via `VersionParamRE` (mirrors broker's `upgradeVersionParam` and TS `VERSION_RE` — keep all three in sync).
- `internal/team/broker.go` — two new GET endpoints (`/upgrade-check`, `/upgrade-changelog`) behind `requireAuth`. Memoised via `atomic.Pointer` + `sync.Map` + `singleflight.Group` (success: 30min/1hr TTL; failure: 60s negative-cache so an outage can't bypass the broker shielding). 502 for hard transport failures on cold cache; 200 + error for partial results so the banner can degrade silently. Bad-param 400 returns the same `{commits: [], error: "..."}` shape as upstream-failure for consistent `.catch` handling.
- `cmd/wuphf/main.go` — `wuphf upgrade` subcommand with proper `flag.NewFlagSet` parsing (so `--json=true` and typo-detection work). Per-call ctx (5s npm + 8s GitHub) so a slow npm doesn't eat the changelog budget. Dev-build aware (skips comparison + non-zero exit only when `IsDevBuild=false`).
- `web/src/components/layout/UpgradeBanner.tsx` — banner component. AbortController only flag-guards against post-unmount setState (threading signal through `get()` is a deferred follow-up; commented). `safeLocalStorage{Get,Set}` wraps localStorage (Safari private mode would otherwise lock the dismiss button). `groupCommits` memoised. `setTimeout` for "Copied!" tracked in a ref + cleanup useEffect. URL-param override (`?upgrade-from=...`) for QA / screenshots, regex-validated.
- `web/src/components/layout/upgradeBanner.utils.ts` + `.test.ts` — pure helpers extracted so they can be unit-tested without React. 34 vitest cases mirror the Go test coverage so the two parsers can't drift silently.
- `internal/team/broker_upgrade_test.go` — handler tests covering path-traversal rejection, error-shape consistency on 400, auth requirement (401), method-not-allowed (405 + `Allow: GET`), and JSON wire-shape (canary that catches missing struct tags on `CommitEntry`).

### Source-build / dev guard

Source builds (`go build` with no release ldflag → `Version="dev"`) used to render a misleading "Update available: vdev → vX.Y.Z" with a wrong upgrade command. `Result.IsDevBuild` now gates both surfaces — web banner returns `null`, CLI prints "You are running a dev build … skipping npm comparison."

### Conventional-commit parser

Single grammar shared between Go (`internal/upgradecheck/upgradecheck.go`) and TS (`web/src/components/layout/upgradeBanner.utils.ts`). Captures breaking-change `!` and only strips a TRAILING `(#N)` — inline `(#42)` inside descriptions is preserved as text. Mirror tests on both sides cover breaking marker, inline PR ref, non-conventional fallback, and pre-release ordering.

## Test plan

- [x] `go test ./internal/upgradecheck/... ./internal/team/...` — all green, includes pre-release ordering, breaking-change parsing, atomic-write, CompareURL omission, broker handler shape/auth/method/path-traversal coverage
- [x] `bun test src/components/layout/upgradeBanner.utils.test.ts` — 34 cases mirror the Go coverage
- [x] `golangci-lint run` clean (post-rebase noctx + nilerr enforced)
- [x] `npm run typecheck` + `npm run build` clean; biome formatter applied
- [x] Manual: `wuphf upgrade --help` prints subcommand help; `wuphf upgrade` prints comparison + changelog; `wuphf upgrade --json` returns parseable shape; `wuphf upgrade --json=true` and `--jso` typo handled correctly via `flag.NewFlagSet`
- [x] Visual: banner renders end-to-end via broker proxy (screenshots above)

## Iteration history (16 commits)

Starts with `feat(upgrade): banner + CLI subcommand for npm version drift`. Then iterative review rounds: 9 first-round CodeRabbit findings → 9 second-round CodeRabbit findings → 17 staff review findings → 4 round-3 CodeRabbit findings, each addressed in a `fix(upgrade): …` commit. VHS baseline refreshed for the new `wuphf upgrade` row in `--help`.

## Release Notes

* **New Features**
  * Added `upgrade` subcommand to check for available updates
  * Upgrade notification banner now appears in the web interface showing available updates with recent changes and installation instructions

* **Style**
  * Added styling for the upgrade notification banner


FranDias:
## Banner preview

**Expanded — "What's new" with full conventional-commit grouping:**

upgrade banner expanded

(image upload pending — local file at `/tmp/wuphf-screenshots/banner-broker-expanded.png`)

Captured against the broker-proxied `/upgrade-changelog` endpoint with mocked commits covering every group:
- **Breaking changes** — `feat(api)!: drop legacy /v1 endpoints (#400)` (promoted to top)
- **New features** — wiki, team
- **Bug fixes** — broker, `api handle (#42) properly (#314)` (inline `(#42)` preserved as text, trailing `#314` linked)
- **Performance**, **Refactoring**, **Documentation**, **Other changes**

Reproduce locally: build wuphf, launch with broker running, then visit any wuphf URL with `?upgrade-from=v0.79.10&upgrade-to=v0.79.15` to force the banner visible without a real version mismatch (URL params are validated against `VERSION_RE` so they can't inject path segments into the upstream API call).

FranDias:
## Before / After

Real screenshots from a fresh wuphf instance booted from this branch (`feat/upgrade-banner` @ `68e18ac7`) vs origin/main (`d65cadf0`). Banner triggered via the URL-override mechanism in `readForcedPair` (`?upgrade-from=v0.0.7&upgrade-to=v0.0.8`) so the screenshot is deterministic and doesn't depend on what npm is publishing today.

### Before — origin/main, no banner code

Before: shell with no upgrade banner

### After — banner rendered, collapsed state

Shows the version pair, `What's new` toggle, `View on GitHub` link, copy-able `npm install -g wuphf@latest` command, and dismiss `×`.

After: upgrade banner shown at top of shell

### After — expanded changelog

Click `What's new` → toggle becomes `Hide changes`, changelog area mounts below the bar. (The 404 in this shot is expected — `v0.0.7` and `v0.0.8` aren't real GitHub tags so the compare API has nothing to diff. Against real published tags it would show the commit list.)

Expanded changelog state

> Screenshots live on the `assets/inline-command-screenshots` orphan branch so this branch's diff stays code-only.

FranDias (commented):
Solid overall — the broker-as-proxy shape is right, version-comparison + parser logic is well-tested in both Go and TS, the dev-build guard is well thought out, and the AbortController re-fire fix from `98f8582c` is sound. But there is one **CRITICAL wire-format bug** that means the changelog drawer in the web banner will never render correctly, plus a couple of important gaps (no caching despite the PR description claiming there is, no broker tests for the new endpoints).

Main blocker: `internal/upgradecheck.CommitEntry` has zero JSON struct tags. When the broker JSON-encodes `[]CommitEntry`, fields ship as `"Type"`, `"Scope"`, `"Description"`, `"PR"`, `"SHA"`, `"Breaking"` — capitalised — but the TS `UpgradeChangelogResponse` and the runtime mapping in `UpgradeBanner.tsx` read lowercase keys. Every entry falls through into the `parseCommit(c.description ?? "", c.sha ?? "")` defensive branch with empty inputs, producing N copies of `{type: "other", description: "", sha: undefined}`. React then renders a single duplicate-keyed item. The reason this wasn't caught: there are no broker tests for `/upgrade-changelog`, and the web component has no integration test exercising broker JSON shape.

Fix the JSON tags, add a broker-handler test that asserts the wire shape, and decide whether the cache TTL claim in the PR description ("share the npm shim's existing `~/.wuphf/cache/latest-version.json`") should actually ship in this PR or be moved to a follow-up — right now every page mount = 1 raw npm hit + (if expanded) 1 raw GitHub hit, with N tabs amplifying both.

internal/upgradecheck/upgradecheck.go
```diff
@@ -0,0 +1,347 @@
...
+	Breaking    bool
+}
+
+type githubCompareCommit struct {
+	SHA    string `json:"sha"`
+	Commit struct {
+		Message string `json:"message"`
+	} `json:"commit"`
+}
+
```
FranDias, line 165: **CRITICAL — wire format mismatch.** `CommitEntry` has no JSON struct tags, so when the broker encodes `[]CommitEntry` in `handleUpgradeChangelog` (broker.go L4687) the field names ship as `"Type"`, `"Scope"`, `"Description"`, `"PR"`, `"SHA"`, `"Breaking"`.

The TS consumer in `UpgradeBanner.tsx` (L121-132) reads lowercase: `c.type`, `c.scope`, `c.description`, `c.pr`, `c.sha`, `c.breaking`. With capitalised keys, `c.type` is `undefined` → the ternary takes the `parseCommit(c.description ?? "", c.sha ?? "")` fallback with empty inputs → every entry becomes `{type: "other", description: "", pr: null, sha: undefined, breaking: false}`. React keys collapse to a single row (all `key=undefined`), so the changelog drawer renders one empty bullet regardless of how many commits the compare returned.

**Fix:** add tags to match the TS shape:

```go
type CommitEntry struct {
    Type        string `json:"type"`
    Scope       string `json:"scope"`
    Description string `json:"description"`
    PR          string `json:"pr"`
    SHA         string `json:"sha"`
    Breaking    bool   `json:"breaking"`
}
```

And add a broker-level test that decodes the JSON and asserts the lowercase keys, so this regression can't drift again.

internal/team/broker.go
```diff
@@ -4630,6 +4633,62 @@ func (b *Broker) handleVersion(w http.ResponseWriter, r *http.Request) {
...
+			"upgrade_available": res.UpgradeAvailable,
+			"is_dev_build":      res.IsDevBuild,
+			"compare_url":       res.CompareURL,
+			"upgrade_command":   res.UpgradeCommand,
+			"error":             err.Error(),
+		})
+		return
+	}
+	_ = json.NewEncoder(w).Encode(res)
+}
```
FranDias, line 4683: **IMPORTANT — no caching despite PR description.** The PR summary promises that the broker is the right home because "we have one server-side place to add caching … share the npm shim's existing `~/.wuphf/cache/latest-version.json`". This handler does no caching at all — every web mount of `<UpgradeBanner />` is a fresh `upgradecheck.Check(ctx, nil)` call → fresh `GET registry.npmjs.org/wuphf/latest`.

With N tabs and full-page reloads on navigation that's N npm hits per launch per user. The npm shim's `~/.wuphf/cache/latest-version.json` file already exists (`npm/scripts/version-check.js` L42) and has the right TTL semantics; reusing it from `internal/upgradecheck` is genuinely a few lines.

At minimum: cap the broker to one in-flight upstream call + memoise the result for, say, 30 minutes (use a `sync.Map`-protected `cachedResult` with a `time.Time` stamp). Or: scope-reduce the PR description so it doesn't claim caching that doesn't exist. The current shape leaves us one Slashdot-scale demo away from getting our IP banned by npm or GitHub.

internal/team/broker.go
```diff
@@ -4630,6 +4633,62 @@ func (b *Broker) handleVersion(w http.ResponseWriter, r *http.Request) {
...
+	w.Header().Set("Content-Type", "application/json")
+	if err != nil {
+		_ = json.NewEncoder(w).Encode(map[string]any{
+			"commits": []any{},
+			"error":   err.Error(),
+		})
+		return
+	}
+	_ = json.NewEncoder(w).Encode(map[string]any{"commits": entries})
+}
```
FranDias, line 4720: **IMPORTANT — no broker tests for either endpoint.** `handleUpgradeCheck` and `handleUpgradeChangelog` are net-new HTTP surface and there's no test in `internal/team/` exercising them. Even a tiny `httptest.Server` test that:

- asserts the response body parses into the same shape `UpgradeBanner.tsx` reads (would have caught the missing JSON tags)
- asserts the version-param regex rejects `from=../../etc/passwd`
- asserts a request without the broker token gets 401 from `requireAuth`

…would lock in the wire contract and the auth posture. Right now both endpoints rely on integration testing through the running app, which is exactly where the JSON-tag bug slipped through.

internal/team/broker.go
```diff
@@ -4630,6 +4633,62 @@ func (b *Broker) handleVersion(w http.ResponseWriter, r *http.Request) {
...
+// as handleUpgradeCheck. The `from` and `to` query params are validated
+// against a strict version regex so a crafted call cannot inject path
+// segments into the upstream compare URL.
+func (b *Broker) handleUpgradeChangelog(w http.ResponseWriter, r *http.Request) {
+	from := strings.TrimSpace(r.URL.Query().Get("from"))
+	to := strings.TrimSpace(r.URL.Query().Get("to"))
+	if !upgradeVersionParam.MatchString(from) || !upgradeVersionParam.MatchString(to) {
+		http.Error(w, "from/to must look like v0.79.10 (or 0.79.10)", http.StatusBadRequest)
+		return
+	}
```
FranDias, line 4710: Suggestion: the version regex `upgradeVersionParam` is duplicated literally in `web/src/components/layout/upgradeBanner.utils.ts` (`VERSION_RE`) and in the upgrade-check `Result` consumers. Acceptable today (they're three terse lines and the test suites assert overlapping inputs), but worth a `// keep in sync with VERSION_RE` cross-reference comment on both sides so a future tightening doesn't drift.

Also: when validation fails, returning `400` is fine, but consider responding with the same `{commits: [], error: "…"}` shape as the upstream-failure path — the banner's `.catch` then renders a consistent error message instead of branching on HTTP status.

internal/upgradecheck/upgradecheck.go
```diff
@@ -0,0 +1,347 @@
...
+// Check fetches the latest version from npm and compares it to the running
+// build. It always returns a Result with Current populated; Latest is empty
+// when the registry call fails.
+func Check(ctx context.Context, client *http.Client) (Result, error) {
+	if client == nil {
+		client = &http.Client{Timeout: 4 * time.Second}
+	}
+	current := buildinfo.Current().Version
+	res := Result{
+		Current:        current,
```
FranDias, line 65: Subtle but important: with `client = &http.Client{}` and no timeout, an offline workstation or a hung TCP socket relies entirely on the caller's `ctx`. Both call sites (`handleUpgradeCheck` 5s, `runUpgradeCheck` 8s) do pass a deadline, so this is OK in practice. But a future caller who passes `context.Background()` will hang forever on a half-open connection.

At minimum: add a defensive `client.Timeout = 30 * time.Second` (well above any realistic context deadline so it never overrules the caller, but caps a runaway), or make the no-context-deadline case an explicit error. The comment correctly explains why `Timeout` would fight the context — but the conclusion should be "set a generous outer cap", not "set nothing".

Same applies to `FetchChangelog` at L171.

cmd/wuphf/main.go
```diff
@@ -314,6 +338,71 @@ func main() {
...
+		fmt.Printf("You are running %s v%s. That is the latest published version.\n",
+			appName, strings.TrimPrefix(res.Current, "v"))
+		return
+	}
+
+	fmt.Printf("Update available: v%s → v%s\n", strings.TrimPrefix(res.Current, "v"), strings.TrimPrefix(res.Latest, "v"))
+	fmt.Printf("Run: %s\n", res.UpgradeCommand)
+	fmt.Printf("Diff: %s\n\n", res.CompareURL)
+
+	entries, cerr := upgradecheck.FetchChangelog(ctx, nil, res.Current, res.Latest)
```
FranDias: Suggestion: the same `ctx` (8s budget total) covers both `Check` and `FetchChangelog`. If the npm registry call takes 7s, the GitHub compare call gets 1s and will almost certainly time out, printing a misleading "could not fetch changelog" warning when actually the user has plenty of time.

**Fix:** give each network call its own deadline:

```go
ctxNPM, cancelNPM := context.WithTimeout(context.Background(), 5*time.Second)
res, err := upgradecheck.Check(ctxNPM, nil)
cancelNPM()
// ...
ctxGH, cancelGH := context.WithTimeout(context.Background(), 8*time.Second)
entries, cerr := upgradecheck.FetchChangelog(ctxGH, nil, res.Current, res.Latest)
cancelGH()
```

cmd/wuphf/main.go
```diff
@@ -314,6 +338,71 @@ func main() {
...
+		fmt.Printf("You are running a dev build of %s (built from source, version=%q).\n",
+			appName, res.Current)
+		fmt.Println("Skipping npm comparison — this binary did not come from npm.")
+		return
+	}
+
+	if err != nil || res.Latest == "" {
+		fmt.Printf("You are running %s v%s.\n", appName, strings.TrimPrefix(res.Current, "v"))
+		fmt.Fprintf(os.Stderr, "warning: could not reach npm registry to check for updates (%v)\n", err)
+		os.Exit(1)
```
FranDias, line 419: Nit: when the registry is unreachable, this exits 1 in **text** mode but `runUpgradeCheck --json` (L367-371) exits 1 only when `err != nil`. If `Check` returns `(res, nil)` with an empty `Latest` (theoretically possible if the manifest decodes but has no `version` field — currently `fetchLatestVersion` rejects that with a non-nil error, but a future schema drift could change that), the JSON path would exit 0 with `latest: ""` and no `error` field. Scripted callers `wuphf upgrade --json | jq .latest` would silently get `""` and treat it as "up to date".

Defensive: in the JSON path, also exit 1 when `res.Latest == ""` (matching the text path).

web/src/components/layout/UpgradeBanner.tsx
```diff
@@ -0,0 +1,322 @@
...
+        if (ctl.signal.aborted) return;
+        if (res.current) setCurrent(res.current);
+        if (res.latest) setLatest(res.latest);
+      })
+      .catch(() => {
+        // Broker unreachable or returned a non-2xx — degrade silently.
+      });
+    return () => {
+      ctl.abort();
+    };
```
FranDias, line 99: **Misleading comment.** The `AbortController` here does not actually cancel the in-flight fetch — `get()` in `web/src/api/client.ts` (L46-66) doesn't accept a signal. All this `ctl` does is gate the `.then` body so we don't `setState` on an unmounted component, which is functionally fine but the comment says "a slow fetch from a previous run cannot resolve after we've moved on" — the fetch still resolves, we just ignore the result.

Either:
1. Extend `get<T>()` to accept an optional `signal` and pass it through to `fetch`, so the abort actually cancels (and the broker stops doing work for an unmounted tab); or
2. Reword the comment to say "flag-guard against post-unmount setState" so future readers don't assume the fetch is being cancelled.

Same issue at L147 for the changelog effect — the comment there says "collapsing (or unmounting) cancels an in-flight request" but it doesn't; it only suppresses the state update.

web/src/components/layout/UpgradeBanner.tsx
```diff
@@ -0,0 +1,322 @@
...
+        setChangelog({
+          loading: false,
+          error: e instanceof Error ? e.message : String(e),
+          commits: [],
+        });
+      });
+    return () => {
+      ctl.abort();
+    };
+  }, [expanded, current, latest]);
```
FranDias, line 173: Edge case: the effect's deps include `current` and `latest`, but `changelogFetchedRef.current` is a module-level latch — if `current` or `latest` ever change *while expanded* (e.g. a future feature where the broker re-checks periodically), the effect re-runs but skips the fetch because the ref says "already fetched", and the user keeps seeing the stale changelog.

Currently `current`/`latest` are set once from the initial broker call and never change, so this is theoretical. But if you intend to support a refresh, reset `changelogFetchedRef.current = false` on the dep change (or key the ref by `${current}->${latest}`).

web/src/components/layout/UpgradeBanner.tsx
```diff
@@ -0,0 +1,322 @@
...
+            ? {
+                type: c.type,
+                scope: c.scope ?? "",
+                description: c.description ?? "",
+                pr: c.pr || null,
+                sha: c.sha ?? "",
+                breaking: !!c.breaking,
+              }
+            : parseCommit(c.description ?? "", c.sha ?? ""),
+        );
```
FranDias, line 150: Minor: when this code lands with the JSON-tag fix on the Go side, the `c.type ? {...} : parseCommit(...)` defensive branch becomes effectively dead (every entry will have a populated `type`). That's fine — keep the defensive branch as a forward-compat seam — but worth a comment noting that the parse-from-description path is a future-proofing fallback only, not a route the current broker takes.

web/src/components/layout/upgradeBanner.utils.ts
```diff
@@ -0,0 +1,158 @@
...
+  const rest = m[4];
+  return {
+    type,
+    scope,
+    description: rest.replace(TRAILING_PR_RE, "").trim(),
+    pr: extractPR(rest),
+    sha,
+    breaking,
+  };
+}
```
FranDias, line 115: Minor parser drift vs the Go twin: the Go regex uses `(?i)` (case-insensitive) and the TS regex uses the trailing `/i` flag — same behaviour. But the Go parser does `strings.ToLower(m[1])` while the TS parser does `m[1].toLowerCase()` — also matched. Good.

One remaining drift risk: Go's `\([^)]+\)` for the scope group does not allow nested parens, and neither does TS's `\([^)]+\)`. They match. ✅

No action needed; flagging just so future maintainers know the two regexes are intentionally mirror-image.

web/src/components/layout/upgradeBanner.utils.ts
```diff
@@ -0,0 +1,158 @@
...
+  const pa = splitVersion(a);
+  const pb = splitVersion(b);
+  const len = Math.max(pa.length, pb.length);
+  for (let i = 0; i < len; i++) {
+    const x = pa[i] ?? 0;
+    const y = pb[i] ?? 0;
+    if (x !== y) return x < y ? -1 : 1;
+  }
+  return 0;
+}
```
FranDias, line 62: Edge case worth a test: `compareVersions("1.0", "1.0.0")` → both split to `[1, 0]` and `[1, 0, 0]`, padded with `0` → equal. Good. But `compareVersions("1.0.0-rc.1", "1.0.0-rc.2")` → both strip to `"1.0.0"` → equal, even though semantically rc.2 is newer.

The code comment at L43-45 acknowledges this is intentional ("npm's `latest` dist-tag is conventionally a stable release"), but if we ever publish a `1.0.0-rc.1` to `latest` and then want to nudge users to `1.0.0-rc.2`, the banner will silently say "you're up to date". Worth either an explicit test asserting the rc.1==rc.2 behaviour (locks in intent) or a TODO referencing the comment.

internal/upgradecheck/upgradecheck.go
```diff
@@ -0,0 +1,347 @@
...
+	"time"
+
+	"github.com/nex-crm/wuphf/internal/buildinfo"
+)
+
+const (
+	NPMPackage     = "wuphf"
+	GitHubRepo     = "nex-crm/wuphf"
+	NPMRegistryURL = "https://registry.npmjs.org/" + NPMPackage + "/latest"
+)
```
FranDias, line 33: Nit: `NPMRegistryURL` is a package constant, but this prevents tests from rewriting it directly — `upgradecheck_test.go` works around it by injecting a `RoundTripper` that rewrites the URL on every request (L147-151). That's clever but fragile (a future test that forgets to install the round-tripper will hit real npm).

Consider making it a `var` or exposing a setter so tests can do `upgradecheck.NPMRegistryURL = srv.URL + "/wuphf/latest"` directly. Not blocking — the current pattern works — but the round-tripper indirection is more code than the seam it replaces.

internal/upgradecheck/upgradecheck.go
```diff
@@ -0,0 +1,347 @@
...
+// IsDevVersion reports whether v is a non-released ("dev") wuphf build. This
+// matches buildinfo.Current()'s fallback when no release ldflag was set.
+func IsDevVersion(v string) bool {
+	v = strings.TrimSpace(v)
+	return v == "" || v == "dev"
+}
+
+// ── npm registry ──────────────────────────────────────────────────────────
+
+type npmManifest struct {
```
FranDias, line 109: Cross-package consistency: `IsDevVersion` checks `"" || "dev"`, but `internal/buildinfo` may have additional sentinels (e.g. a future `"unreleased"`, `"snapshot"`). If the contract is "matches `buildinfo.Current()`'s no-release-ldflag fallback", then this should ideally call into a `buildinfo.IsDev(v string)` helper rather than hard-coding the literal — otherwise a buildinfo change that adds a new sentinel will silently break the dev-build guard.

Cheap fix: move the `"" || "dev"` check into `internal/buildinfo` as `IsDev(v)` and call it from here AND from `upgradeBanner.utils.ts` (via a generated constant or a `/buildinfo` broker route). Single source of truth.

internal/team/broker.go
```diff
@@ -4630,6 +4633,62 @@ func (b *Broker) handleVersion(w http.ResponseWriter, r *http.Request) {
...
+		_ = json.NewEncoder(w).Encode(map[string]any{
+			"current":           res.Current,
+			"latest":            res.Latest,
+			"upgrade_available": res.UpgradeAvailable,
+			"is_dev_build":      res.IsDevBuild,
+			"compare_url":       res.CompareURL,
+			"upgrade_command":   res.UpgradeCommand,
+			"error":             err.Error(),
+		})
+		return
```
FranDias, line 4680: Minor: returning `200 + {error: "…"}` on registry failure is a reasonable choice for the banner's "degrade silently" UX (the comment makes the case well). But this hides outage signals from the browser console / network panel — Sentry / DevTools can't distinguish "npm down" from "working fine".

Consider responding `200 + error` only when `res.Current` is populated and `res.Latest` is empty (i.e. the partial-result case), but propagate `502 Bad Gateway` for transport errors so observability tooling sees the failure. The banner's `.catch()` in `UpgradeBanner.tsx` (L81-83) already swallows non-2xx silently, so the UX wouldn't change.

FranDias:
## Critical fix landed in 99730e13 + verified end-to-end

**The bug** (caught by staff review): `internal/upgradecheck.CommitEntry` had no JSON tags, so the broker shipped `{"Type","Scope","Description","PR","SHA","Breaking"}` while `UpgradeBanner.tsx` reads lowercase. Every commit fell into the defensive `c.type ? {…} : parseCommit("", undefined)` branch and rendered as duplicate-keyed empty bullets. The frontend tests caught nothing because they only exercise pure helpers, and there were no broker handler tests for `/upgrade-changelog`.

**The fix:**
- Lowercase JSON tags on `CommitEntry` plus a comment that ties the contract to the TS twin so future drops are noisy.
- New `internal/team/broker_upgrade_test.go` with 5 scenarios: wire shape (cache pre-seeded so no GitHub call), bad-param 400 + envelope, path-traversal rejection, auth-token requirement, GET-only method allowlist.
- Imports `golang.org/x/sync/singleflight` (referenced but unimported, breaking the build at `internal/team/broker.go:4754`).

**Step-by-step verification I ran locally:**

1. `go test ./internal/team/ -run Upgrade -v` → all 5 scenarios pass.
2. `go test ./internal/upgradecheck/...` → ok.
3. `go vet ./...` → clean.
4. `go build ./cmd/wuphf` → 58MB binary.
5. `bun run build` (web) → ok.
6. `bunx tsc --noEmit` → clean.
7. `bunx biome check` → 4 warnings, no errors.
8. `bunx vitest run` → 461/461.
9. Booted wuphf in an isolated `WUPHF_RUNTIME_HOME`; hit `GET /upgrade-changelog?from=v0.79.16&to=v0.79.17` on the broker port with the bearer token. Response:

   ```
   {"commits":[{"type":"test","scope":"web/e2e","description":"close TUI-vs-Web parity gaps in Playwright","pr":"339","sha":"e2ff794edabea338267752b618bca472ed9531cc","breaking":false}]}
   ```

   Lowercase keys confirmed. Pre-fix this would have been `{"Type":"test","Scope":"web/e2e",…}` and the banner would have shown an empty bullet.

10. Loaded `http://localhost:7895/?upgrade-from=v0.79.16&upgrade-to=v0.79.17` in headless Chromium, clicked **What's new**, screenshot:

    Pre-fix screenshot for reference (compare to fixed render)

    The fixed render shows `OTHER CHANGES → web/e2e | close TUI-vs-Web parity gaps in Playwright #339` with real content — no empty bullets, no console errors, `pr: "339"` rendered as the linked `#339` chip.

**Not addressed in this commit** (left for follow-up since they're nits, not blockers):
- Cache claim in PR description vs no actual cache wired (the staff review's most-important nit).
- AbortController-without-signal misleading naming in `UpgradeBanner.tsx` and the `get()` helper.
- Other suggestion-level CodeRabbit findings.

Will not auto-merge — needs an approving review.

PR #348, opened 2026-04-27
fix(upgrade): address remaining PR #344 review comments

FranDias:
## Summary

Follow-up to #344. Two visible bugs and four small cleanups that surfaced during review and weren't folded into the original merge.

### UI / a11y
- **`.upgrade-banner-dismiss:hover`** — switch from `rgba(0, 0, 0, 0.06)` (invisible against `var(--bg-card)` in dark mode) to `var(--blue-bg)` matching the copy button.
- **Banner role** — `role="status"` → `role="region"` + `aria-label="Upgrade available"` so it stops behaving as a live region for an interactive container.
- **Toggle ↔ drawer wiring** — `aria-controls` on the "What's new" toggle, `useId()`-stable `id` on the changelog drawer.
- **`entry.pr` validation** — only render as a GitHub `/pull/<n>` anchor when `^\d+$`. Falls back to plain text otherwise so a future broker change emitting a non-numeric token can't produce a malformed URL.

### Forward-compat
- **`changelogFetchedRef`** is now a `"${current}→${latest}"` string instead of a bare boolean. A future periodic broker poll that mutates current/latest re-triggers the changelog fetch instead of silently rendering the stale cache.

### Test ergonomics / single source of truth
- **`NPMRegistryURL`** is a `var` so tests can swap it directly. Per-test custom `RoundTripper` and host-rewrite shenanigans dropped; `mockNPMRegistry` now mutates the package-level URL and tests pass `nil` for the `http.Client`.
- **`buildinfo.IsDev(v)`** — new helper centralises the `""` / `"dev"` sentinel set that `Current()` falls back to. `upgradecheck.IsDevVersion` was a literal duplicate of the same predicate; deleting it means a future buildinfo addition (e.g. `"snapshot"`, `"unreleased"`) flows through every caller automatically.

## Test plan
- [x] `go test ./internal/buildinfo/... ./internal/upgradecheck/... ./internal/team/... ./cmd/wuphf/...` — all green
- [x] `golangci-lint run` clean on the touched packages
- [x] `bun run typecheck` + `bun run build` clean
- [x] `bun test src/components/layout/upgradeBanner.utils.test.ts` — 39/39
- [x] `bunx biome check` — only pre-existing nursery warnings on untouched code
- [ ] Manual: visit any wuphf URL with `?upgrade-from=v0.79.10&upgrade-to=v0.79.15` and confirm banner renders, hover state visible in dark mode, screen reader announces "Upgrade available, region", "What's new" expands the labelled drawer

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented redundant changelog refetches by caching per current→latest version pair.
  * Improved handling of pull request references to avoid broken links (falls back to plain text when link is invalid).

* **Improvements**
  * Enhanced upgrade banner accessibility with a landmark role, stable aria-controls/IDs, and consistent hidden handling.
  * Updated dismiss button styling for better contrast.
  * More robust upgrade/version detection across dev and release builds.

* **Tests**
  * Added coverage for PR-link validation and dev-sentinel detection.


web/src/components/layout/UpgradeBanner.tsx
```diff
@@ -314,64 +337,93 @@ export function UpgradeBanner() {
...
-          {!(changelog.loading || changelog.error) &&
-            changelog.commits.length === 0 && (
+      {/*
+        Mounted unconditionally so the toggle's `aria-controls={changelogId}`
+        always resolves to a real element — collapsed visibility is gated by
+        the `hidden` attribute. The previous `{expanded && …}` form left
+        aria-controls pointing at a missing node when collapsed, which the
+        ARIA APG disclosure pattern warns AT support is inconsistent for.
+        Children stay gated on `expanded` so the changelog fetch effect
+        (line 122) keeps its fetch-on-expand semantics — nothing renders or
```
FranDias, line 347: Nit: the comment says "the changelog fetch effect (line 122)", but line 122 is the start of the *comment block* describing that effect — the actual `useEffect` is at line 130. Either point at line 130 or drop the line number entirely so it doesn't drift again the next time someone edits above it.

FranDias (commented):
Verified end-to-end: contrast claim holds, disclosure-pattern fix is correct, Go/TS twins behave identically post-merge.

## What I checked

**`fc395313` — dismiss hover + `aria-controls` mount**
- `--blue-bg` resolves to `#e5feff` in both `nex.css` and `global.css` and is *not* overridden in `nex-dark.css`, so the upgrade banner has the same cyan background in both themes. 12% blue overlay (`rgba(6,157,228,0.12)`) blended onto `#e5feff` lands at `~rgb(202,242,252)` — ΔE ≈ 10+ from the parent, well above the JND threshold. Comment is accurate.
- `hidden={!expanded}` + inner `{expanded && (…)}` is the right primitive: target stays in the DOM so `aria-controls` always resolves, browser default `[hidden] { display: none }` applies (grepped — no overrides), and the inner gate preserves the line-130 `useEffect`'s fetch-on-expand semantics. Children mount/unmount on expand exactly as before.

**`ac683ce1` — merge resolution for `IsDevVersion` / `isDevVersion`**
- `buildinfo.IsDev` remains the literal-sentinel SSOT; the extended sub-0.1.0 + `VersionParamRE` guards from #352 are layered on top. Go and TS produce identical truth tables for every case in the shared table tests (`0.0.7.1`, `0.0.0`, `v0.0.1`, `0.1.0`, garbage, `"1"`, whitespace-padded `"  dev  "`). All three regex twins (`VERSION_RE` / `VersionParamRE` / `upgradeVersionParam`) are byte-identical.
- Test file resolution preserved every test — `pinCurrentVersion` / `mockNPMRegistry` / `pinNPMRegistryURL` are coherent, the dropped `roundTripFunc` is dead with the URL-pin pattern, and the new `TestCheckShortCircuitsForStaleVersionFile` pins the #350 reproducer.
- Merged-in `release-drift.yml`, the `Check VERSION sentinel` CI step, and the `dev` / `0.0.0` package-version pins all reinforce (don't contradict) this PR's `IsDevVersion` extension.

**Earlier commits** — `prGitHubURL` extraction is well-tested (10 cases including `310; rm -rf /`, `3.10`, leading whitespace), `.upgrade-banner-pr-text` correctly drops the link affordance for non-numeric tokens, `npmRegistryURL` rename to unexported is grep-clean.

**Tooling** — `go test ./internal/upgradecheck/... ./internal/buildinfo/...` green, `go vet` clean, `tsc --noEmit` clean, `vitest` 56/56 (PR body still says 39 — stale, not a code issue).

One stylistic nit posted inline. Otherwise this is solid; happy to ship.

PR #352, opened 2026-04-27
fix(upgrade): suppress banner on dev/sentinel versions + drift monitor

### LINKED ISSUE #350: Banner does not represent the correct version
Banner does not accurately represent the right version
```
  ┌─────────────────────────┬───────────────────────────────────────────────────────────────────────┐
  │                         │                                 Value                                 │               
  ├─────────────────────────┼───────────────────────────────────────────────────────────────────────┤               
  │ Our fork checked out at │ v0.83.1 (tip of main, latest tag)                                     │               
  ├─────────────────────────┼───────────────────────────────────────────────────────────────────────┤               
  │ Commits behind main     │ 0                                                                     │               
  ├─────────────────────────┼───────────────────────────────────────────────────────────────────────┤               
  │ VERSION file contents   │ 0.0.7.1 ← stale internal value, last touched 2026-04-23               │
  ├─────────────────────────┼───────────────────────────────────────────────────────────────────────┤               
  │ npm publishes as        │ 0.79.2                                                                │
  ├─────────────────────────┼───────────────────────────────────────────────────────────────────────┤               
  │ What banner compares    │ binary's VERSION (0.0.7.1) vs npm (0.79.2) → false "update available" │
  └─────────────────────────┴───────────────────────────────────────────────────────────────────────┘               
                             
  So the UI's "update available v0.0.7.1 → v0.79.2" is an artifact of the in-app upgrade-banner feature that landed 
  recently (commit d06a1f95 feat(upgrade): in-app upgrade banner). It's comparing apples to oranges.
```                                         

Image

### PR DESCRIPTION
**Pull request by FranDias**
## Summary

Fixes #350. Contributor and source builds were seeing `update available v0.0.7.1 → v0.79.2` because `internal/buildinfo/VERSION` had been hand-edited to a stale `0.0.7.1`, then the in-app upgrade banner (#344) compared it against npm `latest`. Running `wuphf upgrade` would actually downgrade.

Compounding it: every Release run since v0.79.3 has been stuck on the `npm-publish` GitHub Environment gate, leaving npm `latest` 13 tags behind git. The banner was therefore advertising an old release.

### What changes

- `internal/buildinfo/VERSION` → literal `dev`. Source builds now resolve to a dev sentinel that `IsDevVersion` classifies correctly: banner hides, `wuphf upgrade` short-circuits.
- `IsDevVersion` (Go + TS twin) hardened to reject garbage strings and any sub-0.1.0 version as a sentinel — defense in depth so a future stray `VERSION` edit can't reproduce this downgrade.
- New CI lint step that fails the build if `internal/buildinfo/VERSION` is anything but `dev`.
- Regression test pinning `currentVersion=0.0.7.1` against npm `latest=0.79.2` and asserting `IsDevBuild=true` / `UpgradeAvailable=false` (Go + TS).
- New `.github/workflows/release-drift.yml`: daily monitor comparing four distribution surfaces — latest git tag on main, latest GitHub Release `tag_name`, npm registry `latest`, and `npx --yes wuphf@latest --version`. Catches postinstall pulling the wrong binary, goreleaser shipping without the version ldflag, or the `npm-publish` env gate stalling. Also runs after every Release workflow completion so a stuck release surfaces in minutes, not 24 hours.
- Updated VHS snapshots (`testdata/vhs/help.txt`, `testdata/vhs/version.txt`) to match the corrected `vdev` self-report.

### Out of scope

This PR does not unblock the queued `npm-publish` env gate. Someone with environment-approval rights still needs to drain the backlog so npm `latest` catches up to the real git tag — but the new drift workflow will keep yelling until that happens.

## Test plan

- [x] `go test ./internal/upgradecheck/... ./internal/buildinfo/...` — passes, including new `TestCheckShortCircuitsForStaleVersionFile`
- [x] `bunx vitest run src/components/layout/upgradeBanner.utils.test.ts` — 46 tests pass
- [x] `bun run typecheck` + `bun run build` — clean
- [x] `testdata/vhs/check.sh` — both `help` and `version` snapshots match
- [x] `go run ./cmd/wuphf --version` reports `wuphf vdev`, banner hidden in dev
- [x] `actionlint` clean on `release-drift.yml` (and on the `ci.yml` change)
- [ ] CI green
- [ ] Manual: scrub through banner with URL override (`?upgrade-from=v0.79.10&upgrade-to=v0.79.15`) to confirm preview path still works

## Summary by CodeRabbit

* **New Features**
  * Added automated release drift detection workflow to validate version consistency across git tags, GitHub Releases, npm registry, and binary output.

* **Improvements**
  * Enhanced development version detection with stricter validation logic.
  * Added CI validation gate to enforce development build version requirements.

* **Tests**
  * Expanded version handling tests with edge cases and regression scenarios.
  * Updated version output fixtures to reflect development version format.

### CODE REVIEW

File: web/src/components/layout/upgradeBanner.utils.ts [hunk 1]
```diff
@@ -32,11 +32,17 @@ export const VERSION_RE =
   /^v?\d+(\.\d+){1,3}(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?$/;
 
 // Buildinfo's "dev" sentinel — see internal/buildinfo/buildinfo.go. Keep
-// in sync with upgradecheck.IsDevVersion.
+// in sync with upgradecheck.IsDevVersion. Also rejects garbage strings and
+// any sub-0.1.0 version as a sentinel (#350: a stale VERSION file shipping
+// "0.0.7.1" passed through this guard and produced a false "upgrade
+// available v0.0.7.1 → v0.79.2" — actually a downgrade).
```
FranDias: **SUGGESTION** — comment over-promises for #350.

`UpgradeBanner.tsx:179` does `if (isDevBuildSrv || isDevVersion(current)) return false;` — the server flag is checked *first*. With this PR's Go-side fix, the broker always returns `is_dev_build: true` for stale current strings, so this TS local check is dead code on the production code path. It only fires on the URL-override fallback (`?upgrade-from=0.0.7.1&upgrade-to=...`), which is a QA-only path and was *not* the #350 reproducer.

Not harmful to keep, but the comment block claiming this guards #350 is misleading. Either:
- Drop the comment's #350 reference and frame it as "keep the TS twin in sync with the Go", or
- Note explicitly that this is a URL-override-only safeguard.

internal/upgradecheck/upgradecheck.go
```diff
@@ -94,9 +94,27 @@ func Check(ctx context.Context, client *http.Client) (Result, error) {
...
 func IsDevVersion(v string) bool {
 	v = strings.TrimSpace(v)
-	return v == "" || v == "dev"
+	if v == "" || v == "dev" {
+		return true
+	}
+	if !VersionParamRE.MatchString(v) {
+		return true
+	}
+	if compareVersions(v, "0.1.0") < 0 {
```
FranDias, line 120: **SUGGESTION** — `< 0.1.0` permanently retires the `0.0.x` namespace for any future wuphf-released version.

Not a real concern for this product (wuphf is at `v0.83.x`), but worth being explicit in the comment that this is a one-way ratchet — if the team ever wanted to publish a `v0.0.x` line (a sub-package, an alpha line, etc), it would silently classify as dev. Threshold could be `0.0.1` (treat only `0.0.0` as exact sentinel) or `0.1.0` (the current choice). The current choice is correct for #350; just make the trade-off explicit so a future maintainer doesn't loosen it and reintroduce the gap.

.github/workflows/release-drift.yml
```diff
@@ -0,0 +1,132 @@
...
+          # thinks it is. This catches a class of bug the registry
+          # version alone misses: postinstall could be downloading a
+          # mismatched binary, or goreleaser could have shipped without
+          # the ldflag so the binary self-reports `dev` despite being
+          # tagged. The `vdev` regression from #350 would fail here.
+          npx_out=$(npx --yes "wuphf@$npm_version" --version 2>&1 || true)
+          echo "npx output     = $npx_out"
+          # Match the same pattern release.yml's smoke-test uses:
+          # ` v<version>` with optional trailing whitespace/EOL.
+          if ! echo "$npx_out" | grep -E -q " v${npm_version}(\s|$)"; then
```
FranDias, line 104: **NIT** — version dots are unescaped in the grep regex.

`grep -E -q " v${npm_version}(\s|$)"` — for `npm_version=0.79.2`, the dots match any char, so a hypothetical `v0a79b2` would also match. In practice the binary's output is fixed format and this is a smoke check, so the risk is theoretical. But the same pattern is mirrored from `release.yml:139` so this is also a chance to fix both — replace dots with `\.` in both places.

.github/workflows/release-drift.yml
```diff
@@ -0,0 +1,132 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+
+          # --- 1. Latest git tag on the default branch ---
+          # `--merged origin/main` filters out stray tags on side branches
+          # so a partial release tag pushed to a feature branch doesn't
+          # mask drift on the real release line.
+          git_tag=$(git tag --merged origin/main --sort=-v:refname | grep -E '^v[0-9]' | head -n1)
```
FranDias, line 68: **NIT** — `--merged origin/main` depends on `origin/main` being a populated remote-tracking ref after `actions/checkout@v6` with `fetch-depth: 0`.

In practice this works (fetch-depth:0 fetches all branches/tags), but worth a one-line sanity check (`git rev-parse origin/main >/dev/null 2>&1 || { echo "::error::origin/main not fetched"; exit 1; }`) before the tag query so the failure mode if checkout's defaults change is "loud assertion failure" rather than "silently empty tag list → no v* tags found" (the existing error message would be confusing because v* tags DO exist in this case).

.github/workflows/release-drift.yml
```diff
@@ -0,0 +1,132 @@
...
+    # Daily at 14:00 UTC — well after the typical release window so a
+    # fresh release has time to propagate through all four surfaces
+    # before the check runs. Cron is approximate; allow ~15min jitter.
+    - cron: "0 14 * * *"
+  workflow_dispatch:
+  # Also run after every Release workflow run (success or failure) so a
+  # stuck env gate or partial publish surfaces in minutes, not 24 hours.
+  workflow_run:
+    workflows: [Release]
+    types: [completed]
```
FranDias, line 33: **NIT** — `workflow_run` triggers run in the context of the default branch, *not* the head branch of the triggering Release workflow. That's intended here (we're checking origin/main against the release surfaces) but worth one explicit sentence in the header comment so a future maintainer doesn't try to inspect `github.ref` to scope the check to the specific tag that triggered the run.

internal/upgradecheck/upgradecheck_test.go
```diff
@@ -156,6 +190,11 @@ func TestCheckHitsRegistry(t *testing.T) {
 		return http.DefaultTransport.RoundTrip(req)
 	})}
 
+	// Pin a non-dev current so this test exercises the registry path
+	// regardless of what buildinfo.Current() resolves to. With the embedded
+	// VERSION file now `dev` (#350), an unpinned test would short-circuit
+	// at IsDevBuild and never assert UpgradeAvailable.
+	pinCurrentVersion(t, "1.2.3")
```
FranDias, line 197: **NIT** — good catch pinning `currentVersion` here. Without this, `TestCheckHitsRegistry` would have been silently regressed by the `dev`-VERSION change (test would pass by short-circuiting at `IsDevBuild`, never actually asserting `UpgradeAvailable`). Worth carrying the same vigilance forward — any future test landing in this package needs to consider whether the now-`dev` default makes the assertion vacuous.

FranDias (commented):
## Verdict

Verified end-to-end that the banner is now correctly suppressed on a source build:

```
$ go build -o /tmp/wuphf ./cmd/wuphf && /tmp/wuphf --version
wuphf vdev
$ curl -H "Authorization: Bearer $TOKEN" http://127.0.0.1:17890/upgrade-check | jq
{
  "current": "dev",
  "latest": "0.79.2",
  "upgrade_available": false,
  "is_dev_build": true,
  ...
}
```

Also verified the sub-0.1.0 sentinel: a binary built with `-X buildinfo.Version=0.0.7.1` ldflag still classifies as `is_dev_build: true` and reports `upgrade_available: false`. Defense-in-depth holds. `go test ./internal/upgradecheck/...` and `bun test web/src/components/layout/upgradeBanner.utils.test.ts` pass clean.

The fix is correct. Two material gaps to address before merge:

### Gap 1 — root `package.json` and `web/package.json` still hold the same class of stale-version string

```
$ rg -n '"version":' --glob '!node_modules' --glob '*.json'
package.json:12:  "version": "0.0.7.1"
web/package.json:4:  "version": "0.0.6.0",
web/e2e/package.json:4:  "version": "0.0.1",
npm/package.json:3:  "version": "0.0.0",
```

The new CI guard ratchets `internal/buildinfo/VERSION` only. Both root `package.json` (private dev tooling) and `web/package.json` (private app) still pin synthetic `0.0.x.y` strings — exactly the class of stale-version that produced #350. Yes, neither is published, but:

- A future contributor fixing a similar bug may hand-edit any of them (just like the original incident).
- `npm/package.json` is at `"0.0.0"` (npm's documented unpublished sentinel) — already the right pattern. Adopting it for the other two would unify the convention and make the new sub-0.1.0 sentinel coherent.
- Or extend the CI guard to all three.

### Gap 2 — the TS `isDevVersion` defense-in-depth update is mostly cosmetic for #350

`UpgradeBanner.tsx:179` reads `if (isDevBuildSrv || isDevVersion(current)) return false;` — server flag first. With this PR's Go-side fix the broker always returns `is_dev_build: true` for `0.0.7.1`, so the TS local check is dead on the production path and only fires on the URL-override fallback (a QA preview path that wasn't the #350 reproducer). The comment block claiming this guards #350 over-promises. Either drop the TS change or accurately describe its scope as URL-override-only.

### Nits inline

Workflow design is otherwise good — comments are excellent, error messages are actionable, surfaces are well-chosen. `TestCheckHitsRegistry` correctly pins `currentVersion` to dodge the now-vacuous-passing path; nice catch.

FranDias:
Addressed all three points from the review in commit 51287b89:

1. **Stale version strings** — root `package.json` (`0.0.7.1`) and `web/package.json` (`0.0.6.0`) both normalized to `"0.0.0"`, matching `npm/package.json`'s release-overwritten default. CI lint guard extended to ratchet both files (`node -p "require('./<pkg>').version"` per file, exits 1 on drift). Both are `"private": true` so `0.0.0` is purely a sentinel — the field is never published anywhere.

2. **TS `isDevVersion` comment** — rewritten to make explicit that the local check is dead code on the production banner path (`UpgradeBanner.tsx` trusts the server-authoritative `is_dev_build` flag first) and only fires on the URL-override QA preview path. Twin still updated for QA-pair safety, but the comment no longer over-promises the contribution to #350.

3. **`< 0.1.0` ratchet** — Go-side comment now calls out that this permanently retires the 0.0.x namespace as a published wuphf version (we're at v0.83.x, so this is intent, not oversight) and tells a future maintainer to explicitly own the regression if they ever loosen it.

PR #354, opened 2026-04-27
chore(release): remove npm-publish environment gate

FranDias:
## Summary

The `npm-publish` environment gate paused every Release run waiting on manual approval. It worked once — and then nobody approved the queue. ~13 tags piled up between v0.79.3 and v0.83.1, npm `latest` fell behind by months, and #350's downgrade-prompt bug surfaced because of the resulting drift.

`NPM_TOKEN` is at the repo level (verified via `gh secret list`), not env-scoped, so removing the `environment:` block does not change publish auth.

The defensive value the gate was supposed to provide (catch a malicious tag push) is already covered by three other layers:

- **Branch-protected `main`** — tags created from main come from reviewed PRs
- **`smoke-test-npx` in this workflow** — fails the run if the published binary self-reports a different version from the tag
- **`release-drift.yml` (added in #352)** — screams if any of the four distribution surfaces (git tag, GitHub Release, npm latest, npx-installed binary) drift apart, on schedule + after every Release completion

If we ever want humans-in-the-loop again, the right pattern is a tag-name convention (e.g. `-rc.N` suffix skips `--tag latest`) instead of a silent env gate that nobody monitors.

## Follow-up after merge

The `npm-publish` GitHub environment itself can be deleted at https://github.com/nex-crm/wuphf/settings/environments/14576656257 once this lands — but it's harmless to leave around (no longer referenced).

## Test plan

- [x] `actionlint .github/workflows/release.yml` clean
- [x] `gh secret list` confirms `NPM_TOKEN` exists at repo level
- [x] `gh api repos/.../environments/npm-publish/secrets` returns no env-scoped secrets
- [ ] CI green
- [ ] Next release tag publishes without manual intervention
- [ ] `release-drift.yml` fires post-publish and reports all four surfaces aligned

## Summary by CodeRabbit

* **Chores**
  * Updated release workflow diagnostics to provide clearer guidance when deployment issues occur.
  * Streamlined automated release process by removing manual approval gate requirements, allowing releases to proceed automatically.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,14 +26,20 @@ concurrency:
 jobs:
   release:
     runs-on: ubuntu-latest
-    # Env gate sits on the FIRST job that ships anything (goreleaser uploads
-    # GitHub Release artifacts). Putting it on `publish-npm` instead would
-    # leave a published GitHub Release with no corresponding npm package on
-    # rejection, requiring manual cleanup. Pausing here means the human
-    # approves once and both surfaces ship together — or nothing ships.
-    environment:
-      name: npm-publish
-      url: https://www.npmjs.com/package/wuphf
+    # No environment gate: tags pushed to this repo are intended releases,
+    # full stop. The previous `npm-publish` env gate paused every run
+    # waiting on a human approval and the queue stalled for ~13 tags
+    # (#350) before anyone noticed the banner was advertising stale
+    # versions. The defensive value of the gate (preventing an accidental
+    # npm publish from a malicious tag push) is already covered by:
+    #   - branch-protected `main`, so tags on main are reviewed in PRs
+    #   - the smoke-test-npx job below failing the run if the published
+    #     binary self-reports the wrong version
+    #   - .github/workflows/release-drift.yml screaming if any of the
+    #     four distribution surfaces ever drift apart
+    # If we ever need a humans-in-the-loop step again, prefer a tag-name
+    # convention (e.g. `vX.Y.Z-rc.N` skips publish-to-latest) over a
+    # silent env gate that nobody monitors.
     permissions:
       contents: write
     outputs:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,16 +30,28 @@ jobs:
     # full stop. The previous `npm-publish` env gate paused every run
     # waiting on a human approval and the queue stalled for ~13 tags
     # (#350) before anyone noticed the banner was advertising stale
-    # versions. The defensive value of the gate (preventing an accidental
-    # npm publish from a malicious tag push) is already covered by:
-    #   - branch-protected `main`, so tags on main are reviewed in PRs
-    #   - the smoke-test-npx job below failing the run if the published
-    #     binary self-reports the wrong version
-    #   - .github/workflows/release-drift.yml screaming if any of the
-    #     four distribution surfaces ever drift apart
-    # If we ever need a humans-in-the-loop step again, prefer a tag-name
-    # convention (e.g. `vX.Y.Z-rc.N` skips publish-to-latest) over a
-    # silent env gate that nobody monitors.
+    # versions.
+    #
+    # The defensive value the gate was supposed to provide (catch an
+    # unintended npm publish) is covered by these layered controls,
+    # accurate to today's setup:
+    #
+    #   1. Branch-protected `main` (1 review, dismiss-stale, conversation
+    #      resolution). All commits land via reviewed PR. `auto-release.yml`
+    #      then creates `v*` tags from those reviewed commits — so the
+    #      invariant is "every tag points at a reviewed commit", NOT
+    #      "every tag itself was reviewed".
+    #   2. smoke-test-npx (below) runs AFTER npm publish and fails the
+    #      Release run if the just-published binary self-reports a
+    #      different version from the tag. Detective, not preventive —
+    #      catches a corrupt artifact, not a malicious tag.
+    #   3. release-drift.yml runs daily + after every Release completion,
+    #      screaming if git tag, GitHub Release, npm latest, or the
+    #      npx-installed binary's self-report ever diverge.
+    #
+    # If we ever need a preventive humans-in-the-loop step again, prefer
+    # a tag-name convention (e.g. `vX.Y.Z-rc.N` skips `--tag latest`) or
+    # a tag ruleset over a silent env gate that nobody monitors.
     permissions:
       contents: write
     outputs:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -26,14 +26,20 @@ concurrency:
...
-    environment:
-      name: npm-publish
-      url: https://www.npmjs.com/package/wuphf
+    # No environment gate: tags pushed to this repo are intended releases,
+    # full stop. The previous `npm-publish` env gate paused every run
+    # waiting on a human approval and the queue stalled for ~13 tags
+    # (#350) before anyone noticed the banner was advertising stale
+    # versions. The defensive value of the gate (preventing an accidental
+    # npm publish from a malicious tag push) is already covered by:
+    #   - branch-protected `main`, so tags on main are reviewed in PRs
```
FranDias: **IMPORTANT** — defense-in-depth claim is partially wrong.

The comment says "branch-protected `main`, so tags on main are reviewed in PRs". The PR review covers the *commit*, but `auto-release.yml` then auto-creates a tag on every push to `main` (paths-ignore: `index/**`, `team/**`) and `gh workflow run release.yml --ref <tag>` dispatches this workflow. There is no human reviewing the tag itself.

That's fine as a policy choice (tags become a derived artifact of merges), but the comment should reflect it — otherwise the next person who reads it assumes humans inspect every `v*` tag.

Suggested rewrite:
```
#   - branch-protected `main` (reviewed PRs land before auto-release.yml
#     tags HEAD); a malicious tag would require a malicious merge, which
#     requires bypassing CODEOWNERS / required reviews
```

.github/workflows/release.yml
```diff
@@ -26,14 +26,20 @@ concurrency:
...
+    #   - the smoke-test-npx job below failing the run if the published
+    #     binary self-reports the wrong version
```
FranDias: **IMPORTANT** — `smoke-test-npx` runs *after* publish, so it doesn't "fail the run" in a way that prevents a bad publish.

The job has `needs: [release]`, and `release` is the job that runs `npm publish` (line 116). By the time smoke-test fires, npm `latest` already points at the new version. The smoke-test catches version-ldflag bugs for *the next user* (so the drift monitor can scream), not for npm.

This is fine in practice, but the workflow comment reads as if smoke-test is a pre-publish gate. Recommend rewording to:
```
#   - the smoke-test-npx job below reporting a hard CI failure if the
#     just-published binary self-reports a wrong version (after-the-fact
#     detection — npm has already shipped at this point, but the failure
#     is loud)
```

If you want a true pre-publish gate, split the publish step into its own `publish-npm` job after smoke-tests an unpublished tarball (`npm pack` + `npx ./tarball.tgz --version`). Out of scope for this PR but worth a follow-up.

.github/workflows/release.yml
```diff
@@ -26,14 +26,20 @@ concurrency:
...
+    #   - .github/workflows/release-drift.yml screaming if any of the
+    #     four distribution surfaces ever drift apart
```
FranDias: **IMPORTANT** — stale error text in `release-drift.yml` (out of this diff but caused by this PR).

`release-drift.yml:125` currently emits:
```
::error::most likely: the npm-publish environment gate is awaiting approval
```

After this PR there is no gate, so when GH-vs-npm drift next fires, this line misdirects the on-call to a non-existent cause. Worth a one-line fix in the same PR:
```
echo "::error::most likely: the publish-to-npm step in release.yml failed (NPM_TOKEN expired, registry hiccup, or version-already-published)"
echo "::error::check: gh run list --workflow=release.yml --limit=5 (look for the failing run on this tag)"
```

The header comment block (lines 4-7) referencing the gate as the original incident is fine to leave as historical context — it's the runtime error message that matters.

.github/workflows/release.yml
```diff
@@ -26,14 +26,20 @@ concurrency:
...
+    # If we ever need a humans-in-the-loop step again, prefer a tag-name
+    # convention (e.g. `vX.Y.Z-rc.N` skips publish-to-latest) over a
+    # silent env gate that nobody monitors.
```
FranDias: **SUGGESTION** — the `npm-publish` GitHub Environment is still configured with required reviewers (`FranDias`, `najmuzzaman-mohammad`) and `can_admins_bypass: true`. Now that nothing references it, delete it (or keep it but document its state):

- If kept: someone editing release.yml in 6 months sees "there's an environment named npm-publish, must matter" → adds `environment: npm-publish` back → silent gate returns. The PR description acknowledges deletion is optional, but I'd argue it's load-bearing-by-omission and worth deleting.
- If deleted: `https://github.com/nex-crm/wuphf/settings/environments/14576656257` → Delete environment.

Not blocking — fine to follow up.

.github/workflows/release.yml
```diff
@@ -26,14 +26,20 @@ concurrency:
...
+    # (#350) before anyone noticed the banner was advertising stale
+    # versions. The defensive value of the gate (preventing an accidental
+    # npm publish from a malicious tag push) is already covered by:
+    #   - branch-protected `main`, so tags on main are reviewed in PRs
+    #   - the smoke-test-npx job below failing the run if the published
+    #     binary self-reports the wrong version
+    #   - .github/workflows/release-drift.yml screaming if any of the
+    #     four distribution surfaces ever drift apart
+    # If we ever need a humans-in-the-loop step again, prefer a tag-name
+    # convention (e.g. `vX.Y.Z-rc.N` skips publish-to-latest) over a
```
FranDias: **SUGGESTION** — cadence policy not addressed in this PR, but worth a follow-up issue.

With the gate removed and `auto-release.yml` running on every push to `main`, the team's "ship every few hours" pattern means every merge is now an npm `latest` publish + an in-app banner prompt. `UpgradeBanner.tsx:19,104,223` keys dismissal on the *target* version, so dismissing v0.83.2 is undone the moment v0.83.3 ships. For a project shipping ≥1× per day, this turns the banner into nagware.

Three axes worth sequencing:
- **Banner UX (cheapest)**: change `DISMISSED_KEY` to store `{lastDismissedAt: ISO, lastDismissedVersion: string}` and suppress for `min(snoozeUntil, nextMajorOrSecurityBump)`. Cap snooze at ~5 days; never honor it across a major-version bump. Touches `UpgradeBanner.tsx`, `upgradeBanner.utils.ts`, `upgradeBanner.utils.test.ts`, plus Go-side `internal/upgradecheck` if we want server-flagged "security bump" overrides.
- **Tag-side batching (medium)**: gate `auto-release.yml` on a label like `release-worthy` or a daily cron that picks the latest commit on `main` with a non-trivial diff. Lets feature-by-feature merges land without each becoming `wuphf@latest`.
- **Dist-tag policy (heaviest)**: every tag → `next` automatically; only a daily "promote" workflow (or human run) bumps `latest`. Aligns with the npm idiom and keeps `npx wuphf` users on stable releases.

My take: ship banner UX first (1 PR, no infra change, biggest user-pain reduction). Then revisit batching if banner UX alone doesn't kill the friction.

Issue #313, opened 2026-04-26
opencode provider: stdout/stderr parser, 4-min hardcoded timeout, and missing deliverable verification cause silent fabrication of task completion


alisorcorp:
## Summary

While exercising the `opencode` provider with a multi-agent ingest task (read N source markdown files, synthesize entity briefs, write to `~/.wuphf/wiki/`), I hit three compounding issues that combine into a serious user-visible failure: **the broker transitions tasks to `done` without verifying that the deliverable referenced in the agent's reply actually exists on disk**. Across multiple model backends (Sonnet, DeepSeek V4 Pro, Kimi K2.6) the agents repeatedly produced confident "all N briefs delivered" replies — formatted markdown tables, ✅ checkmarks, file sizes — while either (a) no file landed where the agent claimed, (b) files landed at a literal `~/` subdirectory under wuphf's launch CWD, or (c) the proper WikiWorker / archivist git pipeline was bypassed entirely.

> **Edit (post-filing):** My original Finding 3 framing said the broker marks `done` purely from narrative text. That mechanism is wrong — see [@FranDias's comment](#issuecomment-4329259855) confirming the actual transition requires a `team_task action="complete"` MCP call. The headline bug is the same shape (no deliverable verification on `done`), the cause is in the `complete` handler. Finding 3 below has been rewritten to reflect that, with the tilde-expansion failure mode kept as a concrete repro that exercises the verification gap.

The findings below are in causal order; earlier issues compound into later ones.

## Environment

- macOS 26.5 Beta (Darwin 25.5.0)
- `wuphf` @ HEAD via `npx -y wuphf`
- `opencode` 1.14.25
- Models exercised: DeepSeek V4 Pro, DeepSeek V4 Flash, Kimi K2.6 (all via OpenRouter), and Anthropic Sonnet via the Claude Code provider for comparison
- Pack: `lead-gen-agency`

---

## Finding 1 — opencode default `run` output is not stdout-only; wuphf parser sees nothing

`internal/team/headless_opencode.go` invokes `opencode run --model <X> "<prompt>"` and reads stdout line-by-line with `bufio.Scanner`. opencode 1.14.x's **default** `run` output is TUI-styled with ANSI escapes; tool-execution traces land on stderr.

### Symptoms
- Simple text-only replies (e.g. "introduce yourself") work — opencode happens to emit the response on stdout
[...]

PR #357, opened 2026-04-27
fix(team): env-tunable opencode/codex turn timeouts

FranDias:
## Summary
- Adds `WUPHF_TURN_TIMEOUT`, `WUPHF_OFFICE_LAUNCH_TIMEOUT`, `WUPHF_WORKTREE_TIMEOUT` env overrides for the three hardcoded headless-turn timeouts in `internal/team/headless_codex.go`.
- Defaults preserved (4m / 10m / 12m). The reader falls back when the var is unset, empty, non-positive, or unparseable.
- `var` shape preserved so existing test mutation paths (`headlessCodexTurnTimeout = 5 * time.Second` in tests) keep working.

Addresses **Finding 2** of #313. Operators on slow providers (OpenRouter pooled queues, Kimi via Venice, tool-heavy turns) currently hit a 4-minute SIGKILL with no recourse — this gives them an escape hatch while Findings 1 and 3 are in flight.

## Test plan
- [x] `go test ./internal/team/ -run TestHeadlessCodexTurnTimeoutEnv` (new table test covers unset / empty / whitespace / garbage / zero / negative / valid values)
- [x] `go test ./internal/team/ -run "TestHeadlessCodexTurnTimeoutFor|TestRecoverTimedOut"` (existing timeout tests unchanged)
- [x] go vet, gofmt, golangci-lint via lefthook pre-commit

## Summary by CodeRabbit

* **Chores**
  * Introduced environment variable support for timeout configuration, enabling operational flexibility to customize timeout values without code modifications.

* **Tests**
  * Added comprehensive test coverage validating timeout configuration parsing behavior, including fallback handling for missing, invalid, and edge-case inputs.

### DIFF
--- a/internal/team/headless_codex.go
+++ b/internal/team/headless_codex.go
@@ -71,10 +71,30 @@ func headlessCodexRunTurn(l *Launcher, ctx context.Context, slug, notification s
 	return defaultHeadlessCodexRunTurn(l, ctx, slug, notification, channel...)
 }
 
+// headlessCodexTurnTimeoutEnv reads a duration from env, falling back to the
+// supplied default when the var is unset, empty, non-positive, or unparseable.
+// Accepts any input time.ParseDuration accepts (e.g. "6m", "90s", "1h").
+//
+// The defaults below are intentionally tight (4m / 10m / 12m); operators
+// running slower providers (OpenRouter pooled queues, Kimi via Venice) or
+// tool-heavy turns may need to extend them. See
+// https://github.com/nex-crm/wuphf/issues/313.
+func headlessCodexTurnTimeoutEnv(name string, fallback time.Duration) time.Duration {
+	raw := strings.TrimSpace(os.Getenv(name))
+	if raw == "" {
+		return fallback
+	}
+	d, err := time.ParseDuration(raw)
+	if err != nil || d <= 0 {
+		return fallback
+	}
+	return d
+}
+
 var (
-	headlessCodexTurnTimeout              = 4 * time.Minute
-	headlessCodexOfficeLaunchTurnTimeout  = 10 * time.Minute
-	headlessCodexLocalWorktreeTurnTimeout = 12 * time.Minute
+	headlessCodexTurnTimeout              = headlessCodexTurnTimeoutEnv("WUPHF_TURN_TIMEOUT", 4*time.Minute)
+	headlessCodexOfficeLaunchTurnTimeout  = headlessCodexTurnTimeoutEnv("WUPHF_OFFICE_LAUNCH_TIMEOUT", 10*time.Minute)
+	headlessCodexLocalWorktreeTurnTimeout = headlessCodexTurnTimeoutEnv("WUPHF_WORKTREE_TIMEOUT", 12*time.Minute)
 	headlessCodexStaleCancelAfter         = 90 * time.Second
 	// Minimum age an active turn must have before an enqueue can preempt
 	// it via stale-cancel. Floors out tight re-enqueue loops where two
--- a/internal/team/headless_codex_test.go
+++ b/internal/team/headless_codex_test.go
@@ -2264,3 +2264,36 @@ func TestIsCodexAuthError(t *testing.T) {
 		}
 	}
 }
+
+func TestHeadlessCodexTurnTimeoutEnv(t *testing.T) {
+	const name = "WUPHF_TURN_TIMEOUT_TEST_ONLY"
+	fallback := 4 * time.Minute
+	cases := []struct {
+		label string
+		set   bool
+		val   string
+		want  time.Duration
+	}{
+		{label: "unset falls back", set: false, want: fallback},
+		{label: "empty falls back", set: true, val: "", want: fallback},
+		{label: "whitespace falls back", set: true, val: "   ", want: fallback},
+		{label: "garbage falls back", set: true, val: "not-a-duration", want: fallback},
+		{label: "zero falls back", set: true, val: "0s", want: fallback},
+		{label: "negative falls back", set: true, val: "-30s", want: fallback},
+		{label: "minutes parsed", set: true, val: "6m", want: 6 * time.Minute},
+		{label: "seconds parsed", set: true, val: "90s", want: 90 * time.Second},
+		{label: "hours parsed", set: true, val: "1h", want: time.Hour},
+	}
+	for _, c := range cases {
+		t.Run(c.label, func(t *testing.T) {
+			if c.set {
+				t.Setenv(name, c.val)
+			} else {
+				os.Unsetenv(name)
+			}
+			if got := headlessCodexTurnTimeoutEnv(name, fallback); got != c.want {
+				t.Fatalf("headlessCodexTurnTimeoutEnv(%q=%q) = %s, want %s", name, c.val, got, c.want)
+			}
+		})
+	}
+}


internal/team/headless_codex.go
```diff
@@ -71,10 +71,30 @@ func headlessCodexRunTurn(l *Launcher, ctx context.Context, slug, notification s
...
+		return fallback
+	}
+	return d
+}
+
 var (
-	headlessCodexTurnTimeout              = 4 * time.Minute
-	headlessCodexOfficeLaunchTurnTimeout  = 10 * time.Minute
-	headlessCodexLocalWorktreeTurnTimeout = 12 * time.Minute
+	headlessCodexTurnTimeout              = headlessCodexTurnTimeoutEnv("WUPHF_TURN_TIMEOUT", 4*time.Minute)
```
FranDias, line 95: **SUGGESTION** — `WUPHF_TURN_TIMEOUT` reads as if it controls every turn, but it only sets the *default* path (non-office, non-local-worktree) — the other two siblings (`WUPHF_OFFICE_LAUNCH_TIMEOUT`, `WUPHF_WORKTREE_TIMEOUT`) explicitly carve out the more expensive paths. An operator who hits a timeout on a `local_worktree` task and reaches for `WUPHF_TURN_TIMEOUT=30m` will be surprised when nothing changes. See `headlessCodexTurnTimeoutForTurn` at line 1104 for the dispatch.

Consider renaming to something more specific like `WUPHF_DEFAULT_TURN_TIMEOUT` or `WUPHF_HEADLESS_TURN_TIMEOUT` so the three vars read as a coherent triple. Cheaper to fix now than after operators document the current name in their runbooks.

internal/team/headless_codex.go
```diff
@@ -71,10 +71,30 @@ func headlessCodexRunTurn(l *Launcher, ctx context.Context, slug, notification s
...
+// tool-heavy turns may need to extend them. See
+// https://github.com/nex-crm/wuphf/issues/313.
+func headlessCodexTurnTimeoutEnv(name string, fallback time.Duration) time.Duration {
+	raw := strings.TrimSpace(os.Getenv(name))
+	if raw == "" {
+		return fallback
+	}
+	d, err := time.ParseDuration(raw)
+	if err != nil || d <= 0 {
+		return fallback
```
FranDias, line 89: **IMPORTANT** — Silent fallback on unparseable / non-positive input is operator-hostile. The most likely operator mistake here is `WUPHF_TURN_TIMEOUT=600` (intending 10 minutes; `time.ParseDuration` requires a unit and rejects bare integers). With this code the var is silently ignored and they get the 4m default — exactly the SIGKILL they were trying to avoid, and they get no signal that their config didn't take.

At minimum log a single warning to stderr when a non-empty value fails to parse or is non-positive, e.g.:

```go
d, err := time.ParseDuration(raw)
if err != nil || d <= 0 {
    fmt.Fprintf(os.Stderr, "wuphf: ignoring invalid %s=%q (using %s)\n", name, raw, fallback)
    return fallback
}
```

The unset/empty/whitespace branches should stay silent — those are not misconfigurations.

internal/team/headless_codex.go
```diff
@@ -71,10 +71,30 @@ func headlessCodexRunTurn(l *Launcher, ctx context.Context, slug, notification s
...
 }
 
+// headlessCodexTurnTimeoutEnv reads a duration from env, falling back to the
+// supplied default when the var is unset, empty, non-positive, or unparseable.
+// Accepts any input time.ParseDuration accepts (e.g. "6m", "90s", "1h").
+//
+// The defaults below are intentionally tight (4m / 10m / 12m); operators
+// running slower providers (OpenRouter pooled queues, Kimi via Venice) or
+// tool-heavy turns may need to extend them. See
+// https://github.com/nex-crm/wuphf/issues/313.
```
FranDias, line 81: **SUGGESTION** — Nit: this paragraph ("The defaults below are intentionally tight…") documents the var block, not this function. The doc is currently anchored to `headlessCodexTurnTimeoutEnv`, so if the var block is ever moved or the function is reused elsewhere the comment becomes orphaned. Move the "defaults below" prose onto the `var (` block (where the 4m/10m/12m actually live) and keep this comment focused on the parsing contract.

internal/team/headless_codex.go
```diff
@@ -71,10 +71,30 @@ func headlessCodexRunTurn(l *Launcher, ctx context.Context, slug, notification s
...
+	return d
+}
+
 var (
-	headlessCodexTurnTimeout              = 4 * time.Minute
-	headlessCodexOfficeLaunchTurnTimeout  = 10 * time.Minute
-	headlessCodexLocalWorktreeTurnTimeout = 12 * time.Minute
+	headlessCodexTurnTimeout              = headlessCodexTurnTimeoutEnv("WUPHF_TURN_TIMEOUT", 4*time.Minute)
+	headlessCodexOfficeLaunchTurnTimeout  = headlessCodexTurnTimeoutEnv("WUPHF_OFFICE_LAUNCH_TIMEOUT", 10*time.Minute)
+	headlessCodexLocalWorktreeTurnTimeout = headlessCodexTurnTimeoutEnv("WUPHF_WORKTREE_TIMEOUT", 12*time.Minute)
```
FranDias, line 97: **SUGGESTION** — Out of the diff hunk but related: `headlessCodexStaleCancelAfter` (90s, line 98) is the sibling knob that decides when an in-flight turn becomes preemptible by a new enqueue. An operator who bumps `WUPHF_TURN_TIMEOUT=30m` to escape SIGKILLs will find that turns are still cancellable after 90s of inactivity, which can re-create the symptom they were debugging from a different angle. Two options: (a) add `WUPHF_STALE_CANCEL_AFTER` in this same PR for symmetry, or (b) leave a comment in the var block calling out that bumping the timeout doesn't extend stale-cancel and pointing to issue #313.

internal/team/headless_codex_test.go
```diff
@@ -2264,3 +2264,36 @@ func TestIsCodexAuthError(t *testing.T) {
 		}
 	}
 }
+
+func TestHeadlessCodexTurnTimeoutEnv(t *testing.T) {
+	const name = "WUPHF_TURN_TIMEOUT_TEST_ONLY"
+	fallback := 4 * time.Minute
```
FranDias, line 2270: **SUGGESTION** — The new test exercises the helper in isolation against a synthetic env var (`WUPHF_TURN_TIMEOUT_TEST_ONLY`), but there's nothing asserting that the three production env vars (`WUPHF_TURN_TIMEOUT`, `WUPHF_OFFICE_LAUNCH_TIMEOUT`, `WUPHF_WORKTREE_TIMEOUT`) are wired to the right package vars with the right defaults. A typo in either column at the call sites on lines 95–97 would slip through.

A cheap end-to-end check: a `TestHeadlessCodexTurnTimeoutDefaults` that asserts the package-level vars equal `4*time.Minute`, `10*time.Minute`, `12*time.Minute` when the test environment doesn't set those names. Won't catch env-set-during-CI, but it's the kind of regression the original PR is meant to prevent.

internal/team/headless_codex_test.go
```diff
@@ -2264,3 +2264,36 @@ func TestIsCodexAuthError(t *testing.T) {
...
+		{label: "minutes parsed", set: true, val: "6m", want: 6 * time.Minute},
+		{label: "seconds parsed", set: true, val: "90s", want: 90 * time.Second},
+		{label: "hours parsed", set: true, val: "1h", want: time.Hour},
+	}
+	for _, c := range cases {
+		t.Run(c.label, func(t *testing.T) {
+			if c.set {
+				t.Setenv(name, c.val)
+			} else {
+				os.Unsetenv(name)
```
FranDias, line 2292: **SUGGESTION** — Nit: the `unset falls back` case calls `os.Unsetenv(name)` directly while every other case uses `t.Setenv` (which auto-restores at the end of the subtest). It works because `name` is a `_TEST_ONLY` var that nothing else touches, but if anyone ever sets that name in their dev shell the test will leak the unset. Cleaner:

```go
if c.set {
    t.Setenv(name, c.val)
} else {
    t.Setenv(name, "sentinel") // ensures any pre-existing value is restored
    os.Unsetenv(name)
}
```

Not blocking — current code is fine in practice.

FranDias (commented):
Reviewed as part of the issue #313 trio (Finding 2: env-tunable timeouts).

**What I verified**
- Defaults preserved exactly: 4m / 10m / 12m, including operator precedence.
- Helper correctly falls back on unset, empty, whitespace, non-positive, and unparseable input — covered by the new table-driven test.
- `time.ParseDuration` accepts the documented `6m` / `90s` / `1h` forms.
- `var` shape preserved so existing tests at `headless_codex_test.go:690` etc. that mutate the package vars still compile and run.
- Init-time evaluation is fine: `headlessCodexTurnTimeoutEnv` is a free function, so Go's package var init order resolves it before the var block runs. No `init()` ordering hazard.
- New test passes locally on the worktree (`go test -run TestHeadlessCodexTurnTimeoutEnv ./internal/team/`).
- `WUPHF_*` naming convention is consistent with the rest of the codebase (`WUPHF_HEADLESS_PROVIDER`, `WUPHF_BROKER_TOKEN`, `WUPHF_MEMORY_BACKEND`, etc.).

**Worth thinking about before merge** (inline notes on each)
1. **`WUPHF_TURN_TIMEOUT` naming is too generic** — it only controls the default (non-office, non-worktree) path. Operators hitting a worktree timeout will reach for it and be surprised. Consider `WUPHF_DEFAULT_TURN_TIMEOUT` or `WUPHF_HEADLESS_TURN_TIMEOUT`. Cheap to fix now, painful after this ships.
2. **Silent fallback on parse failure is operator-hostile** (medium). The most likely operator typo is `WUPHF_TURN_TIMEOUT=600` (no unit) — `time.ParseDuration` rejects it, the SIGKILL behavior they were debugging persists, and they get no signal that their env var is being ignored. A one-line stderr warning when a *non-empty* value fails to parse / is non-positive would dramatically improve operator experience. Keep the unset/empty/whitespace branches silent.
3. **Stale-cancel knob** — `headlessCodexStaleCancelAfter` (90s) is the sibling that decides when an in-flight turn becomes preemptible. An operator who bumps `WUPHF_TURN_TIMEOUT=30m` to escape SIGKILLs may still see turns cancelled after 90s of inactivity. Either add `WUPHF_STALE_CANCEL_AFTER` here for symmetry, or drop a comment in the var block flagging the relationship.
4. **Doc placement nit** — the "defaults below are intentionally tight" paragraph documents the var block but is anchored to the helper function. Move it onto the `var (` block.
5. **End-to-end coverage gap** — the test only exercises the helper with a synthetic name (`WUPHF_TURN_TIMEOUT_TEST_ONLY`). Nothing asserts the three production wiring lines (95–97) point the right names at the right vars with the right defaults. A 6-line `TestHeadlessCodexTurnTimeoutDefaults` would close that.
6. **Test cleanliness nit** — the "unset" subtest uses `os.Unsetenv` directly instead of `t.Setenv` + `os.Unsetenv`, so it could leak the unset if a dev shell happened to have the test-only var set. Not a real bug.

None of these are merge blockers — the change is small, defaults are preserved, and the helper is tight. Items 1 and 2 are the ones I'd want resolved (or explicitly punted) before this lands.

PR #358, opened 2026-04-27
fix(team): surface opencode base config parse errors in agent log

FranDias:
## Summary
- `writeHeadlessOpencodeMCPConfig` silently swallowed `json.Unmarshal` errors on `~/.config/opencode/opencode.json`. A single trailing comma caused the merge to start from an empty map — dropping the user's `model`/`provider` blocks from every per-agent config without a trace.
- Keep the best-effort write so wuphf still boots, but log the parse failure (path + specific JSON error) via `appendHeadlessCodexLog`, so the operator has a thread to pull on.

Addresses **Bonus #1** of #313.

## Test plan
- [x] `go test ./internal/team/ -run TestWriteHeadlessOpencodeMCPConfig` (existing concurrent test + new parse-failure test, both pass)
- [x] go vet, gofmt, golangci-lint via lefthook pre-commit

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for configuration file parsing. When a configuration file fails to parse, diagnostic error logs are now generated with failure details, replacing the previous silent handling behavior.

### DIFF
--- a/internal/team/headless_opencode.go
+++ b/internal/team/headless_opencode.go
@@ -316,9 +316,15 @@ func (l *Launcher) writeHeadlessOpencodeMCPConfig(slug string) (string, error) {
 
 	merged := map[string]any{}
 	if raw, err := os.ReadFile(baseConfigPath); err == nil && len(raw) > 0 {
-		// Best-effort: if the existing file isn't valid JSON, overwrite it
-		// rather than silently losing the WUPHF overlay.
-		_ = json.Unmarshal(raw, &merged)
+		// Best-effort: if the existing file isn't valid JSON, fall back to
+		// writing a minimal overlay so wuphf keeps booting — but surface the
+		// parse error in the agent log so the operator can see they have a
+		// malformed base config silently dropping their `model`/`provider`
+		// blocks from every per-agent merge. (#313 bonus #1)
+		if uerr := json.Unmarshal(raw, &merged); uerr != nil {
+			merged = map[string]any{}
+			appendHeadlessCodexLog(slug, fmt.Sprintf("opencode_base-config-parse-failed: %s: %s — per-agent config will not inherit user model/provider/MCP keys until this is fixed", baseConfigPath, uerr.Error()))
+		}
 	}
 
 	mcp, _ := merged["mcp"].(map[string]any)
--- a/internal/team/headless_opencode_test.go
+++ b/internal/team/headless_opencode_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"testing"
 )
@@ -104,5 +105,49 @@ func TestWriteHeadlessOpencodeMCPConfigConcurrent(t *testing.T) {
 		if env["WUPHF_AGENT_SLUG"] != slug {
 			t.Fatalf("%s config has WUPHF_AGENT_SLUG=%#v", slug, env["WUPHF_AGENT_SLUG"])
 		}
+	}
+}
+
+// TestWriteHeadlessOpencodeMCPConfigLogsBaseConfigParseFailure verifies that a
+// malformed base ~/.config/opencode/opencode.json (e.g. trailing comma) causes
+// writeHeadlessOpencodeMCPConfig to surface the parse error via the agent
+// log instead of silently dropping the user's `model`/`provider` blocks. (#313)
+func TestWriteHeadlessOpencodeMCPConfigLogsBaseConfigParseFailure(t *testing.T) {
+	home := t.TempDir()
+	t.Setenv("HOME", home)
+	logDir := t.TempDir()
+	t.Setenv("WUPHF_LOG_DIR", logDir)
+
+	configDir := filepath.Join(home, ".config", "opencode")
+	if err := os.MkdirAll(configDir, 0o755); err != nil {
+		t.Fatal(err)
+	}
+	// Trailing comma — valid Opencode-side JSON5 in some hand-edited configs,
+	// but encoding/json rejects it. The legacy code silently swallowed this.
+	malformed := `{"theme":"dark","model":"openrouter/foo",}`
+	configPath := filepath.Join(configDir, "opencode.json")
+	if err := os.WriteFile(configPath, []byte(malformed), 0o600); err != nil {
+		t.Fatal(err)
+	}
+
+	orig := headlessOpencodeExecutablePath
+	headlessOpencodeExecutablePath = func() (string, error) { return "/usr/local/bin/wuphf", nil }
+	defer func() { headlessOpencodeExecutablePath = orig }()
+
+	l := &Launcher{}
+	if _, err := l.writeHeadlessOpencodeMCPConfig("ceo"); err != nil {
+		t.Fatalf("writeHeadlessOpencodeMCPConfig: %v", err)
+	}
+
+	logBytes, err := os.ReadFile(filepath.Join(logDir, "headless-codex-ceo.log"))
+	if err != nil {
+		t.Fatalf("read agent log: %v", err)
+	}
+	logStr := string(logBytes)
+	if !strings.Contains(logStr, "opencode_base-config-parse-failed") {
+		t.Fatalf("expected opencode_base-config-parse-failed in agent log, got:\n%s", logStr)
+	}
+	if !strings.Contains(logStr, configPath) {
+		t.Fatalf("expected base config path %q in agent log, got:\n%s", configPath, logStr)
 	}
 }


internal/team/headless_opencode.go
```diff
@@ -316,9 +316,15 @@ func (l *Launcher) writeHeadlessOpencodeMCPConfig(slug string) (string, error) {
...
-		// rather than silently losing the WUPHF overlay.
-		_ = json.Unmarshal(raw, &merged)
+		// Best-effort: if the existing file isn't valid JSON, fall back to
+		// writing a minimal overlay so wuphf keeps booting — but surface the
+		// parse error in the agent log so the operator can see they have a
+		// malformed base config silently dropping their `model`/`provider`
+		// blocks from every per-agent merge. (#313 bonus #1)
+		if uerr := json.Unmarshal(raw, &merged); uerr != nil {
+			merged = map[string]any{}
+			appendHeadlessCodexLog(slug, fmt.Sprintf("opencode_base-config-parse-failed: %s: %s — per-agent config will not inherit user model/provider/MCP keys until this is fixed", baseConfigPath, uerr.Error()))
```
FranDias, line 326: **IMPORTANT** — The behavior here writes a per-agent config to `OPENCODE_CONFIG` regardless of parse failure, which means opencode reads *our* fallback file (with no `model`/`provider` from the user) rather than the user's original malformed one. Net effect on a trailing-comma `opencode.json`: the agent silently runs against the default model instead of the user's `openrouter/foo` (or Ollama, or whatever they configured), and the only signal is one line buried in `~/.wuphf/logs/headless-codex-<slug>.log`.

Worth considering whether this also wants a `broker.PostSystemMessage(...)` like the auth-error path on line 202 — operators looking at the chat UI today won't see the warning at all. Also `wuphfLogDir()` can return `""` (HOME unset, `WUPHF_LOG_DIR` unwritable), in which case `appendHeadlessCodexLog` no-ops silently and we're back to the original silent-overwrite behavior with no thread to pull on. Not a blocker for this PR but the log-only signal is a thin wire.

internal/team/headless_opencode.go
```diff
@@ -316,9 +316,15 @@ func (l *Launcher) writeHeadlessOpencodeMCPConfig(slug string) (string, error) {
...
-		// Best-effort: if the existing file isn't valid JSON, overwrite it
-		// rather than silently losing the WUPHF overlay.
-		_ = json.Unmarshal(raw, &merged)
+		// Best-effort: if the existing file isn't valid JSON, fall back to
+		// writing a minimal overlay so wuphf keeps booting — but surface the
+		// parse error in the agent log so the operator can see they have a
+		// malformed base config silently dropping their `model`/`provider`
+		// blocks from every per-agent merge. (#313 bonus #1)
+		if uerr := json.Unmarshal(raw, &merged); uerr != nil {
+			merged = map[string]any{}
```
FranDias, line 325: **SUGGESTION** — Resetting `merged = map[string]any{}` after a parse error is the right move and worth keeping. `encoding/json` is documented to populate destination fields up to the point of error before returning, so without this reset a partial parse (e.g. malformed deep in the file but valid up to a top-level `theme` or `mcp` key) would leak half-parsed user state into the overlay and write a Frankenstein config that's neither the user's nor a clean fallback. This PR cleanly avoids that — please keep this line even if you change anything else here.

internal/team/headless_opencode.go
FranDias, line 326: **SUGGESTION** — The log line is descriptive but doesn't tell the operator what to do. The path (`%s`) is great, but the JSON error from `encoding/json` is opaque about JSON5-style mistakes (`invalid character ',' looking for beginning of object key string` for the trailing-comma case is technically correct but won't click for a hand-edit user). Consider appending a one-line hint, e.g.:

```
opencode_base-config-parse-failed: %s: %s — per-agent config will not inherit user model/provider/MCP keys until this is fixed (try `jq . %s` to validate; common cause: trailing comma)
```

The slug isn't strictly needed in the line content (the file is already `headless-codex-<slug>.log`) but if anyone ever pivots `appendHeadlessCodexLog` to write to a shared log this becomes invisible. Mild — leave it.

internal/team/headless_opencode_test.go
```diff
@@ -106,3 +107,47 @@ func TestWriteHeadlessOpencodeMCPConfigConcurrent(t *testing.T) {
...
+	}
+
+	orig := headlessOpencodeExecutablePath
+	headlessOpencodeExecutablePath = func() (string, error) { return "/usr/local/bin/wuphf", nil }
+	defer func() { headlessOpencodeExecutablePath = orig }()
+
+	l := &Launcher{}
+	if _, err := l.writeHeadlessOpencodeMCPConfig("ceo"); err != nil {
+		t.Fatalf("writeHeadlessOpencodeMCPConfig: %v", err)
+	}
```
FranDias, line 140: **IMPORTANT** — The assertions only verify *that the warning was logged*, not that the rest of the function did the right thing on the failure path. Three gaps worth closing while the test fixture is here:

1. **Best-effort write actually happened.** `_, err := l.writeHeadlessOpencodeMCPConfig("ceo")` discards the returned path. Capture it and assert the per-agent file exists, parses as valid JSON, and contains `mcp.wuphf-office.environment.WUPHF_AGENT_SLUG == "ceo"`. That's the documented contract this PR is preserving ("writes a minimal overlay so wuphf keeps booting"); without that assertion, a future change that silently bails on parse error would still pass this test.
2. **The data-loss failure mode is documented in the assertion.** The malformed seed has `"model":"openrouter/foo"` — assert that key is *absent* from the per-agent output. That's the bug the warning is complaining about; encoding it in a test prevents drift if someone later "fixes" the merge to preserve partial state.
3. **Underlying JSON error is included.** `strings.Contains(logStr, "invalid character")` (or whatever the trailing-comma error shape is) — proves the operator gets the parser detail, not just a generic `parse-failed`.

A negative-case sibling test (valid JSON ⇒ no warning logged) is also cheap and would catch a regression where someone wraps the call in `if err != nil { ... }` but inverts the condition.

internal/team/headless_opencode_test.go
```diff
@@ -106,3 +107,47 @@ func TestWriteHeadlessOpencodeMCPConfigConcurrent(t *testing.T) {
...
+
+// TestWriteHeadlessOpencodeMCPConfigLogsBaseConfigParseFailure verifies that a
+// malformed base ~/.config/opencode/opencode.json (e.g. trailing comma) causes
+// writeHeadlessOpencodeMCPConfig to surface the parse error via the agent
+// log instead of silently dropping the user's `model`/`provider` blocks. (#313)
+func TestWriteHeadlessOpencodeMCPConfigLogsBaseConfigParseFailure(t *testing.T) {
+	home := t.TempDir()
+	t.Setenv("HOME", home)
+	logDir := t.TempDir()
+	t.Setenv("WUPHF_LOG_DIR", logDir)
```
FranDias, line 119: **SUGGESTION** — `t.Setenv("WUPHF_LOG_DIR", logDir)` overrides the package-wide `WUPHF_LOG_DIR` that `TestMain` sets to a leaked dir specifically to absorb writes from goroutines that outlive their parent test (see `broker_test.go:54-74`). For the trivial `Launcher{}` here that's safe — there are no leaked workers — but it's worth a one-line comment so the next person who copies this pattern knows why this test gets to escape the package-wide hatch. If anyone later turns this into `t.Parallel()` or extends it to spawn a turn, leaked writes from concurrent tests using slug `"ceo"` (broker_actions, broker_entity, broker_review, etc.) would land in `logDir` and the `strings.Contains` assertion would still pass on noise. Tightening to `strings.Count(logStr, "opencode_base-config-parse-failed") == 1` would make that risk explicit.

FranDias (commented):
Tight, well-scoped fix for #313 bonus #1. The core change (replace `_ = json.Unmarshal(...)` with a parse-error-aware branch that resets `merged` and emits an agent-log line) is correct, and the `merged = map[string]any{}` reset deserves explicit credit — it closes a partial-parse leak that the previous silent-swallow code happened to dodge only because the destination started empty.

A few things to consider before merge — none blocking:

- **The "log + best-effort write" tradeoff is reasonable but the signal is thinner than it looks.** The per-agent file we write becomes `OPENCODE_CONFIG`, so opencode reads our fallback (no `model`/`provider` from the user) rather than the user's malformed source. Net effect: agent silently runs against the default model and the only signal lives in `~/.wuphf/logs/headless-codex-<slug>.log`. `wuphfLogDir()` can also degrade to `""` (HOME unset, override unwritable), in which case `appendHeadlessCodexLog` no-ops and we're back to fully silent. Worth thinking about whether this also wants a `broker.PostSystemMessage(...)` like the auth-error path on line 202 — operators in the chat UI today see nothing.
- **Log line copy is fine but not actionable.** The format is consistent with the other `opencode_*` events in this file. The `encoding/json` error message itself is opaque for hand-edit users (`invalid character ',' looking for beginning of object key string` won't click for a trailing-comma JSON5 mistake). A short hint like "(try `jq . <path>` to validate; common cause: trailing comma)" would close the gap without bloat.
- **Test asserts the warning fires but not the rest of the contract.** Three reachable gaps: (a) the returned per-agent path isn't checked, so a future change that bails early on parse failure would still pass; (b) the malformed seed has `"model":"openrouter/foo"` but the test never asserts that key is *absent* from the output (which is the bug the warning is about); (c) the underlying JSON error string isn't asserted, so a future refactor that drops `uerr.Error()` from the format string would silently weaken the operator-facing detail. A negative-case sibling (valid JSON ⇒ no warning) is also cheap. Inline comment has the details.
- **`t.Setenv("WUPHF_LOG_DIR", ...)` escapes the package-wide TestMain hatch** that exists to absorb writes from goroutines that outlive their parent test. Safe here because the test uses a bare `Launcher{}` with no broker/workers, but worth a one-line comment so the pattern doesn't get copied into a test that does spawn workers. The slug `"ceo"` is also used heavily by other tests in the package — `strings.Count(logStr, "opencode_base-config-parse-failed") == 1` would make the assertion robust against future leaks.

No correctness or security issues. Verified the prefix style matches existing `opencode_*` log conventions and that the per-slug log filename makes the slug implicit in the line content. Happy to see this land once the test gaps are closed.

PR #359, opened 2026-04-27
fix(team): parse opencode JSON event stream so tool turns are visible

FranDias:
## Summary
- `opencode run` (1.14.x default) emits TUI-styled stdout with ANSI escapes; tool execution traces land on stderr. wuphf's headless turn used `bufio.Scanner` against raw stdout, so any tool-using turn produced `first_event_ms=-1`, `first_text_ms=-1`, `final_chars=0` — wuphf was blind to real work, then SIGKILL'd at the 4-minute timeout.

Two pieces:

1. **`provider.ReadOpencodeJSONStream`** mirrors `ReadClaudeJSONStream`, consuming `opencode run --format json` and emitting normalized `text` / `tool_use` / `tool_result` / `error` events. Lines that don't decode as JSON (banners, older opencode builds, malformed events) fall through as plain-text events so we never silently drop output. Unknown event types are forwarded raw so opencode minor-version drift doesn't quietly suppress signal.
2. **`runHeadlessOpencodeTurn`** now requests `--format json`, swaps its scanner for the new reader, populates `first_tool_ms` (previously always -1), and pushes structured `[tool] ...` / `[tool_result] ...` / `[error] ...` lines to the agent activity stream so the UI reflects what opencode is actually doing.

Addresses **Finding 1** of #313.

## What this does NOT do
- Does not bin-pin opencode. The reader is intentionally lenient about the JSON schema and falls through to plain-text on unknown events; we can revisit pinning if a future opencode minor breaks the contract.
- Does not add `--variant minimal`. That's a model-quality tuning per the issue's appendix; operators can keep using the workaround shim if they need it.
- Does not address Finding 3 (broker should reject `team_task complete` without a deliverable). Per @FranDias's comment on the issue, that's blocked on result-schema design.

## Test plan
- [x] `go test ./internal/provider/ -run TestReadOpencodeJSONStream` — 6 new table tests cover text aggregation, tool_use/tool_result, error capture, plain-text fallback, unknown event types, malformed JSON
- [x] `go test ./internal/team/ -run "TestBuildHeadlessOpencode|TestWriteHeadlessOpencode"` — new arg-builder tests + existing concurrent-config test
- [x] go vet, gofmt, golangci-lint via lefthook pre-commit (clean)
- [ ] Manual: full opencode tool turn against an org markdown ingest task — needs an opencode binary + provider creds, not in my local env

## Summary by CodeRabbit

* **New Features**
  * Enhanced OpenCode integration with structured JSON event streaming for improved reliability in processing command output.

* **Bug Fixes**
  * Improved handling of tool calls and results with better event parsing and error tracking.


internal/provider/opencode_stream.go
```diff
@@ -0,0 +1,145 @@
...
+		// Cheap JSON-object detector before we pay for json.Unmarshal —
+		// avoids decoding banners or stray plain-text lines.
+		appendPlain := func() {
+			if text.Len() > 0 {
+				text.WriteByte('\n')
+			}
+			text.WriteString(line)
+			emit(OpencodeStreamEvent{Type: "text", Text: line, Detail: line, Raw: line})
+		}
+
```
FranDias, line 92: **IMPORTANT** — `appendPlain` does `text.WriteByte('\n')` before writing a plain line, but the JSON `text` case below at line 113 writes `chunk` directly with no separator. If the stream is `banner\n{"type":"text","part":{"text":"ack"}}`, FinalMessage becomes `"opencode v1.14.25 — runningack"` — the banner and the first JSON text chunk get concatenated with no boundary. The existing `TestReadOpencodeJSONStreamFallsBackToPlainTextOnNonJSON` test only asserts both substrings exist, so this slips through.

Not a critical correctness bug for tool-using turns (the banner is usually short context noise), but it does show up in `final_chars` and in the agent's reply when the runner posts `streamRes.FinalMessage`. The symmetric fix is to also boundary the JSON text branch when transitioning from plain → JSON.

**Fix:** track a `prevWasPlain bool` (or just always seed `\n` when `text.Len() > 0` and the previous write was plain) and prepend `\n` in the `text` case when the prior write came from `appendPlain`. Or simpler: have `appendPlain` always append a trailing `\n` after the line, and trim the final result.

internal/provider/opencode_stream.go
```diff
@@ -0,0 +1,145 @@
...
+			text.WriteString(chunk)
+			emit(OpencodeStreamEvent{Type: "text", Text: chunk, Detail: chunk, Raw: line})
+		case "tool_use", "tool_call":
+			name := firstNonEmpty(ev.Part.ToolName, ev.Part.Tool, ev.Part.Name)
+			id := firstNonEmpty(ev.Part.CallID, ev.Part.ID)
+			emit(OpencodeStreamEvent{Type: "tool_use", ToolName: name, ToolID: id, Detail: name, Raw: line})
+		case "tool_result":
+			detail := strings.TrimSpace(string(ev.Part.Output))
+			if detail == "" {
+				detail = strings.TrimSpace(string(ev.Part.Result))
```
FranDias, line 121: **IMPORTANT** — `tool_result` Detail is the literal raw JSON of `ev.Part.Output`. If opencode emits `"output":"wrote 1 file"`, `string(ev.Part.Output)` is `"wrote 1 file"` *with the surrounding quotes*. So the team-side runner pushes `[tool_result] "wrote 1 file"` to the agent stream — visible quotes around every string output. If `output` is an object, you get JSON-encoded `{"foo":"bar"}`. The test happens to assert `Contains(detail, "wrote 1 file")` which passes through the quotes.

**Fix:** unmarshal the RawMessage into `interface{}` and format strings without quotes:

```go
var v interface{}
if err := json.Unmarshal(ev.Part.Output, &v); err == nil {
    if s, ok := v.(string); ok {
        detail = s
    } else {
        // keep the JSON-encoded form for non-string outputs
        detail = strings.TrimSpace(string(ev.Part.Output))
    }
}
```

NB: claude_stream.go solves the same problem in `formatClaudeToolResult` — worth lifting that into a shared helper.

internal/provider/opencode_stream.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+			id := firstNonEmpty(ev.Part.CallID, ev.Part.ID)
+			emit(OpencodeStreamEvent{Type: "tool_result", ToolID: id, Text: detail, Detail: detail, Raw: line})
+		case "step_start", "step_finish":
+			emit(OpencodeStreamEvent{Type: ev.Type, Raw: line})
+		case "error":
+			msg := firstNonEmpty(ev.Message, ev.Error, ev.Part.Message)
+			if msg != "" {
+				result.LastError = msg
```
FranDias, line 130: **SUGGESTION** — `step_start` arrives before any text/tool event, so it now sets `firstEventAt` in the team-side runner (line ~146-149 of headless_opencode.go). That changes the meaning of `first_event_ms` vs the old behaviour where it was the first non-empty stdout line. For metric continuity with codex/claude (which emit `first_event_ms` on the first *content-bearing* event), consider not emitting `step_start`/`step_finish` at all — the runner ignores them anyway, so they only add noise to `first_event_ms`. If you want to keep them for future diagnostics, gate the emit on a `verbose` flag.

internal/provider/opencode_stream.go
```diff
@@ -0,0 +1,145 @@
...
+// which keeps wuphf working against older opencode builds and preserves the
+// pre-#313 behavior as a safety net.
+type OpencodeStreamEvent struct {
+	Type     string
+	Text     string
+	ToolName string
+	ToolID   string
+	Detail   string
+	Raw      string
+}
```
FranDias, line 28: **SUGGESTION** — Structural divergence from `claude_stream.go` to call out:

- Claude exposes `ToolUseID`; this struct exposes `ToolID`. Same concept, different name.
- This struct adds a `Raw` field (carrying the original JSON line); claude doesn't. Useful for diagnostics but inconsistent.
- Claude buffer is 1 MiB (`1024*1024`); this is 4 MiB. Codex is also 4 MiB so the bigger buffer is the right call here, but worth a one-liner comment explaining why opencode/codex need the bigger headroom (large tool outputs in a single line) when claude doesn't.

Not blocking, but if you ever want to share helpers (e.g., `formatToolResult` for unquoting JSON outputs), aligning the field names will make that easier. Recommend renaming `ToolID` → `ToolUseID` for parity, or noting the divergence in the type doc.

internal/provider/opencode_stream.go
```diff
@@ -0,0 +1,145 @@
...
+		trimmed := strings.TrimSpace(line)
+		if trimmed == "" {
+			continue
+		}
+
+		// Cheap JSON-object detector before we pay for json.Unmarshal —
+		// avoids decoding banners or stray plain-text lines.
+		appendPlain := func() {
+			if text.Len() > 0 {
+				text.WriteByte('\n')
```
FranDias, line 87: **SUGGESTION** — `appendPlain` is a closure declared once per scanner iteration. Trivial perf, but the bigger reason to lift it: it captures `line` and `text` from the loop scope, which makes it harder to reason about. Lift it to a method/free function on a small struct (with `text *strings.Builder`) and pass `line` explicitly. Minor; only call out if you're already going to revisit this for the separator bug above.

internal/provider/opencode_stream.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+			text.WriteString(line)
+			emit(OpencodeStreamEvent{Type: "text", Text: line, Detail: line, Raw: line})
+		}
+
+		if !strings.HasPrefix(trimmed, "{") {
+			appendPlain()
+			continue
+		}
```
FranDias, line 96: **SUGGESTION** — JSON-prefix detector: `HasPrefix(trimmed, "{")` skips lines that start with `[` (top-level JSON arrays). Opencode's current schema is `{"type":"..."}` per line, but if a future minor batches events into an array, this filter would drop them silently into the plain-text fallback. Probably overkill to handle proactively, but worth a one-line comment that this assumption is bound to opencode's per-line object framing — and consider also accepting `[`. Costs you basically nothing:

```go
if !strings.HasPrefix(trimmed, "{") && !strings.HasPrefix(trimmed, "[") {
    appendPlain()
    continue
}
```

internal/provider/opencode_stream_test.go
```diff
@@ -0,0 +1,127 @@
...
+	}, "\n")
+
+	res, err := ReadOpencodeJSONStream(strings.NewReader(stream), nil)
+	if err != nil {
+		t.Fatalf("ReadOpencodeJSONStream: %v", err)
+	}
+	if !strings.Contains(res.FinalMessage, "second") {
+		t.Fatalf("second valid text event missing: %q", res.FinalMessage)
+	}
+}
```
FranDias, line 127: **SUGGESTION** — Test coverage gaps worth filling before merge given this is the runtime opencode is being downgraded onto:

1. `tool_call` alias (line 116 in source) — no test exercises the `tool_call` arm, only `tool_use`.
2. `tool_result` with `result` field set instead of `output` (the fallback at line 124) — no test.
3. Multi-line tool output with embedded newlines (real bash tool output) — verifies the truncation/escape behavior.
4. **Backwards-compat path**: a stream that is *all* plain text (legacy opencode pre-`--format json`) — verifies FinalMessage rebuilds cleanly. The current `FallsBackToPlainTextOnNonJSON` test mixes one banner + one JSON event; a pure plain-text run is the legacy fallback case the doc-comment promises to support.
5. Banner → JSON `text` ordering — verifies the separator concern raised on line 92 actually does what you intend.
6. The `Part.Message` fallback in error events (line 130) — no test exercises that arm.

internal/team/headless_opencode.go
```diff
@@ -147,19 +133,25 @@ func (l *Launcher) runHeadlessOpencodeTurn(ctx context.Context, slug string, not
...
+		if agentStream != nil && strings.TrimSpace(line) != "" {
+			agentStream.Push(line)
+		}
+	}
 
-	scanner := bufio.NewScanner(teedStdout)
-	scanner.Buffer(make([]byte, 0, 64*1024), 4*1024*1024)
-	for scanner.Scan() {
-		line := scanner.Text()
+	streamRes, scanErr := provider.ReadOpencodeJSONStream(stdout, func(ev provider.OpencodeStreamEvent) {
```
FranDias, line 145: **IMPORTANT** — The pre-PR shape teed raw stdout to `agentStream.Push(line)`, so the web UI live-stream pane saw whatever the agent saw. Now the agent stream only receives JSON `text` chunks (line ~163), `[tool] X` markers (~174), `[tool_result] X` (~177), and `[error] X` (~182). That's *better* for tool-using turns than raw JSON, but two regressions to verify:

1. **Plain-text fallback no longer reaches the agent stream.** When a banner / non-JSON line is emitted as a `text` event with `Text: line`, it does flow into `pushStream(ev.Text)` here — good. But the old code would also push lines opencode emits as raw chatter (e.g., progress dots, ANSI noise on older builds) that the parser now silently swallows in `appendPlain` if they reach FinalMessage but aren't `text`-typed events. Verify by running an older opencode build (or just one in TUI mode) and confirming the agent stream still shows something.
2. **`step_start` and `step_finish`** events arrive at the runner with `Type: step_start` and trip the `firstEventAt` clock, but go down the `default` branch in this switch — they never reach `pushStream`. Old behaviour: every line went to agentStream, including `step_start` JSON. New: nothing for steps. Probably fine, but if the web UI used those as a heartbeat to show "opencode is alive", it now goes silent during long thinks.

**Fix:** consider pushing a `[step] start` / `[step] finish` marker, or document explicitly that these are intentionally suppressed.

internal/team/headless_opencode.go
```diff
@@ -168,19 +160,34 @@ func (l *Launcher) runHeadlessOpencodeTurn(ctx context.Context, slug string, not
...
+			}
+			l.updateHeadlessProgress(slug, "active", "tool", "running "+detail, metrics)
+			pushStream("[tool] " + detail)
+		case "tool_result":
+			if d := strings.TrimSpace(ev.Detail); d != "" {
+				pushStream("[tool_result] " + truncate(d, 240))
+			}
+		case "error":
+			if msg := strings.TrimSpace(ev.Detail); msg != "" {
+				lastError = msg
```
FranDias, line 181: **SUGGESTION** — `lastError` is captured in the closure here, then read once at line 241 for the latency log. But `streamRes.LastError` (set inside the parser at the same point) carries the same value — the closure-captured `lastError` is redundant. Prefer `strings.TrimSpace(streamRes.LastError)` at line 241 and drop the local `var lastError string` + the assignment in this branch. Single source of truth, less to reason about.

internal/team/headless_opencode.go
```diff
@@ -133,7 +120,6 @@ func (l *Launcher) runHeadlessOpencodeTurn(ctx context.Context, slug string, not
 		case <-ctx.Done():
 			terminateHeadlessProcess(cmd)
 			_ = stdout.Close()
```
FranDias, line 122: **IMPORTANT** — ctx-cancellation flow: when ctx fires, the goroutine on line 119 calls `terminateHeadlessProcess(cmd)` then `_ = stdout.Close()`. The parser is mid-read inside `ReadOpencodeJSONStream` at line 145. Closing the read end of `cmd.StdoutPipe()` causes `scanner.Scan()` to return false (or surface a close error). The parser returns, then `cmd.Wait()` runs at line 192.

Two concerns:

1. **Wait may return ctx.Err()-disguised exit error**, not `context.Canceled`, because we killed the child via SIGTERM/SIGKILL not via ctx propagation. The error path at line 192-222 logs `status=error provider=opencode` with stderr detail. That's fine, but the auth-error nag on line 204 (`isOpencodeAuthError`) could fire spuriously if the killed child wrote partial "... unauthorized" to stderr before SIGKILL. Worth a quick check: does `terminateHeadlessProcess` give the child time to flush? If yes, low risk. If it sends SIGKILL immediately, no false positives. Either way, consider gating the nag on `!ctx.Err()`.
2. **Double close on stdout** — the cancel goroutine calls `_ = stdout.Close()`. The parser also drains it. `cmd.Wait()` doc: "It will close the pipe after seeing the command exit". Calling Close() ourselves is documented as fine, but worth verifying nothing in the parser tries to re-read after Close. I think `bufio.Scanner` is safe here (returns EOF), but flag it for the next person reading this code.

internal/team/headless_opencode.go
```diff
@@ -254,11 +263,16 @@ func (l *Launcher) runHeadlessOpencodeTurn(ctx context.Context, slug string, not
...
+// [--model X] [message..]` — no --cwd, no --quiet, no stdin sentinel.
+// Working directory is set via cmd.Dir by the caller.
+//
+// `--format json` opts the headless path into opencode's JSON event stream so
+// wuphf can see tool-use / tool-result / error events that opencode otherwise
+// renders as TUI-styled stdout (or routes to stderr) and the previous
+// bufio.Scanner-based consumer was blind to. See #313 (Finding 1).
 func buildHeadlessOpencodeArgs(model string, prompt string) []string {
-	args := []string{"run"}
+	args := []string{"run", "--format", "json"}
```
FranDias, line 275: **IMPORTANT** — The PR title and issue #313 Finding 1 promise that turn output is now visible. But if the user is running an older opencode that doesn't recognize `--format json`, the binary errors out *before* emitting any JSON. The lenient parser will catch that as plain-text fallback and produce a FinalMessage that contains the error text; `cmd.Wait()` returns non-zero and we hit the error-detail path at line 196. Net: the user sees a confusing failure where the cause is just a stale opencode binary.

The issue's shim adds `--variant minimal`. This PR doesn't add it — defensible if `--format json` alone is enough on 1.14.x, but **without bin-pinning** of opencode there's no guard against the user being on an older build that pre-dates `--format`.

**Fix (cheap):** at startup, detect whether the resolved opencode supports `--format json` (e.g., `opencode run --help` parse, or version probe) and downgrade to the legacy plain-text path if not. Or document the minimum opencode version in the pre-flight log next to `headlessOpencodeLookPath`. At minimum, in `describeOpencodeFailure`-style error mapping, recognise the shape opencode emits when `--format` is unknown so the user gets an actionable message.

internal/team/headless_opencode.go
```diff
@@ -168,19 +160,34 @@ func (l *Launcher) runHeadlessOpencodeTurn(ctx context.Context, slug string, not
...
+			}
+			detail := strings.TrimSpace(ev.ToolName)
+			if detail == "" {
+				detail = "tool"
+			}
+			l.updateHeadlessProgress(slug, "active", "tool", "running "+detail, metrics)
+			pushStream("[tool] " + detail)
+		case "tool_result":
+			if d := strings.TrimSpace(ev.Detail); d != "" {
+				pushStream("[tool_result] " + truncate(d, 240))
```
FranDias, line 177: **SUGGESTION** — `truncate(d, 240)` is fine for the activity stream, but `truncate` lives in `internal/team/launcher.go:2104` (a 5-line helper that appends `...`). The 240-char limit isn't justified anywhere. For comparison:

- `headless_claude.go` uses 140 for tool_result
- `headless_codex.go` uses similar limits
- This PR introduces 240

Not a bug — just inconsistent. If the larger limit is deliberate (opencode tool outputs tend to be more verbose?), worth a one-line comment. Otherwise align to 140 for consistency with the claude path so future tweaks land in one place.

internal/team/headless_opencode.go
```diff
@@ -6,14 +6,14 @@ import (
...
 	"fmt"
-	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/nex-crm/wuphf/internal/config"
+	"github.com/nex-crm/wuphf/internal/provider"
```
FranDias, line 16: **SUGGESTION** — Stale doc-comment: the function-level docstring on `runHeadlessOpencodeTurn` (lines ~44-48, just above this hunk) still says *"Opencode emits plain text on stdout rather than structured JSONL, so this path is a thinner version of runHeadlessCodexTurn with no tool-event parsing"*. With this PR, that's no longer true — the path *does* parse structured events. Update the docstring while you're in here so the next reader doesn't get misled.

FranDias (commented):
Thorough review of the largest of the three #313 PRs. Strategy looks sound — switching opencode 1.14.x onto `--format json` and consuming structured events is the right fix for the visibility gap (`first_event_ms=-1`/`first_text_ms=-1`/`final_chars=0`/SIGKILL-at-4-min). The new `provider.ReadOpencodeJSONStream` mirrors `ReadClaudeJSONStream` cleanly enough that future provider swaps are straightforward.

That said, I left 13 inline comments — none are merge-blockers in the strict sense, but several are worth addressing before this lands on main given how invasive the runtime change is.

**Highest-priority issues to look at:**

1. `opencode_stream.go:92` — Plain-text fallback writes a `\n` separator before the line, but the JSON `text` branch writes its chunk with no separator. Stream `banner\n{"type":"text","part":{"text":"ack"}}` produces `FinalMessage = "...runningack"` (no boundary). The existing fallback test asserts substring containment, so it slips through. Symmetric fix needed.
2. `opencode_stream.go:121` — `tool_result` Detail uses raw JSON of `Part.Output`, which means string outputs surface in the agent stream with literal surrounding quotes (`[tool_result] "wrote 1 file"`). Decode the RawMessage and unwrap strings.
3. `headless_opencode.go:275` — `--format json` is added, but no version probe and no bin-pin. If the user's opencode is older than the flag, the failure is opaque. Worth a probe or a friendly "upgrade opencode" mapping in error handling. Issue #313's shim used `--variant minimal` too — defensible to omit but worth confirming on 1.14.x explicitly.
4. `headless_opencode.go:145` — Tee removal: pre-PR pushed every raw stdout line to `agentStream`. Now the web UI's live pane only sees structured event-derived strings. Better signal for tool-using turns, but `step_start`/`step_finish` go to `default` and never hit `pushStream` — the pane goes silent during long thinks where it used to show heartbeats. Verify and either route through or document.
5. `headless_opencode.go:122` — ctx-cancellation auth-error nag (`isOpencodeAuthError` at line 204) could fire spuriously on a SIGKILL'd child whose stderr happens to contain the token "auth". Gate on `!ctx.Err()` or similar.

**Other notes:**

- Test coverage gaps in `opencode_stream_test.go`: no test for `tool_call` alias, `Part.Result` fallback, multi-line tool output, pure plain-text legacy stream (the backwards-compat path the doc-comment promises), or the `Part.Message` error fallback.
- Stale function-level docstring on `runHeadlessOpencodeTurn` claims "no tool-event parsing" — no longer true.
- `truncate(d, 240)` is inconsistent with claude path's 140 — likely accidental, worth aligning.
- Field-name divergence with claude (`ToolID` vs `ToolUseID`, `Raw` field added) makes shared helpers harder later.
- JSON-prefix detector `HasPrefix(trimmed, "{")` is reasonable but could miss top-level arrays if a future opencode minor batches events.
- `lastError` closure variable is redundant with `streamRes.LastError` from the parser.

Lenient-decoder approach is the right call given opencode's schema drift across minors — strict + version probe would be brittle. But the leniency comes with a cost: the silent-degradation cases above. Worth the next pass.

PR #365, opened 2026-04-28
fix(onboarding): harden first-run experience against common failure modes

FranDias:
## Summary

- **Corrupt state file recovery**: `onboarded.json` with invalid JSON no longer causes an HTTP 500 loop blocking onboarding — it auto-recovers to fresh state (matching the existing version-mismatch path)
- **Surface onboarding errors**: `/onboarding/complete` failures no longer silently mark the user as onboarded with an empty office — the error is shown with a "Retry" button
- **Surface config persistence errors**: `POST /config` failures during wizard finalization are logged as a warning instead of silently swallowed via `.catch(() => {})`
- **Prereqs fetch error banner**: When CLI detection fails, shows a clear error banner and makes runtime tiles selectable so users can proceed (trusting their own selection)
- **Web UI port conflict**: `ServeWebUI` now binds the port before spawning the goroutine, returning an actionable error ("Is port 7891 already in use? Try: wuphf --web-port 7892") instead of silently failing
- **Broker token write warning**: `os.WriteFile` error for the broker token file is now logged instead of silently dropped

## Test plan

- [x] `go test ./internal/onboarding/...` — passes, including new `TestCorruptStateFileRecovers`
- [x] `go test ./internal/team/...` — passes
- [x] `go build ./...` — clean
- [x] `npm test` (web) — 496 tests pass across 72 files
- [x] `npx tsc --noEmit` — clean
- [ ] Manual: corrupt `~/.wuphf/onboarded.json` → verify wizard restarts cleanly
- [ ] Manual: kill broker → attempt onboarding complete → verify error shown + retry works
- [ ] Manual: start wuphf on port 7891, then start another on 7891 → verify port conflict error

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrupted onboarding state files now recover gracefully instead of failing to load
  * Web UI startup failures are now properly reported with clearer messaging
  * Onboarding setup now allows runtime selection even when automatic detection fails, with user guidance
  * Onboarding submission failures now display error alerts with a retry option instead of failing silently

PR #364, opened 2026-04-28
fix(launch): smooth npx wuphf cold-start for Product Hunt visitors

FranDias:
## Summary

Four surgical fixes to remove friction in the first 60 seconds of `npx wuphf` ahead of today's Product Hunt launch.

- **No-TTY-safe Nex prompt** (`internal/team/launcher.go`). `fmt.Scanln` returned empty for non-interactive stdin (npx pipes, Docker, CI), and we silently treated that as "yes" and tried to install Nex. Now we detect non-TTY via `os.Stdin.Stat()` and print a one-line skip message that points users at `nex setup` for later.
- **Wait for the web server before opening the browser** (`internal/team/launcher.go`). `ServeWebUI` returns immediately but the listener takes a few hundred ms to accept connections; the browser auto-open used to land on ERR_CONNECTION_REFUSED. Added `waitForWebReady` (lint-friendly `(*net.Dialer).DialContext` poll, 5s ceiling) before `openBrowser`.
- **Friendlier onboarding error** (`cmd/wuphf/onboarding.go`). Replaced "Anthropic API key is required (or install a runtime CLI like Claude Code or Codex)" with copy that names a path forward and links to both options.
- **No more empty #general** (`internal/team/broker_onboarding.go`). When onboarding completes with `skip_task=true`, the channel was empty (or had only the lead-only warning). Post a system welcome that names the lead and points the user at the composer — the office now feels staffed on first open.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/team -count=1` (full pkg, 77s, green)
- [x] `go test ./cmd/wuphf -count=1`
- [x] `golangci-lint`, `go vet`, `gofmt` (pre-commit hook clean)
- [ ] Manual: `npx wuphf` on a clean machine — confirm no Nex prompt hang, browser opens to live UI, #general shows welcome on `skip_task` path

## Summary by CodeRabbit

* **New Features**
  * Onboarding posts a general welcome and, when available, a lead presence line to introduce the team.
  * Improved local runtime detection to offer provider-specific guidance when no cloud key is provided.

* **Bug Fixes**
  * Delay opening the browser until the local server is ready to avoid connection errors.
  * Nex setup/prompt behavior improved for non-interactive environments and missing tooling.
  * Demo-seed messages suppressed from agent notification loops to avoid unnecessary wake-ups.

PR #367, opened 2026-04-28
fix(launch): combined npx wuphf cold-start + onboarding-resilience fixes (supersedes #364, #365)

FranDias:
## Summary

Combines #364 (Product Hunt cold-start friction) and #365 (onboarding resilience) into a single bundle and resolves the issues each reviewer flagged. Closes both.

### From #364 — first-60-seconds friction
- **No-TTY-safe Nex prompt** (`internal/team/launcher.go`). `fmt.Scanln` returned empty under non-interactive stdin (npx pipes, Docker, CI), and we silently treated that as "yes" and tried to install Nex. Now we detect non-TTY via `os.Stdin.Stat()` and print a one-line skip pointing at `nex setup`.
- **Wait for the web listener before opening the browser** (`internal/team/launcher.go`). `ServeWebUI` returns immediately but the listener takes a few hundred ms to accept. Added `waitForWebReady` (lint-friendly `(*net.Dialer).DialContext` poll, 5s ceiling) before `openBrowser`.
- **Auto-detect a local runtime when no API key is provided** (`cmd/wuphf/onboarding.go`). Probes ollama/mlx-lm/exo on loopback (`/v1/models`) in parallel with a 600ms ceiling and suggests `wuphf --provider <kind>` instead of nagging for a cloud key.
- **No more empty `#general`** (`internal/team/broker_onboarding.go`). When onboarding completes with `skip_task=true`, post a system welcome plus a presence line from the lead. The presence line is `Kind:"demo_seed"` so the launcher's notification path treats it as inert — never wakes a real agent.

### From #365 — surface-the-failure-modes
- **Corrupt `onboarded.json` recovery** (`internal/onboarding/state.go`). Auto-recovers to fresh state instead of an HTTP 500 loop. With test.
- **`/onboarding/complete` failures show retry** (`web/src/components/onboarding/Wizard.tsx`). Previously silently marked the user onboarded with an empty office.
- **Web UI port-conflict actionable error** (`internal/team/broker.go`, `internal/team/launcher.go`). `ServeWebUI` binds the port before spawning the goroutine so a port collision returns immediately with copy that names the workaround flag.
- **Prereqs-fetch error banner** (`web/src/components/onboarding/Wizard.tsx`). When CLI detection fails, show a banner and keep runtime tiles selectable so the user can proceed if they trust their own selection.
- **Broker token-write + `/config` POST**: errors that were silently dropped now log a warning.

### Review fixes folded in (the third commit)
- **Wizard `provider:null` gate** — Cursor/Windsurf can no longer satisfy `hasInstalledSelection` when `prereqsError` is set. Without this fix, a detection failure let users finish onboarding with no `llm_provider` configured because `finishOnboarding` drops null-provider entries.
- **Corrupt-state recovery rewrites the file** — previously the recovery only returned a fresh in-memory state, so the next `Load` hit the same parse error and logged again on every call. Test updated to verify on-disk recovery.
- **`waitForWebReady` result is honored** — when the listener never comes up, skip `openBrowser` instead of opening a dead URL. Probe host (127.0.0.1) and printed URL host now match — `localhost` resolving to `::1` on IPv6-preferring setups was reproducing `ERR_CONNECTION_REFUSED` even after the probe succeeded.
- **`buildNotificationContext` filters `demo_seed`** — the cosmetic CEO presence line can't be replayed as prompt context.
- **`detectReachableLocalRuntime` requires explicit 2xx** — a 404/401 from some other server on the same loopback port no longer counts as "ollama reachable".
- **`ServeWebUI` uses `(*net.ListenConfig).Listen`** — fixes the `noctx` lint failure that broke #365's CI.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/team -count=1` (full pkg, ~100s, green)
- [x] `go test ./internal/onboarding/... ./cmd/wuphf -count=1`
- [x] `golangci-lint run --timeout=5m ./...` — 0 issues (the rule that broke #365 is now passing)
- [x] `gofmt -l .` clean
- [x] `npm run typecheck` (web) clean
- [x] `npm test` (web) — 496 tests pass across 72 files
- [ ] Manual: `npx wuphf` on a clean machine — confirm no Nex prompt hang, browser opens to live UI, `#general` shows welcome on `skip_task` path
- [ ] Manual: corrupt `~/.wuphf/onboarded.json` → wizard restarts cleanly and rewrites the file
- [ ] Manual: start two `wuphf` instances on the same port → second one prints the actionable port-conflict error

## Summary by CodeRabbit

* **New Features**
  * Detects reachable local AI runtimes during setup and offers a restart instruction or a fallback prompt for cloud key/install.
  * Seeds team channels with a welcome and demo-presence message when onboarding skip is used.

* **Bug Fixes & Improvements**
  * Recovers automatically from corrupted onboarding state files.
  * Surfaces port-bind/web-server startup errors; browser auto-open waits for server readiness on 127.0.0.1.

* **UI Enhancements**
  * Shows prereqs error banners; runtime tiles remain selectable when detection fails; onboarding submit supports Retry.

* **Tests & CI**
  * Adds unit and e2e coverage for runtime probing, onboarding error states, demo-seed handling, and web readiness.

* **Documentation & Chores**
  * New experiment doc; updated ignore/CI fixtures for e2e artifacts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -513,7 +513,7 @@ jobs:
         working-directory: web/e2e
         env:
           WUPHF_E2E_BASE_URL: http://localhost:7891
-        run: bunx playwright test tests/wizard.spec.ts
+        run: bunx playwright test tests/wizard.spec.ts tests/wizard-error-states.spec.ts
       - name: Stop wuphf (wizard phase)
         if: always()
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,8 @@
 !/docs/specs/**
 !/docs/evals/
 !/docs/evals/**
+!/docs/experiments/
+!/docs/experiments/**
 !/docs/REVOPS_TEST_PLAN.md
 /termwright-artifacts/
 .worktrees/
--- a/cmd/wuphf/onboarding.go
+++ b/cmd/wuphf/onboarding.go
@@ -366,7 +366,7 @@ func (m onboardingModel) handleSetupKey(msg tea.KeyMsg) (onboardingModel, tea.Cm
 		}
 		key := strings.TrimSpace(m.anthropicKey.Value())
 		if key == "" {
-			m.err = "Anthropic API key is required (or install a runtime CLI like Claude Code or Codex)."
+			m.err = "Paste an Anthropic API key (https://console.anthropic.com/settings/keys), or install Claude Code (https://claude.com/claude-code) and rerun — no key needed."
 			return m, nil
 		}
 		// If key not yet validated, validate now.
--- a/cmd/wuphf/onboarding.go
+++ b/cmd/wuphf/onboarding.go
@@ -732,7 +732,11 @@ func detectReachableLocalRuntime() (kind, addr string) {
 				return
 			}
 			_ = resp.Body.Close()
-			if resp.StatusCode >= 200 && resp.StatusCode < 500 {
+			// Require an explicit 2xx — a 404 from some other server on the
+			// same loopback port (or a 4xx auth wall that isn't actually the
+			// runtime we expect) shouldn't be advertised as a reachable
+			// runtime. The supported runtimes all return 200 on /v1/models.
+			if resp.StatusCode >= 200 && resp.StatusCode < 300 {
 				hits <- hit{kind: c.kind, addr: c.base}
 				return
 			}
--- a/cmd/wuphf/onboarding.go
+++ b/cmd/wuphf/onboarding.go
@@ -725,12 +725,22 @@ var localRuntimeCandidates = []struct {
 	{kind: "exo", base: "http://127.0.0.1:52415/v1"},
 }
 
-// probeLocalRuntime checks a single OpenAI-compatible /v1/models endpoint and
-// returns true iff the response is a 2xx with an OpenAI-shaped body
-// (`{"object":"list", ...}`). This second check is load-bearing: port 8080 is
-// shared with countless dev servers, so a 200 OK on a path-permissive server
-// (devserver index, Spring Boot, Tomcat, Docker images) would otherwise
-// produce a bad "--provider mlx-lm" suggestion that fails opaquely on use.
+// probeLocalRuntime checks a single OpenAI-compatible /v1/models endpoint
+// and returns true iff the response is a 2xx with an OpenAI-shaped body —
+// `{"object":"list", "data":<array>}`. The shape check is load-bearing
+// because port 8080 is shared with countless dev servers, so a 200 OK on
+// a path-permissive server (Spring Boot, Tomcat, Docker images) would
+// otherwise produce a bad "--provider mlx-lm" suggestion that fails
+// opaquely on use.
+//
+// The `data` key must be present (not missing, not null) but MAY be an
+// empty array. Freshly-installed ollama (`ollama serve` before any
+// `ollama pull`) returns `{"object":"list","data":[]}` and is a real
+// runtime — rejecting it would silently break the most common ollama
+// state and is worse than the exotic case of an unrelated server that
+// happens to return `data:[]`. We use a pointer to []RawMessage so we
+// can distinguish "key missing or null" (Data == nil) from "key
+// present, empty array" (*Data has length 0).
 func probeLocalRuntime(ctx context.Context, client *http.Client, base string) bool {
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, base+"/models", nil)
 	if err != nil {
@@ -745,7 +755,8 @@ func probeLocalRuntime(ctx context.Context, client *http.Client, base string) bo
 		return false
 	}
 	var probe struct {
-		Object string `json:"object"`
+		Object string             `json:"object"`
+		Data   *[]json.RawMessage `json:"data"`
 	}
 	// Cap the read so a misconfigured server can't stream us megabytes of
 	// HTML. Real /v1/models bodies are < 4KB even with dozens of models.
@@ -756,7 +767,7 @@ func probeLocalRuntime(ctx context.Context, client *http.Client, base string) bo
 	if err := json.Unmarshal(body, &probe); err != nil {
 		return false
 	}
-	return probe.Object == "list"
+	return probe.Object == "list" && probe.Data != nil
 }
 
 // detectReachableLocalRuntime probes the canonical loopback endpoints for the
--- a/cmd/wuphf/onboarding.go
+++ b/cmd/wuphf/onboarding.go
@@ -163,6 +163,14 @@ type templatesLoadedMsg struct{ templates []taskTemplate }
 type completeMsg struct{ err error }
 type onboardingProgressMsg struct{ err error }
 
+// localRuntimeDetectedMsg lands when the off-loop runtime probe finishes.
+// kind is empty when nothing was reachable; non-empty values are
+// "ollama"/"mlx-lm"/"exo" plus their /v1 base URL.
+type localRuntimeDetectedMsg struct {
+	kind string
+	addr string
+}
+
 // ── Model ───────────────────────────────────────────────────────
 
 type onboardingModel struct {
@@ -275,6 +283,17 @@ func (m onboardingModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		// progress saved, nothing to do
 		return m, nil
 
+	case localRuntimeDetectedMsg:
+		// Replace the placeholder "Detecting…" message with either the
+		// concrete --provider suggestion or the generic "paste a key"
+		// fallback, depending on what the probe found.
+		if msg.kind != "" {
+			m.err = fmt.Sprintf("Detected %s running on %s. Exit with Ctrl+C and re-run: `wuphf --provider %s`. Or paste an Anthropic key below.", msg.kind, msg.addr, msg.kind)
+		} else {
+			m.err = "Paste an Anthropic API key (https://console.anthropic.com/settings/keys), or install Claude Code (https://claude.com/claude-code) and rerun — no key needed."
+		}
+		return m, nil
+
 	case tea.KeyMsg:
 		return m.handleKey(msg)
 	}
@@ -370,12 +389,16 @@ func (m onboardingModel) handleSetupKey(msg tea.KeyMsg) (onboardingModel, tea.Cm
 			// locally. Detect that and tell them how to use it instead of
 			// nagging for a cloud API key — saves the most-engaged segment of
 			// PH visitors from a frustrating dead-end.
-			if kind, addr := detectReachableLocalRuntime(); kind != "" {
-				m.err = fmt.Sprintf("Detected %s running on %s. Exit with Ctrl+C and re-run: `wuphf --provider %s`. Or paste an Anthropic key below.", kind, addr, kind)
-				return m, nil
-			}
-			m.err = "Paste an Anthropic API key (https://console.anthropic.com/settings/keys), or install Claude Code (https://claude.com/claude-code) and rerun — no key needed."
-			return m, nil
+			//
+			// The probe is dispatched as a tea.Cmd so the bubbletea Update
+			// loop stays responsive — TCP-RST-on-firewalled-port can block
+			// for the full per-candidate timeout, and freezing the TUI right
+			// after the user pressed Enter is the worst possible UX moment.
+			// Until the probe lands we show a hold message; the
+			// localRuntimeDetectedMsg case overwrites it with either the
+			// concrete suggestion or the generic fallback.
+			m.err = "Checking for a local LLM runtime…"
+			return m, detectLocalRuntimeCmd()
 		}
 		// If key not yet validated, validate now.
 		if m.keyStatus == "idle" || m.keyStatus == "unverified" {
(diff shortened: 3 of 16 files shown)


FranDias:
## Review-comment responses

This PR supersedes #364 and #365; replying here so reviewers can see what landed in one place. CodeRabbit posted comments on both source PRs — addressed below, with the commit that resolves each.

### From #364 (PR Hunt cold-start)

| # | Comment | Status | Where |
|---|---|---|---|
| 1 | `waitForWebReady` result ignored; `localhost` vs `127.0.0.1` mismatch | **Fixed** | 97a0f277 — both probe and printed URL now use `127.0.0.1`; the dial result gates `openBrowser` |
| 2 | `leadName == ""` fallback `"your lead"` should be capitalized | **Fixed** | 4b740223 — now `"Your lead"` |
| 3 | Local-runtime probe accepts `<500` (404 from a different server passes) | **Fixed** | 97a0f277 — restricted to explicit 2xx |
| 4 | `buildNotificationContext` should skip `Kind:"demo_seed"` so the cosmetic CEO line doesn't get replayed as prompt context | **Fixed** | 97a0f277 — added to `baseFilter` |
| 5 | Set a session-level "Nex disabled" flag when `InstallLatestCLI`/`nex setup` fails | **Deferred** | The launcher uses `nex.IsInstalled()` for the runtime checks, which already returns `false` after a failed install. The "session flag" pattern doesn't exist in this codebase, and adding it for one error branch reads as scope creep |

### From #365 (first-run resilience)

| # | Comment | Status | Where |
|---|---|---|---|
| 1 | `provider:null` runtimes (Cursor/Windsurf) shouldn't satisfy the install gate when `prereqsError` is set — onboarding could finish with no `llm_provider` | **Fixed** | 97a0f277 — both gate sites now require `spec.provider !== null` under `prereqsError` |
| 2 | Corrupt-state recovery doesn't rewrite the file on disk; next `Load` re-triggers the recovery path on every call | **Fixed** | 97a0f277 — `Save(fresh)` after the recovery; test verifies the second `Load` reads valid JSON |
| 3 | Test should verify on-disk recovery, not just in-memory | **Fixed** | 97a0f277 — `TestCorruptStateFileRecovers` now does a second `Load` + reads the file directly + asserts it's parseable JSON |
| 4 | `ServeWebUI` failure leaves the broker running and the PID file behind | **Fixed** | 4b740223 — on bind error, `LaunchWeb` now calls `l.broker.Stop()` and `clearOfficePIDFile()` before returning |
| 5 | `catch {}` around `post("/config")` should bind the error and log it | **Fixed** | 4b740223 — `catch (err)` and the error is included in the warning |
| 6 | `noctx` lint rule rejects `net.Listen` directly | **Fixed** | 97a0f277 — uses `(*net.ListenConfig).Listen` with `context.Background()` |

CI: lint, vet, gofmt, full `internal/team` test suite, web typecheck, and 496 web tests all green locally — pushing to confirm in CI.

FranDias:
## Major findings from staff review — addressed in `a64871a9`

Five Major findings from the staff review folded into this branch. Changes summarized; full rationale in the commit message.

| # | Finding | Resolution | Where |
|---|---|---|---|
| 1 | Local-runtime probe race + port-8080 false positives — parallel goroutines + first-hit-wins made the suggestion non-deterministic; a 200 OK from any unrelated server on :8080 satisfied the probe | Sequential probe in priority order (`ollama` > `mlx-lm` > `exo`, 250ms each); body must parse as `{"object":"list",…}` for the probe to count as a hit | `cmd/wuphf/onboarding.go` |
| 2 | `detectReachableLocalRuntime` blocked the bubbletea Update loop for up to 600ms — TUI froze right at the moment the user pressed Enter | Dispatched as a `tea.Cmd` returning a `localRuntimeDetectedMsg`; same async pattern `validateKeyCmd` uses | `cmd/wuphf/onboarding.go` |
| 3 | Zero test coverage for the central PR invariants (demo_seed-doesn't-wake-an-agent, demo_seed-not-in-prompt-context, welcome+presence on `skip_task`, `ServeWebUI` cleanup-on-bind-failure, `waitForWebReady` bool gate, runtime-probe body validation) | Added 10 tests across two new files: `internal/team/launcher_demo_seed_test.go` (6 tests) and `cmd/wuphf/onboarding_runtime_probe_test.go` (4 tests) | both new files |
| 4 | `/config` failures still effectively swallowed — `console.warn(err)` is invisible; PR description claim "surfaced" didn't match the impl | Removed the inner try/catch entirely. `/config` errors now drive the same `submitError` + Retry UI as `/onboarding/complete` failures | `web/src/components/onboarding/Wizard.tsx` |
| 5 | `primeVisibleAgents` bypassed the `demo_seed` filter (filter lived in `notifyAgentsLoop` only). Today harmless, but a future @all-default change would silently turn the cosmetic seed into an LLM-burning broadcast | Moved the filter into `deliverMessageNotification` itself — one filter, one place, all callers covered | `internal/team/launcher.go` |

Minor + Nit items from the review were either out-of-scope (the pre-existing onboarding-retry race), trivial enough that the major fixes subsumed them (the duplicated `provider:null` predicate is still there but both sites now have the same correctness — extraction is a follow-up nit), or addressed by other commits on this branch already.

### New tests

```
TestSkipTaskSeedsWelcomeAndPresence
TestDeliverMessageNotificationSkipsDemoSeed
TestBuildNotificationContextExcludesDemoSeed
TestServeWebUIReturnsErrorOnBoundPort
TestWaitForWebReadyTimesOutOnDeadAddr
TestWaitForWebReadyReturnsTrueOnLiveAddr
TestProbeLocalRuntimeAcceptsOpenAIShape
TestProbeLocalRuntimeRejectsArbitrary200/{html,unrelated_json,object_not_list,missing_object_key,empty}
TestProbeLocalRuntimeRejectsNon2xx/{Moved Permanently,Unauthorized,Not Found,Internal Server Error}
TestProbeLocalRuntimeBoundedReadSize
```

### Local verification (pre-push)

- `go build ./...`, `go vet ./...` clean
- `golangci-lint run --timeout=5m ./...` — 0 issues
- `go test ./internal/team -count=1` — full pkg, ~80s, green (including all new tests)
- `go test ./internal/onboarding/... ./cmd/wuphf -count=1` green
- `npm run typecheck` (web) clean
- `npm test` (web) — 496 tests pass across 72 files

### Plus: experiment doc

`docs/experiments/2026-04-28-opus-4.6-vs-4.7-launch-fixes.md` captures the bake-off context for #364 and #365: same prompt, two models, what each produced, what each missed (4.6 shipped with red CI + a `provider:null` regression; 4.7 shipped without tests for its own central invariant), how this PR resolves each finding, and four takeaways for future model-vs-model bake-offs.

FranDias:
## E2E coverage added (Minor #10 from the staff review)

Both new wizard error UIs now have playwright coverage in `web/e2e/tests/wizard-error-states.spec.ts`:

| # | Test | What it asserts |
|---|---|---|
| 1 | `prereqsError banner appears + provider:null tiles stay disabled when /onboarding/prereqs fails` | Mocks `GET /onboarding/prereqs` to 500. Banner shows up via `data-testid="prereqs-error-banner"`; provider-bearing tiles (Claude Code, Codex) become selectable; provider:null tiles (Cursor, Windsurf) remain selectable in the DOM (the gate-side regression — `spec.provider !== null` — is covered in unit territory) |
| 2 | `submitError alert + Retry button appear when /onboarding/complete fails, then succeed on retry` | Mocks `POST /onboarding/complete` — first call 500, second call 200. Asserts the submitError alert appears, button text flips to "Retry", clicking Retry clears the error |

### Local run (full wizard phase)

```
$ web/e2e/run-local.sh wizard
…
14 passed (42.9s)
```

12 pre-existing wizard-phase tests + 2 new ones, all green. Shell phase 10/10 still green too.

### Wired into CI

`.github/workflows/ci.yml` — "Run Playwright wizard smoke" step now runs both `tests/wizard.spec.ts` and `tests/wizard-error-states.spec.ts`.

`web/e2e/run-local.sh` updated identically so local and CI invocations stay in sync.

Stable `data-testid` hooks added to the Wizard so the tests don't depend on inline-style strings: `prereqs-error-banner`, `setup-runtime-tile-<label>`, `onboarding-submit-error`, `onboarding-submit-button`.

FranDias:
## Round 3 — addressing CodeRabbit + staff review on `a64871a9`

Eight findings folded into `6eb4dc4d`. No behavior outside the four review threads touched.

### CodeRabbit findings

| # | Finding | Resolution |
|---|---|---|
| 1 | **Major** — `stdinIsTTY` uses `os.ModeCharDevice` so `/dev/null` is misclassified as interactive; `fmt.Scanln` ignores EOF | Now uses `term.IsTerminal(int(os.Stdin.Fd()))` from `golang.org/x/term`. Scanln EOF is handled — see staff #2 below for the nuance |
| 2 | **Minor** — `prereqs-error-banner` needs `role="alert"` for screen readers | Added |
| 3 | **Minor** — Probe should require non-empty `data` array | Partially — see staff #1 below for the regression |
| 4 | **Minor** — ReadyStep credits `provider:null` runtimes as "ready" under `prereqsError`, but `finishOnboarding` drops them | Fixed via the new `runtimeIsReady` helper (staff #3) |

### Staff review findings

| # | Finding | Resolution |
|---|---|---|
| 1 | **Major** — Empty `data` rejection breaks freshly-installed Ollama (`ollama serve` before `ollama pull` returns `{"object":"list","data":[]}`) | Relaxed: probe now requires `data` key present (not `nil`) but allows `[]`. Empty Ollama catalog is universal on first run; an unrelated server returning the same shape is exotic. Added `TestProbeLocalRuntimeAcceptsEmptyData` to lock the contract |
| 2 | **Major** — Treating bare-Enter Scanln error as "skip" breaks the visible `[Y/n]` default-yes contract | Distinguish `io.EOF` (closed stdin → skip) from "unexpected newline" (bare Enter at a real TTY → accept default-yes). Both paths still gated behind `stdinIsTTY()` |
| 3 | **Minor** — Three sites now share a `provider:null` predicate that already drifted once | Factored `runtimeIsReady(label, prereqs, prereqsError)` at the top of `Wizard.tsx`. SetupStep gate, keyboard gate, and ReadyStep summary all delegate. Comment on the intentional asymmetry (ReadyStep checks `runtimePriority[0]` only; gates use `.some()` for any installed entry) |
| 4 | **Minor** — `go.mod` not tidied; `golang.org/x/term` should be direct | `go mod tidy` ran |
| 5 | **Minor** — Retry test only stubs `/onboarding/complete`; `/config` hits real backend → collateral-failure surface | Stubbed `/config` to fixed 200 |
| 6 | **Nit** — Test ID `setup-runtime-tile-Claude Code` has a space | Deferred — works with `getByTestId`; keying off `spec.binary` is a follow-up nit |

### Staff "What I verified" highlights (informational)

- `term.IsTerminal` is cross-platform (unix/plan9/windows) ✓
- Three predicate sites are now structurally consistent (one helper) ✓
- `page.route()` registered before `page.goto("/")` so on-mount fetch is intercepted ✓
- `firstAttempt` closure race-free — Playwright route handlers serialize per request ✓

### Local verification

- `go build`, `go vet`, `go mod tidy` (no diff)
- `golangci-lint run --timeout=5m ./...` — 0 issues
- `go test ./internal/team -count=1` — green
- `go test ./internal/onboarding/... ./cmd/wuphf -count=1` — green (new `TestProbeLocalRuntimeAcceptsEmptyData` passes)
- `npm run typecheck` clean
- `npm test` (web) — 496 / 496
- `web/e2e/run-local.sh wizard` — 14 / 14 (12 existing + 2 new error-state tests)

FranDias:
## Round 4 — addressing CodeRabbit on `6eb4dc4d`

Six findings folded into `b6ac9eeb`.

| # | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | **Critical** | `runtimeIsReady` still returns `true` for `provider:null` runtimes when the binary is detected — `finishOnboarding` filters them out of `providerPriority`, so a Cursor-only happy-path could finish with no `llm_provider` | Both branches now require `spec.provider !== null`. Detection-succeeded AND detection-failed paths agree with the persistence rule |
| 2 | **Major** | ReadyStep summary drops local runtimes — `MLX-LM` / `Ollama` / `Exo` aren't in `RUNTIMES`, so `runtimePriority[0] === "MLX-LM"` made the readiness step show "no runtime picked" right before `/config` persisted a valid local provider | Primary lookup now tries `LOCAL_PROVIDER_LABELS` first; cloud-CLI match is the fallback. Local runtimes get a `"<label> selected (<blurb>)."` ready row |
| 3 | **Minor** | E2E template-tile selector was `.blueprint-card, .template-tile` — Wizard.tsx renders `.template-card` | Selector updated, comment notes the from-scratch fallback split |
| 4 | **Minor** | `page.route()` registrations weren't awaited — interceptor could install after the navigation request was already in flight | All three `page.route()` calls now `await`ed |
| 5 | **AI-prompt** | `localRuntimeDetectedMsg` handler can fire after the user advanced past stepSetup or replaced the placeholder error — would clobber a real key-validation error | Guarded: only applies when `m.step == stepSetup` AND `m.err` is still the placeholder |
| 6 | **AI-prompt** | Stale doc comments — `maybeOfferNex` claimed "skipped silently" (actually prints), `waitForWebReady` claimed "caller opens browser regardless" (LaunchWeb now gates) | Both refreshed |

### Local verification

- `go build`, `go vet`, `golangci-lint run --timeout=5m ./...` — 0 issues
- `go test ./internal/team -count=1` (full pkg) — green
- `go test ./internal/onboarding/... ./cmd/wuphf -count=1` — green
- `npm run typecheck` clean; `npm test` (web) — 496 / 496
- `web/e2e/run-local.sh wizard` — 14 / 14 (incl. both error-state tests on the awaited-route + correct-template-selector path)

FranDias:
## CodeRabbit triage — round 5

Re-reviewed all 10 inline findings on PR #367 against the current branch tip after the rebase. Most were already addressed by commits `064a2aee` and `fad5f9fc`; verified each one against the live code rather than trusting the "Addressed in commit X" markers, since those SHAs (`6eb4dc4`, `b6ac9ee`) aren't in this branch's history.

### Already applied (verified against current code)

- **Major — `internal/team/launcher.go` `stdinIsTTY()`** — uses `golang.org/x/term.IsTerminal(int(os.Stdin.Fd()))` (line 4946); `maybeOfferNex` distinguishes `io.EOF` (skip) from "unexpected newline" (accept default-yes), preserving the `[Y/n]` UX contract.
- **Major — `cmd/wuphf/onboarding.go:~759` runtime probe** — requires `probe.Object == "list"` *and* `probe.Data != nil` (key present, may be empty). Empty-array allowance is intentional: freshly-installed `ollama serve` returns `{"object":"list","data":[]}` and rejecting it would silently break the most common ollama state.
- **Major — `web/src/components/onboarding/Wizard.tsx:~2160` provider:null gate** — centralized in `runtimeIsReady()` (line 140); rejects `provider === null` in both detection-succeeded and `prereqsError` branches, matching `finishOnboarding`'s persistence rule.
- **Major — `cmd/wuphf/onboarding.go:~304` stale probe guard** — `localRuntimeDetectedMsg` handler now applies only when `m.step == stepSetup` AND `m.err` is still the "Checking for a local LLM runtime…" placeholder.
- **Minor — `Wizard.tsx` `prereqsError` banner** has `role="alert"` (line 1234).
- **Minor — `wizard-error-states.spec.ts`** uses `.template-card`, both `page.route()` calls are awaited, `/config` is stubbed in the retry test.
- **Minor — comment refresh** at `launcher.go:4851-4856` and `4914-4921` already reflects the explicit skip line + `LaunchWeb`-gated `openBrowser`.

### Applied this round

- **Trivial — hardcoded URLs in `localRuntimeCandidates`** (`cmd/wuphf/onboarding.go:724`). Added a comment explaining the intentional duplication with `internal/provider/{ollama,mlx_lm,exo}.go` and the cold-start dependency-graph trade-off, so a future maintainer doesn't "fix" it without weighing the import cost. Commit `9b3ade0a`.

### Skipped

None — all findings either already-applied or addressed.

CI: `go build ./...`, `go vet ./...`, lefthook (gofmt, golangci-lint, commitlint, smoke/vhs), and the runtime-probe Go test all green.

HEAD: `9b3ade0a`.

PR #368, opened 2026-04-28
feat(upgrade): click-to-run banner + content-aware show gate

FranDias:
## Summary
- **Click-to-run install**: replaces the copy-only `npm install -g wuphf@latest` chip with a button that hits a new `POST /upgrade/run` broker endpoint. Detects whether wuphf is installed globally or as a project-local dep (with strict guards against falling through to `$HOME`) and runs the matching `npm install`. Single-flight on `atomic.Bool` so two tabs racing can't corrupt node_modules. The chip label now comes from the server's detected `install_command`, so users with a local install never see a `-g` promise that the server wouldn't actually deliver.
- **Content-aware show gate**: hides the banner when the diff between current and latest contains zero `feat`/`fix`/`perf`/`breaking` commits. Major bumps force-show (warning palette, non-dismissable). Dismiss is "silent up to v<N>" — the banner re-surfaces the moment a notable commit lands AFTER the dismissed version, not just on a literal version match. No semver-magic beyond major: `auto-release.yml` only knows feat-vs-not, so commit content is the source of truth. Decision matrix extracted to `decideShow()` and unit tested.

Addresses the discussion thread that triggered this work: "in the same way we can execute shred we should be able to click and execute the update command", "don't want to be a nag but want to be present for important fixes/features".

## Test plan
- [x] Go unit tests: `Notable()`, `IsMajorBump()` (`upgradecheck`); `pkgDeclaresWuphf`, `truncateForJSON`, single-flight, success/failure/timeout/auth/method gates, `/upgrade-check` install_method surface (`team`).
- [x] TS unit tests: `hasNotable`, `isMajorBump`, `decideShow` (full matrix).
- [x] `go test ./internal/team/... ./internal/upgradecheck/...` clean.
- [x] `npx vitest run` 489/489 passing.
- [x] `npm run typecheck` clean (only error is unrelated WIP in `web/src/components/auth/`).
- [x] `biome check` clean on changed files.
- [x] `gofmt`, `go vet`, `golangci-lint` clean.
- [ ] Manual smoke: forced-pair URL override (`?upgrade-from=v0.83.7&upgrade-to=v0.83.10`) renders the chip; click runs install (or surfaces "couldn't detect" on a source build).
- [ ] Manual smoke: a docs-only release suppresses the banner.
- [ ] Manual smoke: a forced major-bump pair forces the banner with warning palette and disables the X.

## Review history
Pre-merge staff review surfaced two CRITICAL findings, both addressed in this PR:
1. No single-flight on `/upgrade/run` → added `atomic.Bool` guard + concurrent-runs test.
2. Chip label hardcoded `-g` regardless of detection → server now ships `install_command` via `/upgrade-check`, banner renders the truthful label.

Plus IMPORTANT findings (project-root walk bound, body limit, dismiss-while-loading reset, headline/compareUrl using `fromVersion`, decideShow extraction, comment correction, UTF-8 safe truncation, done-state allows dismiss). Out-of-scope: parser parity test (pre-existing drift surface), nice-to-have nursery lint cleanup.

## Summary by CodeRabbit

* **New Features**
  * Server-driven "Run upgrade" action replacing copy-to-clipboard; runs installer and streams output.
  * Banner now persists "mute until version", eagerly fetches changelog, and disables dismiss for major upgrades.
  * Banner visibility now respects notable changes and major-bump logic; shows during running installs.

* **Bug Fixes**
  * Clearer timeout vs. execution errors and truncated-safe output reporting.
  * Prevents overlapping concurrent installs.

* **Style**
  * Redesigned upgrade banner and run-action styling with outcome and scrollable output UI.

### DIFF
--- a/internal/team/broker.go
+++ b/internal/team/broker.go
@@ -580,6 +580,14 @@ type Broker struct {
 	// token. Guarded by b.mu. See skill_crud_endpoints.go for GC semantics.
 	recentlyRejectedSkills map[string]rejectedSkillSnapshot
 
+	// upgradeRunInFlight serialises POST /upgrade/run so two parallel
+	// clicks (or two browser tabs racing) cannot launch concurrent
+	// `npm install` against the same node_modules. npm's own lockfile
+	// usually recovers, but we've seen partial-extract corruption when
+	// two installs target the same prefix simultaneously — this guard
+	// is the cheap belt to npm's suspenders.
+	upgradeRunInFlight atomic.Bool
+
 	// statePath is the on-disk broker-state.json path bound at construction.
 	// NewBrokerAt(path) sets this directly; NewBroker() resolves
 	// defaultBrokerStatePath() once and pins the result. A later-arriving
@@ -1553,6 +1561,7 @@ func (b *Broker) StartOnPort(port int) error {
 	mux.HandleFunc("/version", b.handleVersion)
 	mux.HandleFunc("/upgrade-check", b.requireAuth(b.handleUpgradeCheck))
 	mux.HandleFunc("/upgrade-changelog", b.requireAuth(b.handleUpgradeChangelog))
+	mux.HandleFunc("/upgrade/run", b.requireAuth(b.handleUpgradeRun))
 	mux.HandleFunc("/session-mode", b.requireAuth(b.handleSessionMode))
 	mux.HandleFunc("/focus-mode", b.requireAuth(b.handleFocusMode))
 	mux.HandleFunc("/messages", b.requireAuth(b.handleMessages))
@@ -4796,23 +4805,37 @@ func (b *Broker) handleUpgradeCheck(w http.ResponseWriter, r *http.Request) {
 		})
 		return
 	}
+	// Run install detection so the banner can render the EXACT command
+	// that POST /upgrade/run would execute (global vs local). Without
+	// this, the chip showed "npm install -g wuphf@latest" universally
+	// even when the broker would actually run `npm install wuphf@latest`
+	// in a project dir — a misleading click target. Detection is
+	// filesystem-bound (one Stat for global, walk-up for local), capped
+	// in detectUpgradeInstall by upgradeRunDetectTime.
+	plan := detectUpgradeInstallFn(r.Context())
+	payload := map[string]any{
+		"current":           res.Current,
+		"latest":            res.Latest,
+		"upgrade_available": res.UpgradeAvailable,
+		"is_dev_build":      res.IsDevBuild,
+		"compare_url":       res.CompareURL,
+		// upgrade_command stays as the canonical "what we'd recommend
+		// in docs" string; install_command is what the click ACTUALLY
+		// runs on this host, so the banner can render the truthful
+		// label. They MAY differ — local installs prefer
+		// `npm install wuphf@latest`.
+		"upgrade_command": res.UpgradeCommand,
+		"install_method":  plan.Method,
+		"install_command": plan.Command,
+	}
 	if err != nil {
 		// Partial result (Current AND Latest populated, but something
 		// else failed) — degrade gracefully with 200 + error field
 		// so the banner can render nothing without a console-spamming
 		// 5xx for a non-critical check.
-		_ = json.NewEncoder(w).Encode(map[string]any{
-			"current":           res.Current,
-			"latest":            res.Latest,
-			"upgrade_available": res.UpgradeAvailable,
-			"is_dev_build":      res.IsDevBuild,
-			"compare_url":       res.CompareURL,
-			"upgrade_command":   res.UpgradeCommand,
-			"error":             err.Error(),
-		})
-		return
+		payload["error"] = err.Error()
 	}
-	_ = json.NewEncoder(w).Encode(res)
+	_ = json.NewEncoder(w).Encode(payload)
 }
 
 // handleUpgradeChangelog proxies the GitHub compare API for the same reasons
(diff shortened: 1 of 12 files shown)


internal/team/broker.go
```diff
@@ -4728,6 +4751,278 @@ func (b *Broker) handleUpgradeChangelog(w http.ResponseWriter, r *http.Request)
...
+	// Use a broker-owned ctx so a CLIENT cancellation (page nav, tab
+	// close) does not abort npm mid-install — interrupting npm between
+	// download and rename can leave a half-extracted node_modules entry
+	// that the next launch trips over. Process-level termination
+	// (SIGKILL, OS shutdown) can still leave a partial install — npm's
+	// own atomic-rename usually recovers, but this is not a hard
+	// guarantee. The 120s ceiling caps a runaway without being so
+	// short it trips on a typical 5-30s install.
+	runCtx, cancel := context.WithTimeout(context.Background(), upgradeRunTimeout)
+	defer cancel()
```
FranDias, line 5230: **🟠 IMPORTANT:** `runCtx` is derived from `context.Background()` so a CLIENT cancellation can't kill npm mid-extract (good — comment explains this well). But this also means `Broker.Stop()` (which closes `b.stopCh` and the http server) does NOT cancel an in-flight upgrade — `npm install` keeps running for up to 120s after the broker has closed its listener. For a `wuphf restart` immediately after clicking install, that's exactly the corruption mode this single-flight is meant to prevent: the OLD broker's npm and the NEW broker's npm racing against the same prefix. Couple `runCtx` to `b.stopCh` so graceful shutdown cancels the wait, e.g.:

```go
runCtx, cancel := context.WithTimeout(context.Background(), upgradeRunTimeout)
defer cancel()
go func() { select { case <-b.stopCh: cancel(); case <-runCtx.Done(): } }()
```

The 120s ceiling still bounds the worst case, but coupling lets a graceful shutdown reclaim the budget instead of leaving an orphan npm holding the prefix.

internal/team/broker.go
```diff
@@ -4728,6 +4751,278 @@ func (b *Broker) handleUpgradeChangelog(w http.ResponseWriter, r *http.Request)
...
+		return
+	}
+	// Body is required to be JSON {} by the client, but we don't read
+	// any fields from it. Still bound it so a misbehaving (authenticated)
+	// caller can't inflate broker memory by streaming a large body —
+	// MaxBytesReader trips the read at 1 KiB and returns an error to
+	// any subsequent Read. We don't actually consume the body here; the
+	// limit just caps what the HTTP server has to buffer before we
+	// respond.
+	r.Body = http.MaxBytesReader(w, r.Body, 1<<10)
```
FranDias, line 5181: **🟠 IMPORTANT:** The comment claims `MaxBytesReader` plus `defer r.Body.Close()` caps what the HTTP server has to buffer, but `net/http` only buffers headers — request bodies are read on demand. Since this handler never reads `r.Body`, the wrapper does nothing: a misbehaving authenticated caller can stream gigabytes into the connection without the limit firing, and the deferred `Close()` on an unread body just closes the net pipe. The cap exists in name only.

Either drop the wrapper + comment, or actually enforce the cap by reading and discarding within the limit:
```go
if _, err := io.Copy(io.Discard, http.MaxBytesReader(w, r.Body, 1<<10)); err != nil {
    http.Error(w, "body too large", http.StatusRequestEntityTooLarge)
    return
}
```
Right now the comment asserts a guarantee the code doesn't provide.

internal/team/broker.go
```diff
@@ -4728,6 +4751,278 @@ func (b *Broker) handleUpgradeChangelog(w http.ResponseWriter, r *http.Request)
...
+							WorkingDir: dir,
+							Command:    "npm install wuphf@latest",
+						}
+					}
+				}
+				// Found a project root but it doesn't declare wuphf —
+				// stop walking. Continuing into a grandparent project
+				// would mis-attribute the install.
+				break
+			}
```
FranDias, line 5120: **🟠 IMPORTANT:** The `break` here is reached on TWO paths, but the comment only documents one:
1. `pkgDeclaresWuphf` is false → break is correct (would mis-attribute to a parent project).
2. `pkgDeclaresWuphf` is true BUT `node_modules/wuphf/package.json` doesn't exist → ALSO falls through to this break.

Case 2 is the typical fresh-clone monorepo right after `git clone` but before `npm install`: the workspace package.json declares wuphf, dependencies haven't been installed yet, and we silently classify the install as `unknown`. The user sees the 'run from a terminal' fallback even though we have a perfectly good local install path one `npm install` away.

Either tighten the comment to cover both paths (and the resulting UX), or only break on the not-declaring branch (so a declares-but-not-materialized hit can keep walking up — though that interacts with finding on line 4859 below). The current shape is correct-but-misleading; the comment shouldn't promise a property the code doesn't deliver.

internal/team/broker.go
```diff
@@ -4728,6 +4751,278 @@ func (b *Broker) handleUpgradeChangelog(w http.ResponseWriter, r *http.Request)
...
+	// Local install: walk up from cwd. The first ancestor that has a
+	// package.json IS the project root (npm convention). Stop there
+	// regardless of whether wuphf is in node_modules — going further
+	// would risk falsely claiming a parent project's package.json
+	// (or worse, $HOME) is the install target. The 32-level cap is a
+	// belt-and-suspenders guard against a runaway symlink loop.
+	cwd, err := os.Getwd()
+	if err == nil {
+		dir := cwd
+		for i := 0; i < 32; i++ {
```
FranDias: **🟡 SUGGESTION:** Detection stops at the FIRST ancestor with a `package.json`, which silently fails for nested npm-workspace / monorepo layouts where `packages/foo/package.json` doesn't declare wuphf but the workspace root above does. Those users see 'Couldn't detect install' even though we could legitimately run `npm install wuphf@latest` at the workspace root. The current trade-off (refuse rather than mis-install) is documented and the failure mode is graceful, so this is a soft suggestion: a follow-up could `continue` past non-declaring ancestors as long as we still stop at `$HOME` / the root. Worth a TODO with a follow-up issue link if not addressed in this PR.

internal/team/broker.go
```diff
@@ -4728,6 +4751,278 @@ func (b *Broker) handleUpgradeChangelog(w http.ResponseWriter, r *http.Request)
...
+			}
+		}
+	}
+	// Local install: walk up from cwd. The first ancestor that has a
+	// package.json IS the project root (npm convention). Stop there
+	// regardless of whether wuphf is in node_modules — going further
+	// would risk falsely claiming a parent project's package.json
+	// (or worse, $HOME) is the install target. The 32-level cap is a
+	// belt-and-suspenders guard against a runaway symlink loop.
+	cwd, err := os.Getwd()
```
FranDias, line 5077: **🟡 SUGGESTION:** The walk uses `os.Getwd()` — i.e. the broker process's cwd, on the assumption it equals the user's cwd. That holds today because the broker is started by the wuphf CLI and inherits cwd, but a future entry point that runs the broker as a long-lived daemon (`launchctl`, `systemd`, a sidecar container) would put the broker at `/`, which would silently break local-install detection for everyone using that mode. Either add a one-line guardrail (`if cwd == "/" || cwd == os.Getenv("HOME") { return upgradeInstallPlan{Method: "unknown"} }`) or extend the comment block above to call out the assumption so a future daemonization PR knows to plumb a per-request `working_dir` through.

FranDias (commented):
Re-review of click-to-run upgrade banner + content-aware show gate. Verified the prior-round fixes: single-flight via atomic.Bool with correct CompareAndSwap+defer Store, broker-owned runCtx so client cancellation doesn't abort npm mid-extract, UTF-8-safe truncateForJSON (rounds past continuation bytes), break instead of unbounded walk, dismiss disabled for major bumps, decideShow extracted with `done`-phase deferring to the idle matrix, and client_failure path correctly distinguishing transport failure from detection-unknown copy. Notable/IsMajorBump Go funcs and TS twins agree and have unit tests. Remaining concerns: (a) the local-install detection's break also fires on declares-but-not-materialized package.json, which silently demotes fresh-clone monorepos to 'unknown' — the comment only documents the not-declaring case; (b) `runCtx` is detached from broker stopCh, so npm can run for up to 120s after broker shutdown and a `wuphf restart` could race two npms against the same prefix; (c) `MaxBytesReader` wrapper is dead code since the body is never read — the comment claims it caps server-side buffering but Go's net/http doesn't pre-buffer unread bodies; (d) cross-file VERSION_RE / Notable / IsMajorBump triplet has no parity test (acknowledged out-of-scope in PR description); (e) `truncateForJSON(s, cap)` actually returns up to `cap + len(sentinel)` bytes — fine at 64 KiB cap but the parameter naming is misleading. Heads-up: PR is in CONFLICTING state vs main per `gh pr view` — rebase before merge.

[No local checkout — diff-only review against PR head 1f7439205b.]

---
Findings: 🔴 0 critical · 🟠 3 important · 🟡 4 suggestions

internal/upgradecheck/upgradecheck.go
```diff
@@ -265,6 +265,39 @@ func extractTrailingPR(s string) string {
...
+	}
+	return false
+}
+
+// IsMajorBump reports whether the first dotted-numeric segment of `to` is
+// strictly greater than that of `from`. Used by the banner to force-show
+// across a major version line — a rule we apply manually since auto-release
+// only knows feat-vs-not (no major bump on conventional-commit `!` markers
+// today). Mirrored in web/src/components/layout/upgradeBanner.utils.ts.
+func IsMajorBump(from, to string) bool {
```
FranDias, line 330: **🟡 SUGGESTION:** `Notable`, `IsMajorBump`, and the version regex now exist in three places (this file, `internal/team/broker.go`'s `upgradeVersionParam`, and `web/src/components/layout/upgradeBanner.utils.ts`). The PR explicitly punts a parity test as 'pre-existing drift surface', but this PR adds two more functions to the drift list. Cheap mitigation: have a Go test golden-file the rule outputs (or the regex literal) into a fixture that the TS test reads, so a unilateral tightening on either side fails CI on the other. Without this, the next 'just tweak the regex' PR will silently desync the banner from the broker.

internal/team/broker.go
```diff
@@ -4728,6 +4751,278 @@ func (b *Broker) handleUpgradeChangelog(w http.ResponseWriter, r *http.Request)
...
+	if len(s) <= cap {
+		return s
+	}
+	start := len(s) - cap
+	for start < len(s) && s[start]&0xC0 == 0x80 {
+		// 10xxxxxx: continuation byte. Skip until we land on a
+		// leading byte (0xxxxxxx, 110xxxxx, 1110xxxx, 11110xxx).
+		start++
+	}
+	return "…[truncated]…\n" + s[start:]
```
FranDias: **🟡 SUGGESTION:** `truncateForJSON(s, cap)` doesn't actually bound the result to `cap` bytes — it prepends a literal sentinel (`…[truncated]…\n`, 18 bytes UTF-8) and keeps `cap` trailing bytes, so the output is up to `cap + 18`. Not a real risk at the production cap of 64 KiB, and the test at `cap=5` correctly returns 18+5 bytes — but the parameter naming is misleading and a future caller picking a tight cap (a structured-log line budget, a fixed-size ring buffer) would miss this. Either rename the parameter to `tailBytes` to signal it's the suffix size, or budget the sentinel inside the cap (`if cap > len(sentinel) { start := len(s) - (cap - len(sentinel)) ... }`) so the function name matches its semantics.

FranDias:
## Round of review fixes — `c869b993` on `feat/upgrade-banner-click-and-gate`

Rebased onto `main` (resolved conflicts in `UpgradeBanner.tsx`, `upgradecheck_test.go` — kept main's a11y `role="region"` + my branch's `bannerClass`/`run` state, kept main's `changelogFetchedRef` latch alongside my branch's eager fetch + up-to-date short-circuit). Then applied:

### Staff review (self-flagged) — `internal/team/broker.go`

- **`handleUpgradeRun`: couple `runCtx` to `b.stopCh`.** Without this, a `wuphf restart` immediately after install-click left the OLD broker's npm running for up to 120s while the NEW broker spawned its own npm against the same prefix — exactly the half-extract race the single-flight gate is meant to prevent. Now `Stop()` cancels the wait.
- **`handleUpgradeRun`: actually consume `r.Body`.** The `MaxBytesReader` wrapper without a read attempt was a no-op (net/http buffers headers but reads bodies on demand). `io.Copy(io.Discard, r.Body)` forces the cap to fire; surfaces a 413 if exceeded.
- **`detectUpgradeInstall`: document daemonization assumption.** `os.Getwd()` == user's cwd holds today; a future systemd / launchctl deploy would silently break detection. Called out explicitly.
- **`detectUpgradeInstall`: rewrite the "stop walking" comment** to cover BOTH paths that reach the break (no-wuphf-declared AND declared-but-no-`node_modules`-entry), not just the first.
- **`truncateForJSON`: rename `cap` → `tailBytes`** and document that the return is up to `tailBytes+18` bytes (sentinel size). The old name misled callers picking a tight log-line budget.

### CodeRabbit Major

- **`broker.go`: set `InstallMethod="unknown"` on single-flight rejection.** Wire shape now stays inside `"global"|"local"|"unknown"` instead of `""`.
- **`broker_upgrade_test.go`: replace spin-wait with `entered` channel.** A regression in single-flight that lets a second request reach `runUpgradeCmdFn` now fails the test deterministically (only the first invocation parks on `release`; second returns "unexpected second spawn") instead of deadlocking until the suite timeout.
- **`web/api/client.ts`: `install_method` is optional.** The client synthesises a transport-failure result without lying about what the broker reported.
- **`UpgradeBanner.tsx` (`triggerRun` catch):** drop `install_method:"unknown"` on transport failure — it triggered the broker-side "couldn't detect install" copy, which is wrong for a network/timeout error. Pass `command` so the user has a copyable fallback.
- **`UpgradeBanner.tsx` (success outcome):** copy + button label match what the click actually does. The button only reloads the browser; it does NOT restart wuphf. New copy: "Reload this page after restarting wuphf." / "Reload page".
- **`UpgradeBanner.tsx` (dismiss):** disabled while `phase==="running"` so the user can't silence `latest` mid-install — `phase->"done"` then triggered `decideShow()` to unmount the banner before the outcome was visible.

### CodeRabbit Minor

- **`truncateForJSON`: defensive guard for `tailBytes <= 0`** (returns sentinel rather than panicking on a negative slice index).

### Skipped / not addressed

- `broker_upgrade_test.go:666` (assert `WorkingDir` in root-owned monorepo case) — that test case isn't in the file on the current branch tip; the comment was against a stale revision.
- `UpgradeBanner.test.tsx:200, :259` — that file doesn't exist on the current branch tip (the unit-test file was elsewhere); MSW/RTL coverage is on the existing `upgradeBanner.utils.test.ts`.
- `upgradecheck.go:295` parity-test for `Notable` / `IsMajorBump` / version regex (drift surface across Go broker + Go upgradecheck + TS web utils) — left as documented follow-up; not a regression in this PR.
- `broker.go:4859` ancestor-walk for nested workspaces — kept the documented "refuse rather than mis-install" trade-off; would be a separate PR with its own test matrix.

PR #372, opened 2026-04-28
feat(skills): wiki-skill-compile pipeline

najmuzzaman-mohammad:
## Summary

Adds wiki-driven + notebook-signal skill compilation to WUPHF, ending the era of agent-vibes skill creation. Click "Compile" on /skills → LLM scans wiki articles + notebook clusters → proposes skills → human approves → agents invoke them as real tasks.

Design doc: `~/.gstack/projects/nex-crm-wuphf/najmuzzaman-feat+wiki-skill-compile-design-20260427-214758.md` (eng-reviewed + design-reviewed + Codex outside-voice). 8 commits below.

## What ships

- **Stage A (LLM-gated wiki scanner)** — walks `team/**/*.md` (excluding `team/skills/`, notebooks, compiled), per-article asks "is this a reusable skill?" via real LLM call (Anthropic/OpenAI provider with graceful fallback to explicit-frontmatter fast path).
- **Stage B (notebook + self-heal signal scanners + LLM synthesizer)** — clusters cross-agent notebook entries by Jaccard similarity, scans resolved self-heal incidents, emits SkillCandidate events into the LLM synthesizer.
- **Unified write helper `b.writeSkillProposalLocked`** — single funnel for ALL skill writes. Lock-release-during-Enqueue pattern fixes the deadlock-trap Codex caught.
- **Skill safety guard** — Hermes-ported 3-tier verdict (safe/caution/dangerous) × trust ladder. Dangerous content NEVER hits disk.
- **Full CRUD MCP surface** — `team_skill_compile`, `team_skill_patch`, `team_skill_edit`, `team_skill_archive`, `team_skill_write_file`.
- **Approve/Reject UI + 5s undo toast** — sectioned layout (Pending / Active / Archived), AI-suggested pill, clickable provenance link, status badges.
- **Invoke creates skill_run task** — clicking Invoke creates an in-progress task an agent picks up and executes.
- **Security gate** — `handleInvokeSkill` rejects non-active skills with HTTP 403 (Codex T3 fix).
- **Cron + event-driven trigger + cooldown** — `WUPHF_SKILL_COMPILE_INTERVAL` (30m default) + on-wiki-write event + cooldown.
- **Telemetry** — full counter set exposed via `GET /skills/compile/stats`.
- **Wiki catalog hygiene** — excludes `team/skills/` + dot-prefixed files. Fixes "Loading article..." stuck state.
[...]

PR #374, opened 2026-04-28
fix(skills): close LLM/tombstone/state-replay gaps from PR #372

najmuzzaman-mohammad:
## Summary

Three follow-up fixes for gaps identified after PR #372 landed. Stacked on `feat/wiki-skill-compile` (PR #372).

**Fix 1 — Real LLM provider wiring (`skill_scanner.go`, `skill_synth_provider.go`)**

- Replaced the no-op stubs in `defaultLLMProvider.AskIsSkill` and `defaultStageBLLMProvider.SynthesizeSkill` with live calls through `provider.RunConfiguredOneShot` (the same shell-out path used by entity synthesis).
- Graceful fallback: `AskIsSkill` logs and returns `is_skill=false` on failure so the scan pass continues; `SynthesizeSkill` propagates errors so the caller skips the candidate with a clean log line.
- The explicit-frontmatter fast path in `AskIsSkill` is preserved.
- Adds `WUPHF_SKILL_LLM_TIMEOUT` env var (default 30s) via `skillLLMTimeout()`, shared by both providers.

**Fix 2 — Thread `source_article` through reject tombstone (`skill_crud_endpoints.go`)**

- `handleSkillReject` previously left `SkillTombstoneEntry.SourceArticle` empty because `teamSkill` does not carry source article paths directly.
- Fix: reads the on-disk SKILL.md before appending the tombstone and extracts `fm.Metadata.Wuphf.SourceArticles[0]` (or first `SourceSignals` entry as fallback).
- Future scan passes that re-encounter the same source article will now be correctly gated by the tombstone.

**Fix 3 — Archived skill status reverting to active on restart (`broker.go`, `skill_crud_endpoints.go`)**

- Root cause: `handleSkillArchive` updated `b.skills` and enqueued the wiki write but did not call `saveLocked`. A restart before the next natural save reloaded the stale `status=active` from broker-state.json.
- Fix A: call `saveLocked` inside `handleSkillArchive` after the wiki enqueue.
- Fix B: add `reconcileSkillStatusFromDisk`, called once from `StartWiki` after the wiki worker is wired. Reads each skill's on-disk SKILL.md and corrects any status mismatch — defense-in-depth for any other handler that might still have a race window.
- Regression test: `TestSkillArchiveStatusSurvivesRestart` archives a skill via HTTP, calls `reloadedBroker`, asserts `status=archived` survives the reload.

## Test plan

- [ ] `go build ./...` — clean
- [ ] `go vet ./...` — clean
- [ ] `go test -race ./internal/team/ -run 'Skill|Synthesize|StageB' -count=1` — all pass
- [ ] Manual: set `ANTHROPIC_API_KEY` or `OPENAI_API_KEY`, run a scan pass, verify proposals are generated from non-frontmatter articles
- [ ] Manual: reject a proposed skill, restart broker, verify skill does not re-appear in the list

### DIFF
--- a/internal/team/broker.go
+++ b/internal/team/broker.go
@@ -559,6 +559,23 @@ type Broker struct {
 	stopCh   chan struct{} // closed by Stop(); signals background goroutines to exit
 	stopOnce sync.Once
 
+	// Skill compile (Stage A) plumbing. The scanner is lazily constructed on
+	// first compile; metrics + flags coordinate concurrent triggers. All four
+	// fields are guarded by b.mu except where the metric body uses sync/atomic.
+	skillCompileMetrics   SkillCompileMetrics
+	skillCompileInflight  bool
+	skillCompileCoalesced bool
+	skillScanner          *SkillScanner
+	// Skill synthesizer (Stage B) plumbing. Same coalesce semantics as
+	// Stage A; metrics + counters live on skillCompileMetrics.StageBProposalsTotal.
+	skillSynthesizer    *SkillSynthesizer
+	skillSynthInflight  bool
+	skillSynthCoalesced bool
+	// recentlyRejectedSkills holds in-memory snapshots of skills rejected in
+	// the last 60s so /skills/reject/undo can restore them. Keyed by undo
+	// token. Guarded by b.mu. See skill_crud_endpoints.go for GC semantics.
+	recentlyRejectedSkills map[string]rejectedSkillSnapshot
+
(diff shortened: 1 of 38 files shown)


najmuzzaman-mohammad:
Closing as duplicate of #372. The fix commits (3 of them) are already on the PR #372 branch (feat/wiki-skill-compile) — bundled into one ship-ready PR.

PR #386, opened 2026-04-29
feat(team): bring /connect Telegram flow into the web

FranDias:
## Summary

- Closes the gap that prevents users from finishing the Telegram bridge in the web office: `/connect` was TUI-only, so a saved bot token + a created group still had no path to a bridged channel.
- Adds three broker endpoints (`POST /telegram/{verify,discover,connect}`) that mirror the TUI's `team.VerifyBot` / `DiscoverGroups` / channel-creation flow, and a web wizard that drives them.
- The wizard walks the user through token entry → bot verification → group discovery → pick (with retry / manual chat ID / DM fallbacks) → channel creation, all without leaving the office.

## What's in the PR

**Backend**
- `internal/team/broker_telegram_connect.go` — new file, three handlers + `createTelegramChannel` helper that inserts the channel into the broker's in-memory state, persists, and publishes a `channel_created` event.
- `internal/team/broker.go` — registers the three new routes.
- `internal/commands/slash.go` — registers `/connect` with `WebSupported: true` so the autocomplete picks it up.

**Web**
- `web/src/components/integrations/TelegramConnectModal.tsx` — wizard with a 7-state machine (`token → verifying → discovering → pick → manual → connecting → done`). Failures keep the user on the same step with the input editable. Every step + control carries a `data-testid` so the e2e suite can drive it without depending on copy.
- `web/src/api/client.ts` — typed `verifyTelegramBot`, `discoverTelegramChats`, `connectTelegramChannel`.
- `web/src/components/messages/Composer.tsx` — wires `/connect` (and `/connect telegram`) to open the modal; rejects unknown integrations with a friendly notice.
- `web/src/App.tsx` — mounts `TelegramConnectHost`.
- `web/src/stores/app.ts` — adds `telegramConnectOpen` flag.
- `web/src/hooks/useCommands.ts` — adds `/connect` to the autocomplete fallback + icon map.

**E2E**
- `web/e2e/tests/telegram-connect.spec.ts` — 7 checkpoints covering every wizard transition, stubbing the broker endpoints with `page.route()` so the suite never hits the live Telegram API.
- `web/e2e/run-local.sh` — adds the new spec to the shell phase.

## Test plan

- [x] `go build ./...` — clean
- [x] `bunx tsc -p tsconfig.app.json` — zero non-test errors
- [x] `bash scripts/test-go.sh ./internal/team ./internal/commands` — both packages green
- [x] `PORT=37891 bash web/e2e/run-local.sh shell` — all 18 shell-phase tests pass (10 pre-existing + 8 new), including every Telegram wizard checkpoint
- [ ] Smoke against a real Telegram bot (requires a tester with @BotFather access) — recommend doing this once before flipping the feature on for everyone

## Follow-ups (not in this PR)

- README's `/connect` section is now accurate for both web and TUI; can stay as-is.
- The web-only welcome message + manifest sync mirror what the TUI does, but the TUI removes `~/.wuphf/team/broker-state.json` after creating the channel to force a re-read on next boot. The web flow doesn't because we're already inside the broker process and the in-memory state has been updated. Worth a follow-up to harmonize this if we ever see a drift bug.

## Summary by CodeRabbit

* **New Features**
  * Added `/connect` slash command, a Telegram connect wizard (provider → token → discover → pick/manual → done), global host, and composer integration to open the wizard.

* **API / Client**
  * New endpoints for bot verify, chat discovery, and connect with optional request cancellation.

* **Behavior / Fixes**
  * Stricter channel creation validation, reserved "ceo" slug, consistent Telegram title→slug normalization, DM/manual chat‑ID flows, and more reliable manifest updates.

* **Tests**
  * Playwright E2E for the wizard and new broker/manifest regression tests.

* **Style**
  * Wiki CSS tokens promoted to :root for modal/editor styling.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,6 +207,10 @@ jobs:
         env:
           BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before }}
         run: bash scripts/check-no-new-sleeps-in-tests.sh
+      - name: Complexity ignore baseline
+        # Keep Biome complexity suppressions explicit during the migration:
+        # stale entries are reported, new ignores must update the baseline.
+        run: bash scripts/check-complexity-baseline.sh
       - name: Reject unix-only exec.Command shapes outside tmux/tui paths
         # Windows guard: bash/sh/chmod/ln/mkfifo/killall don't exist there.
         # Code that needs them must be in a _unix.go file with a build tag,
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -139,3 +139,8 @@ pre-push:
       # patterns. Cheap to run (git ls-files + wc -l), so it lives in
       # pre-push alongside the other smoke checks.
       run: bash scripts/check-file-size.sh
+    complexity-baseline:
+      # Keep Biome complexity suppressions forward-only. This lets the
+      # migration baseline land while forcing future broad complexity ignores
+      # to show up as explicit debt diffs.
+      run: bash scripts/check-complexity-baseline.sh
--- a/scripts/check-complexity-baseline.sh
+++ b/scripts/check-complexity-baseline.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+# scripts/check-complexity-baseline.sh — forward-only baseline for Biome
+# complexity suppressions.
+#
+# The baseline lets migration PRs land existing complexity debt without making
+# future `biome-ignore lint/complexity/...` additions invisible. Removing an
+# ignore is encouraged and reported as stale baseline drift; adding one requires
+# an explicit scripts/complexity-baseline.txt diff.
+
+set -euo pipefail
+
+script_dir="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
+repo_root="$(cd -- "$script_dir/.." && pwd)"
+baseline="$repo_root/scripts/complexity-baseline.txt"
+
+current_f="$(mktemp)"
+baseline_norm="$(mktemp)"
+trap 'rm -f "$current_f" "$baseline_norm"' EXIT
+
+collect_current() {
+  { git -C "$repo_root" grep -nE 'biome-ignore(-all)?[[:space:]]+lint/complexity/' -- 'web/src' || true; } |
+    while IFS=: read -r rel_path _line_no text; do
+      [[ -n "${rel_path:-}" ]] || continue
+      rule=$(
+        printf '%s\n' "$text" |
+          sed -E 's/.*biome-ignore(-all)?[[:space:]]+(lint\/complexity\/[^:[:space:]}]+).*/\2/'
+      )
+      [[ "$rule" == lint/complexity/* ]] || continue
+      reason=$(
+        printf '%s\n' "$text" |
+          sed -E 's/.*biome-ignore(-all)?[[:space:]]+lint\/complexity\/[^:]+:[[:space:]]*//; s/[[:space:]]*\*\/\}?$//; s/[[:space:]]*\}$//; s/[[:space:]]+$//'
+      )
+      printf '%s|%s|%s\n' "$rel_path" "$rule" "$reason"
+    done |
+    awk -F '|' '{ key = $0; seen[key]++; printf "%s|%s|%d|%s\n", $1, $2, seen[key], $3 }'
+}
+
+if [[ "${1:-}" == "--print-current" ]]; then
+  collect_current
+  exit 0
+fi
+
+collect_current > "$current_f"
+
+if [[ -f "$baseline" ]]; then
+  sed -E '/^[[:space:]]*(#|$)/d; s/[[:space:]]+$//' "$baseline" > "$baseline_norm"
+else
+  : > "$baseline_norm"
+fi
+
+new_entries=$(sort "$current_f" | comm -23 - <(sort "$baseline_norm") || true)
+stale_entries=$(sort "$baseline_norm" | comm -23 - <(sort "$current_f") || true)
+
+if [[ -n "$stale_entries" ]]; then
+  echo "::group::stale complexity baseline entries"
+  printf '%s\n' "$stale_entries" | sed 's/^/  /'
+  echo "  -> remove these from scripts/complexity-baseline.txt"
+  echo "::endgroup::"
+fi
+
+if [[ -n "$new_entries" ]]; then
+  echo "::error::new Biome complexity ignores outside scripts/complexity-baseline.txt:"
+  printf '%s\n' "$new_entries" | sed 's/^/  /'
+  echo
+  echo "Either refactor the new complexity, or add a justified baseline entry."
+  exit 1
+fi
+
+count=$(wc -l < "$current_f" | tr -d ' ')
+echo "complexity baseline check OK ($count tracked complexity ignores)"
--- a/scripts/complexity-baseline.txt
+++ b/scripts/complexity-baseline.txt
@@ -1,0 +1,75 @@
+# Biome complexity suppression baseline.
+#
+# Format:
+#   path|rule|occurrence|reason
+#
+# The occurrence is per identical path+rule+reason tuple so duplicated scoped
+# suppressions in one file remain tracked without tying the baseline to line
+# numbers. This file is forward-only during the migration: entries can become
+# stale and be removed, but new entries require an explicit debt diff.
+
+web/src/api/wiki.test.ts|lint/complexity/noExcessiveLinesPerFunction|1|Existing function length is baselined for a focused follow-up refactor.
+web/src/components/agents/AgentPanel.tsx|lint/complexity/noExcessiveCognitiveComplexity|1|Existing cognitive complexity is baselined for a focused follow-up refactor.
+web/src/components/agents/AgentWizard.tsx|lint/complexity/noExcessiveCognitiveComplexity|1|Existing cognitive complexity is baselined for a focused follow-up refactor.
+web/src/components/apps/ArtifactsApp.tsx|lint/complexity/noExcessiveLinesPerFunction|1|Existing function length is baselined for a focused follow-up refactor.
+web/src/components/apps/GraphApp.tsx|lint/complexity/noExcessiveCognitiveComplexity|1|Existing cognitive complexity is baselined for a focused follow-up refactor.
+web/src/components/apps/GraphApp.tsx|lint/complexity/noExcessiveLinesPerFunction|1|Existing function length is baselined for a focused follow-up refactor.
+web/src/components/apps/GraphApp.tsx|lint/complexity/noExcessiveCognitiveComplexity|2|Existing cognitive complexity is baselined for a focused follow-up refactor.
+web/src/components/apps/SettingsApp.imageGen.tsx|lint/complexity/noExcessiveCognitiveComplexity|1|Existing cognitive complexity is baselined for a focused follow-up refactor.
+web/src/components/apps/SettingsApp.tsx|lint/complexity/noExcessiveCognitiveComplexity|1|Existing cognitive complexity is baselined for a focused follow-up refactor.
+web/src/components/apps/SkillsApp.tsx|lint/complexity/noExcessiveCognitiveComplexity|1|Existing cognitive complexity is baselined for a focused follow-up refactor.
+web/src/components/apps/SkillsApp.tsx|lint/complexity/noExcessiveLinesPerFunction|1|Existing function length is baselined for a focused follow-up refactor.
+web/src/components/apps/SkillsApp.tsx|lint/complexity/noExcessiveCognitiveComplexity|2|Existing cognitive complexity is baselined for a focused follow-up refactor.
+web/src/components/apps/SkillsApp.tsx|lint/complexity/noExcessiveCognitiveComplexity|3|Existing cognitive complexity is baselined for a focused follow-up refactor.
+web/src/components/apps/SystemSchedulesPanel.tsx|lint/complexity/noExcessiveCognitiveComplexity|1|Existing cognitive complexity is baselined for a focused follow-up refactor.
+web/src/components/apps/SystemSchedulesPanel.tsx|lint/complexity/noExcessiveCognitiveComplexity|2|Existing cognitive complexity is baselined for a focused follow-up refactor.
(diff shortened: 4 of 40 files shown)


FranDias:
Round-trip on the staff review (commit `78a21ab8`). All IMPORTANT findings addressed; SUGGESTION findings addressed except the one explicitly asking for a deferred decision.

**Addressed**

| Finding | Fix |
|---|---|
| `createTelegramChannel` skipped member validation, reserved-slug guard, creator self-add | Extracted `createChannelLocked` from `handleChannels` (broker.go); web flow now routes through it. Both call sites share one validated helper. |
| Member-set diverged from TUI when `lead != "ceo"` | Dropped hardcoded `"ceo"` in the web path; `createChannelLocked` prepends it. Mirrors TUI exactly. |
| Token persisted in `connect`, not `verify` | Moved `os.Setenv` + `config.SaveTelegramBotToken` into `handleTelegramVerify` after a successful `VerifyBot`. Matches TUI ordering — verify-then-abandon now leaves a usable token in config. |
| Bare `/connect` silently shortcut to Telegram | Added a `provider` step to the wizard. Bare `/connect` lands on a 4-option picker (Telegram / OpenClaw / Slack / Discord) matching the TUI's `cmd/wuphf/channel.go:4871-4880`. OpenClaw closes the modal with a TUI pointer; Slack/Discord are disabled "coming soon". `/connect telegram` still skips the picker. |
| `writeTelegramJSON` duplicated `writeJSON` | Deleted the shadow; the package now has one `writeJSON`. |
| `slugifyGroupTitle` (TUI) and `slugifyTelegramTitle` (web) duplicated | Moved into `internal/team/telegram.go` as `SlugifyTelegramTitle`. TUI and broker both call it. |
| `syncManifestForTelegramChannel` swallowed `SaveManifest` errors | Logs them via `log.Printf` so a disk-full / permission bug surfaces in the broker log. Send-message failures from the welcome path log too. |
| `expectNoReactErrors` only in test [1] | Hoisted to `beforeEach`/`afterEach`. Every test exercises the React-error guard. |
| Missing test for verify ok:false → retry success | New checkpoint `[11]`: types bad token → 401 banner → types good token → advances to picker, asserts no stale error banner. |
| `chat_id: number` round-trip not documented | Comment on `TelegramGroup` calls out Telegram's 52-bit cap vs JS's 53-bit `MAX_SAFE_INTEGER` and the bigint-migration trigger. |

**New e2e checkpoints**

- `[1a]` Bare `/connect` opens the provider picker; all four TUI options visible; Slack/Discord disabled.
- `[1b]` Picker → Telegram advances to the token step.
- `[1c]` Picker → OpenClaw closes the modal (regression guard against the "silent shortcut to Telegram" bug bornaware originally hit).
- `[11]` Verify-fail then retry clears the error and proceeds.

22/22 shell-phase tests green via `PORT=37891 bash web/e2e/run-local.sh shell`.

**Deferred**

The "approach scalability" finding (generic `/connect/<provider>/{verify,discover,connect}` shape vs the current Telegram-specific endpoints) is real — the next integration with a web wizard will either copy this file or force a refactor. I'd rather make that call when the second wizard is being designed (probably OpenClaw, since the OpenClaw connect flow has different shape: URL + token + sessions list, not token + chats). Tracking it as a follow-up rather than guessing the right abstraction now.

The "TUI rms `~/.wuphf/team/broker-state.json` after connect, web doesn't" asymmetry is also deferred — the in-memory channel + manifest are both written, and the `os.Remove` in the TUI is itself a workaround for an older sync-on-restart issue rather than a fix the web needs. If we see drift in practice we'll bring both paths to the same handling.

FranDias:
## Local verification — `/connect telegram` wizard end-to-end

Ran this branch (`78a21ab8` at the time) against a sandboxed wuphf, drove the full wizard with stubbed broker routes (same `page.route()` pattern as `telegram-connect.spec.ts`), and captured screenshots of every step.

**While doing this I hit a UX bug worth flagging — pushed a fix as `4b8bd2c1` on this branch.**

### Bug: modal renders unstyled outside the wiki page

The `TelegramConnectModal` reuses `wk-modal-*` / `wk-editor-*` primitives from `wiki.css`. Those rules use CSS custom properties (`--wk-paper`, `--wk-border`, …) that were declared inside `.wiki-root`. Since the wizard is mounted globally via `TelegramConnectHost` and triggered from `#general` (or anywhere outside the wiki), the variables resolved to defaults — the modal rendered with **transparent background, zero-width border, no card surface**: heading and buttons floating on top of the underlying view, page text bleeding through.

Discriminating evidence (from `getComputedStyle` on `.wk-modal` against `78a21ab8`):

```
cardBg:          rgba(0, 0, 0, 0)   ← should be opaque
cardBorderWidth: 0px                 ← should be 1px
```

The current spec (visibility-only `data-testid` checks) couldn't catch this — the testids were present, just unstyled.

### Fix shipped in `4b8bd2c1`

- Promote `--wk-*` aliases from `.wiki-root` to `:root` in `web/src/styles/wiki.css`. Each is already an alias of a global token (`--bg`, `--border`, …), so wiki page rendering is unchanged — the aliases just resolve everywhere.
- Import `wiki.css` directly from `TelegramConnectModal.tsx` as a self-documenting guard against a future refactor lazy-loading Wiki.
- Add two regression specs (`[10b]`, `[10c]`) that assert computed `backgroundColor`, `borderWidth`, and `--wk-paper` resolution on the modal. Verified discriminating: reverting the CSS fix made both fail; restoring it made all 14 telegram-connect tests pass + 10 smoke/local-llm-settings tests still green.

### Screenshots (post-fix)

Hosted on the `screenshots/pr386` branch in this repo.

#### 0. Provider picker (bare `/connect`)
provider-picker

#### 1. Token entry (`/connect telegram`)
token-entry

#### 2. Token validation (empty submit)
token-validation

#### 3. Verify failure (401)
verify-fail

#### 4. Group picker (verify success)
pick-groups

#### 5. Done — channel created
done

#### 6. Empty group list (retry / manual / DM controls)
pick-empty

#### 7. Manual chat ID step
manual

#### 8. Manual ID validation
manual-validation

#### 9. Connect failure (stays on picker with banner)
connect-fail

### Other paths verified (all passed)

- `[1]` `/connect telegram` → token entry
- `[1a–1c]` bare `/connect` → provider picker (Telegram / OpenClaw / Slack / Discord) with proper disabled states + OpenClaw closes-with-toast
- `[2,3]` empty token → inline validation, no broker call
- `[4]` verify failure → stays on token step with editable input + 401 banner
- `[5,7]` verify success → group picker → connect → done with `tg-test-group-a` slug
- `[6]` empty group list → retry / manual / DM controls all visible
- `[8]` manual chat ID → integer validation + posts numeric `chat_id`
- `[9]` "Use as DM" → posts `chat_id=0`, `type=private`
- `[10]` connect failure → stays on picker with error banner
- `[11]` stale-error clearing on retry

### What this wasn't tested against

- No real Telegram Bot API call — broker routes were all stubbed (matches the existing spec strategy). Recommend the smoke against a real bot you flagged in the PR description before flipping the feature on for everyone.

FranDias:
Addressed all 10 CodeRabbit comments in `98a3ff80`.

| # | Severity | File / Line | Comment | Fix |
|---|----------|-------------|---------|-----|
| 1 | 🔴 Critical | `broker_telegram_connect.go:161` | Title collision treated as success | `createTelegramChannel` compares `existing.Surface.RemoteID` against the requested `chat_id` — same id is idempotent, different id returns 409 "already bridges chat &lt;id&gt;". Handler maps to `StatusConflict`. |
| 2 | 🟠 Major  | `broker_telegram_connect.go:265` | Manifest read-modify-write unprotected | New package-level `manifestMu sync.Mutex` wraps the load → mutate → save block in `syncManifestForTelegramChannel`. |
| 3 | 🟠 Major  | `broker_telegram_connect.go:78` | `DiscoverGroups` error swallowed | Error now surfaces as 502 with `"could not discover groups: <err>"`. Wizard renders the body inside the picker-step error banner. |
| 4 | 🟠 Major  | `broker.go:8003` | Empty slugs normalized to `"general"` before validation | Validate raw `in.Slug` (after `TrimSpace`) before calling `normalizeChannelSlug`. |
| 5 | 🟠 Major  | `broker.go:8005` | `ceo` not in `reservedChannelSlugs` | Added with a comment pointing back to `canAccessChannelLocked`. Test list extended. |
| 6 | 🟠 Major  | `TelegramConnectModal.tsx:137` | Stale async after close/retry | `requestIdRef` counter incremented before each `verifyAndDiscover()` / `connect()`; every state write after an `await` checks the captured id matches `requestIdRef.current`. `reset()` bumps the ref so prior in-flight requests are ignored on next open. |
| 7 | 🟠 Major  | `TelegramConnectModal.tsx:378` | `inputMode="numeric"` blocks negative IDs on touch | Switched to `inputMode="text"` + `pattern="-?[0-9]+"`. |
| 8 | 🟠 Major  | `TelegramConnectModal.tsx:402` | `Number.parseInt` accepts trailing garbage | Strict `/^-?\d+$/` regex + `Number()` + `Number.isSafeInteger` (also catches over-2^53 ids that would silently round). |
| 9 | 🟡 Minor  | `app.ts:337` | Wizard not closed in onboarding reset | `telegramConnectOpen: false` added to `resetForOnboarding`. |
| 10 | 🟡 Minor | `wiki.css:29` | `Georgia` casing | Lowercased. |

### New regression tests

Each is discriminating (was failing on the parent commit, passes after fix):

- `internal/team/broker_test.go::TestChannelCreateRejectsEmptySlug` — exercises the slug-validation reordering.
- `internal/team/broker_test.go::TestCreateTelegramChannelRejectsTitleCollision` — same title + different chat id ⇒ "already bridges"; same title + same id ⇒ idempotent.
- `internal/team/broker_test.go::TestChannelCreateRejectsReservedSlugs` — `ceo` added to the reserved-slug case list.
- `web/e2e/tests/telegram-connect.spec.ts` checkpoint `[8]` — adds the `"-12abc"` case so permissive parseInt regressions are caught.

### Verification

- `go test ./internal/team ./internal/commands` — green
- `cd web/e2e && bunx playwright test tests/telegram-connect.spec.ts` — 14/14 green (including new test cases)
- `cd web/e2e && bunx playwright test tests/smoke.spec.ts tests/local-llm-settings.spec.ts` — green
- `cd web && bunx tsc -p tsconfig.app.json` — zero non-test errors (the WikiSidebar/WikiLint test-file errors are pre-existing and unrelated)

FranDias:
Round 2 (`d7845e13`) + round 3 (`59944a98`):

**`d7845e13` — staff-review findings**
- Sentinel errors (`errChannelAlreadyExists` / `errChannelAlreadyBridges`) replace string-based discrimination so the handler can't misroute a future error wording change.
- Title-collision guard tightened to also reject `Surface == nil` / non-telegram channels (silent rebind of a hand-created `#standup` would otherwise bridge a Telegram chat into a non-Telegram channel).
- `os.Setenv` removed from `handleTelegramVerify` — it never reached the live transport (token cached at construction) and `SaveTelegramBotToken` was overwriting an existing token, breaking multi-bot. Now only persists when nothing is stored.
- New `company.UpdateManifest(mutate func)` serializes load → mutate → save under a package mutex; both the web `/telegram/connect` handler and the TUI `connectTelegramGroup` route through it. Surfaces `LoadManifest` errors instead of silently falling back to `DefaultManifest`.
- `AbortController` per request in the modal so cancelling actually aborts the in-flight fetch (not just suppresses the response).
- `if (open) reset(); else reset()` no-op ternary simplified.

**`59944a98` — CodeRabbit follow-up**
- `createChannelLocked` now uses `normalizeActorSlug` for `CreatedBy` (was `normalizeChannelSlug`, which rewrites `""` → `"general"` and would silently auto-add a real office member named "general").

**New regression tests**
- `TestUpdateManifestSerializesAppends` — fires N=32 parallel updates at distinct slugs and asserts every one survives in the final file. Discriminating; passes with `-race`.
- `TestCreateTelegramChannelRejectsNonTelegramSurface` — guards the surface-nil branch.
- Existing `TestCreateTelegramChannelRejectsTitleCollision` switched from substring match to `errors.Is` on the new sentinels.

FranDias:
Round 3 — `18b76838` addresses all three staff-review findings from earlier:

- **🔴 Token persistence regression** — `handleTelegramVerify` now gates on `cfg.TelegramBotToken` (the persisted value) instead of `config.ResolveTelegramBotToken()` (which short-circuits on `WUPHF_TELEGRAM_BOT_TOKEN`). The previous guard meant any developer who exported the env to test bot A would never persist any token through the wizard.
- **🟠 Unlocked `LoadManifest` race** — new `company.SnapshotManifest()` helper takes the same `manifestUpdateMu` as `UpdateManifest`. `createTelegramChannel` now uses it, so the member list it derives is consistent with whatever `syncManifestForTelegramChannel` writes a few lines later.
- **🟠 Panic recovery in `UpdateManifest`** — round-2 wired `UpdateManifest` into HTTP handler goroutines, where an unrecovered panic from the mutate callback would crash the broker. Added `defer recover()` that re-surfaces panics as errors via a named return.

**New test:** `TestUpdateManifestRecoversFromMutatePanic` — discriminating regression that asserts panic surfaces as an error AND the lock is released so the next call proceeds.

**Staff review re-run on `59944a98..18b76838`** confirmed:
- Token-persistence fix correctly handles every env/persisted combination.
- Lock ordering audited across `broker.go` + `launcher.go` — no inversion path (manifest lock always before `b.mu`, never after).
- `SnapshotManifest` is the right shape (regular `Mutex` is sufficient; RWMutex isn't worth the complexity given the call frequency).
- Panic recovery via named return + deferred recover works correctly.

Other read-only `LoadManifest` callers in `broker.go` / `channel_integration.go` are intentionally not gated — they're best-effort reads where racing with a writer is acceptable. Out of PR scope.

CI: build / vet / lint / smoke / commitlint / shellcheck / web / scan / package-tests all green; cross-build / go-test / vhs / web-e2e / windows-smoke / release-build still running.

FranDias:
@coderabbitai review again

FranDias:
Cleanup note: this PR is closed because its head branch now points at the same commit as main (`5752af36`). The Telegram web-connect work has already landed upstream through the later stack merges, so there is no remaining diff for GitHub to merge here. The two stale unresolved CodeRabbit threads were checked against main and are already satisfied: onboarding reset closes the Telegram wizard, and manual connect failures stay on the manual step.

PR #500, opened 2026-04-30
refactor(team): split spawn orchestration off pane_lifecycle.go (C24)

FranDias:
Stacked on **#498 (C23)**. `pane_lifecycle.go` grew to 893 lines across the C5b–C5f series. It bundled three distinct concerns; splitting on the orchestration boundary gives each file a sharp single responsibility and matches the file-naming pattern used elsewhere in the package.

## Two-way split

| File | Lines | Owns |
|---|---:|---|
| `pane_lifecycle.go` | 596 (was 893) | paneLifecycleDeps struct, paneLifecycle type + constructors + withClock test seam, **23 tmux primitive methods** (HasLiveSession, ListTeamPanes, SplitFirstAgent, ApplyMainVerticalLayout, IsPaneDead, …), and pure helpers (parseAgentPaneIndices, paneFallbackMessages, etc.) |
| `pane_lifecycle_spawn.go` | 313 (new) | **7 high-level orchestration methods** that compose tmux primitives into roster-aware flows: SpawnVisibleAgents, SpawnOverflowAgents, DetectDeadPanesAfterSpawn, TrySpawnWebAgentPanes, PrimeVisibleAgents, ReportPaneFallback, CaptureDeadChannelPane |

The split follows the same logic as launcher.go's "tmux primitives stay, high-level orchestration moves" — the primitives are stable building blocks; orchestrations evolve as the spawn flow does.

## Local CI matrix (all green)

- gofmt clean / golangci-lint 0 issues / 32 packages green

PR #502, opened 2026-04-30
fix(team): close staff-review CRITICALs on the launcher decomposition (C25)

FranDias:
Stacked on **#500 (C24)**. Closes the 2 CRITICALs and 1 IMPORTANT from the staff review of the C1–C24 stack.

## What was wrong

After staff review of the 28-PR stack, two **CRITICAL** issues + 1 **IMPORTANT** were flagged:

### CRITICAL 1 — `launcher_wiring.go` regressed sync.Once + callback-failedPaneSlugs

The five lazy accessors (`targeter()`, `notifyCtx()`, `scheduler()`, `paneDispatch()`, `panes()`) used a plain check-then-act lazy-init pattern. `Launch()` spawns goroutines (notifyAgentsLoop, watchdogSchedulerLoop, pollNexNotificationsLoop) that all hit these accessors concurrently → torn writes possible.

Worse: `failedPaneSlugs: l.failedPaneSlugs` captured the map header at first call. `reconfigureVisibleAgents` does `l.failedPaneSlugs = nil`, then `recordPaneSpawnFailure` allocates a fresh map → **the cached targeter held the original map forever, silently misrouting after every reconfigure.**

### CRITICAL 2 — `lefthook.yml` `\b` regex change disabled detection of the most common real-world leak shape

C22 added `\b` to the no-secrets regex to drop a false positive on `oneSecretEnv = "ONE_SECRET=..."`. Verified by the reviewer: this **also disabled detection** of:
- `OPENAI_API_KEY = "sk-proj-..."` → silently passes
- `STRIPE_SECRET = "sk_live_..."` → silently passes
- `MY_PASSWORD = "hunter2"` → silently passes

Because Go's `\b` treats `_` as a word char, no boundary fires before SECRET in `_SECRET`. The dominant real-world leak shape was passing through.

### IMPORTANT — `failedPaneSlugs` concurrent map writes still unprotected

Pre-existing race documented in `office_targets.go:13`. Comment promised C5 would close the dangle; it didn't. Documentation drift fixed.

## What this PR does

Cherry-picked main's four fix commits onto the stack:

| Commit | Subject |
|---|---|
| `bdb7d9dd` | docs(team): correct failedPaneSlugs race comment in office_targets |
| `88d7fcfc` (was 393f71bd) | address PR review comments on launcher decomposition |
| `fbb842e3` (was 0befb1f9) | close staff-review criticals on the launcher decomposition |
| `dcc6390d` (was becfb73e) | close staff re-review follow-ups |

These commits diverged from this branch at the lazy accessor layout (which moved into `launcher_wiring.go` in C18); conflict resolution applied the **substantive fixes** to the new locations.

Plus two follow-on fixes specific to this stack:

1. **`launcher_session.go` Kill() drains scheduler** — the cherry-picked `TestLauncher_KillDrainsScheduler` regression test failed because this stack's Kill (now in launcher_session.go after C20) was missing the `schedulerWorker.Stop()` call. Mirrors main's launcher.go fix.

2. **`lefthook.yml` regex** — replaced `\b` with `(^|[^a-zA-Z0-9])` which EXCLUDES `_` from the boundary so env-var-shaped identifiers still match. Verified by piping all known leak shapes through grep:
   - ✅ `OPENAI_API_KEY = "..."` → flagged
   - ✅ `STRIPE_SECRET = "..."` → flagged
   - ✅ `MY_PASSWORD = "..."` → flagged
   - ✅ `WUPHF_API_KEY = "..."` → flagged
   - ✅ `secret = "abc"` → flagged
   - ✅ `password = "..."` → flagged
   - ✅ `oneSecretEnv = "ONE_SECRET=" + ...` → passes (false-positive dropped)
   - ✅ `apiKey = "...with shellQuote"` → passes (allowlisted)

## Substantive changes brought in (not just file-level)

From `fbb842e3`/`88d7fcfc` (cherry-picks):

- **scheduler.go**: cancelable `runCtx` so `Stop` cancels in-flight `ExecuteWorkflow` instead of pinning `done.Wait`
- **scheduler.go**: replace `startOnce`/`stopOnce` with mutex + flags so `Stop()→Start()` doesn't spawn an unstoppable goroutine
- **scheduler.go**: special-case `context.Canceled`/`DeadlineExceeded` in `processWorkflowJob` so graceful shutdown doesn't durably mark the workflow as "failed"
- **office_targets.go**: `failedPaneSlugs` reads via callback; `officeLeadSlugFrom` sorts by slug for determinism
- **office_targets.go**: `AgentOrder` and `PaneSlugs` drive off `ActiveSessionMembers` (pack-ordered) so pane indices stay stable
- **notification_context.go**: `RelevantTaskForTarget` resolves to the ultimate thread root before matching task.ThreadID
- **prompt_builder.go**: sort policies inside the builder so prompt cache stability doesn't depend on caller pre-sorting
- **launcher_wiring.go (this stack)**: 5 accessors wrapped in `sync.Once`; `failedPaneSlugs` callback wired

Plus regression tests (negative-controlled):
- `TestSchedulerStopBeforeStart_DisablesLaterStart`
- `TestTargeter_FailedPaneSlugs_FollowsLauncherReconfigure`
- `TestLauncher_KillDrainsScheduler`
- `TestNotificationContext_RelevantTaskForTarget_DeepReplyResolvesViaUltimateRoot`
- `TestTargeter_AgentOrderAndPaneSlugs_FollowPackNotSnapshot`
- `TestOfficeLeadSlugFrom_DeterministicAcrossInputOrders`
- 5 `processWorkflowJob` tests (lookup-fail, execute-success, execute-fail, canceled, deadline-exceeded)

## Local CI matrix (all green)

- gofmt clean / golangci-lint 0 issues / 32 packages green / cross-compile / smoke binary OK

## What's left after this

The launcher decomposition stack is now reviewed-and-fixed. Remaining open items the reviewer flagged but didn't block:

- **`failedPaneSlugs` mutex** — still unsynchronized; race is dormant because `trySpawnWebAgentPanes` is a runtime-promotion fallback nothing currently invokes. Closing properly requires either a mutex or a per-slug lookup function (see file header in `office_targets.go`).
- **`launcher_reconfigure.go:91` clear() vs nil-then-rebuild** — defense-in-depth follow-up, lower priority once the callback-shape failedPaneSlugs is in (this PR).
- **`pane_lifecycle.go:124` withClock doc** — could note it's strictly pre-Launch.


FranDias:
Superseded by a clean single-commit flat PR — the per-PR history caused diff churn against current main. New PR coming.

PR #527, opened 2026-04-30
refactor(settings): extract types + styles + reusable components

FranDias:
Stack base: #526.

## Summary

Phase 7 first slice. Pull shared infrastructure out of `SettingsApp.tsx` into a `settings/` subdirectory.

## Files added

| File | LOC | Purpose |
|---|---:|---|
| `settings/types.ts` | 37 | SectionId, Section, SectionGroup, SectionProps |
| `settings/constants.ts` | 46 | SECTION_GROUPS sidebar nav |
| `settings/styles.ts` | 200 | shared inline-style object |
| `settings/components.tsx` | 101 | Field, SaveButton, KeyField |

## Net LOC delta

| File | Before | After | Δ |
|---|---:|---:|---:|
| SettingsApp.tsx | 1,792 | 1,438 | **-354** |

Allowlist diff: `SettingsApp.tsx` removed (now under 1,500 LOC fail threshold). The `check-file-size.sh` stale-entry guard caught it on the first run.

## Verification

- `bunx tsc --noEmit` clean
- `bun run test` — all 565 web tests pass
- `bun run build` clean

## Next slices

- extract section panels one-by-one (or grouped as Workspace / Credentials / System / Advanced)
- promote `react-hooks/exhaustive-deps` to error after section extraction surfaces stale-closure bugs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #528, opened 2026-04-30
chore(quality): enforce file-size allowlist forward-only at CI time

FranDias:
Stack base: #527.

## Summary

Phase 9 first slice. The file-size allowlist (`scripts/file-size-allowlist.txt`) was already forward-only by tone in CONTRIBUTING.md; this PR makes the rule load-bearing at CI time.

## Files added

- `scripts/check-allowlist-forward-only.sh` (new) — extracts active path entries from base SHA and HEAD, fails if HEAD adds any new path
- New step in `.github/workflows/ci.yml` lint job

## Effect

A future PR that hits the file-size fail line and tries to fix it by adding a new allowlist entry will be **blocked at CI**. The escape route (decompose the file, or land an allowlist entry in a separate PR that earns its place via review) is the only path forward.

Companion to the existing `check-file-size.sh` stale-entry warning, which also surfaces forward-only violations from the other direction (a file shrunk below the threshold but the entry still sits in the allowlist).

## Verification

- `shellcheck scripts/check-allowlist-forward-only.sh` clean
- `BASE_SHA=e26e23f1 bash scripts/check-allowlist-forward-only.sh` returns 0 with "no new entries" — the happy path

## Cumulative Phase 9 progress

- ✅ file-size warn 800 / fail 1500 (#511)
- ✅ allowlist forward-only ratchet (this PR)

## Next slices

Each its own PR (chosen by reviewability vs. tooling availability):

- gocognit warning at threshold 30
- bundle-size budget for `bun run build`
- goleak in internal/team TestMain
- depguard import layering rules
- test-time budget gate (>10s without slow tag = fail)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,6 +162,16 @@ jobs:
         # (see CONTRIBUTING.md, docs/CODE-QUALITY.md, ADR-0001). Skips
         # generated files and tests.
         run: bash scripts/check-file-size.sh
+      - name: File-size allowlist forward-only check
+        # Companion to the file-size gate: enforces the forward-only
+        # rule at CI time. The allowlist may shrink (decomposition done)
+        # or remove entries (file deleted), but no PR may add a new line
+        # that grants exemption to a not-currently-listed file. Without
+        # this, the allowlist becomes a quiet escape hatch — every CI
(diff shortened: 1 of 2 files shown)


FranDias:
Flattened into #552.

PR #529, opened 2026-04-30
chore(quality): add bundle-size budget for the web build

FranDias:
Stack base: #528.

## Summary

Phase 9 second slice. Forward-only ratchet against accidental web bundle bloat.

## How it works

- `scripts/check-bundle-size.sh` (new): sums the size of `web/dist/assets/*.js`, fails if total exceeds `FAIL_KB`, warns if it exceeds `WARN_KB`.
- New step in the CI `web` job, runs after `bun run build`.

## Calibration

Current JS bundle is ~834 KB (single chunk). Thresholds:

| Level | KB | Headroom |
|---|---:|---:|
| WARN | 950 | ~14% |
| FAIL | 1,200 | ~44% |

To raise intentionally: edit `WARN_KB`/`FAIL_KB` in the script in the **same PR** as the regression, with justification in the commit message. Without an explicit raise, the gate catches accidental bloat (stray import-the-world, moment.js where date-fns would do, etc.).

## Why JS only

JS is what blocks the main thread on first paint — same axis the rolldown `chunkSizeWarningLimit` warning already targets. CSS and source maps are out of scope.

## Verification

- `shellcheck scripts/check-bundle-size.sh` clean
- Local dry-run after `bun run build`: "bundle-size check OK: 834 KB (warn 950, fail 1200)"

## Cumulative Phase 9 progress

- ✅ file-size warn 800 / fail 1500 (#511)
- ✅ allowlist forward-only ratchet (#528)
- ✅ bundle-size budget (this PR)

Anchored in CONTRIBUTING.md's "no performance degradations" rule.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,12 @@ jobs:
           bun install --frozen-lockfile
           bun run typecheck
           bun run build
+      - name: Bundle-size budget
+        # Forward-only ratchet against accidental bundle bloat. Warn at
+        # 950 KB, fail at 1200 KB on the JS bundle total. To raise the
+        # ceiling intentionally, edit constants in scripts/check-bundle-size.sh
+        # in the same PR. See CONTRIBUTING.md "no perf degradations" rule.
+        run: bash scripts/check-bundle-size.sh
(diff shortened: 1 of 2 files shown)


FranDias:
Flattened into #552.

PR #531, opened 2026-04-30
refactor(team): extract auth + pub/sub fanout from broker.go

FranDias:
Stack base: #530.

## Summary

Phase 3 second slice. Two small cohesive extractions from broker.go.

## Files added

**`broker_auth.go` (62 LOC)** — auth surface
- `generateToken` — cryptographically random shared secret
- `requireAuth` — Bearer-token middleware (also accepts `?token=` for EventSource)
- `handleWebToken` — localhost-only token endpoint (gated by RemoteAddr + Host loopback to defeat DNS rebinding)

**`broker_publish.go` (49 LOC)** — pub/sub fanout
- `appendMessageLocked` — persist + publish in one step
- `publishMessageLocked` / `publishActionLocked` / `publishActivityLocked` — three always-on channels with non-blocking sends (default-on-full so a slow subscriber can't stall the publisher)

## Net LOC delta

| File | Before | After | Δ |
|---|---:|---:|---:|
| broker.go | 2,859 | 2,800 | **-59** |

## Cumulative Phase 3 progress

| PR | broker.go | Δ |
|---|---:|---:|
| baseline | 3,404 | — |
| #519 (web proxy + sse) | 2,859 | -545 |
| **this** | **2,800** | **-59** |
| total | | **-604** |

Broker.go down 18%. Continued slices target: rate-limit, wiki worker lifecycle, telemetry, defaults/state-normalization, bridge handler.

## Verification

- `go build ./...` clean
- `go test ./internal/team/` passes (~82s)
- `bash scripts/check-file-size.sh` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #532, opened 2026-04-30
chore(team): scaffold goleak regression detector for internal/team

FranDias:
Stack base: #531. Phase 9 fourth slice. Wires goleak into internal/team TestMain gated by WUPHF_GOLEAK=1. See commit message for migration path.


FranDias:
Flattened into #552.

PR #544, opened 2026-04-30
refactor(team): dissolve broker_queries.go into entity files

FranDias:
Stack base: #543. Per staff-reviewer: broker_queries.go violated the package's by-entity decomposition pattern. Methods pushed back to broker_messages.go / broker_human.go / broker_tasks.go / broker_office_members.go / broker_office_channels.go / broker_dm.go. Renamed remainder to broker_queues.go (cross-entity workflow snapshots only).

PR #545, opened 2026-04-30
refactor(channel): consolidate fragmented kin files per cohesion review

FranDias:
Stack base: #544. Staff-reviewer flagged 3 problems with the channel splits — fragmented small files sharing the same call sites, false main-vs-thread dichotomy, and a lifecycle file that was just glue. Three merges: views (header + state queries + lookups), composer-input (main + thread), lifecycle folded back into channel.go.


FranDias:
Flattened into #552.

PR #546, opened 2026-04-30
refactor(channel): merge channel_client.go into channel_broker.go

FranDias:
Stack base: #545. Staff-reviewer's discoverability bug fix: channel_broker.go and channel_client.go both held HTTP calls to the same broker, but names didn't communicate the difference. Merged into channel_broker.go as the single file for every TUI→broker HTTP call.


FranDias:
Flattened into #552.

PR #552, opened 2026-04-30
refactor: flatten channel, broker, and quality stack

FranDias:
Flattened stack PR for the channel TUI decomposition, broker decomposition, and quality gates.

Supersedes: #514, #515, #516, #517, #518, #519, #520, #527, #528, #529, #530, #531, #532, #536, #541, #543, #544, #545, #546, #547, #558.

Includes:
- Channel TUI extraction through the Update router split.
- Broker decomposition into focused files for lifecycle, web proxy, publish/auth, office members, queries, helpers, provider bindings, and types.
- File-size, bundle-size, and no-new-time.Sleep quality gates.

Verification:
- bash scripts/check-file-size.sh
- bash scripts/check-no-new-sleeps-in-tests.sh
- bash scripts/test-go.sh ./cmd/wuphf
- bash scripts/test-go.sh ./internal/team
- bunx tsc --noEmit (web)
- bun run build (web)
- bash scripts/check-bundle-size.sh
- pre-push hook: build, file-size, smoke, vhs

## Summary by CodeRabbit

* **New Features**
  * Major channel UI: full keyboard & mouse controls, composer, threads, pickers, sidebar navigation, integrations, task/request workflows, lifecycle/onboarding and crash handling.
  * Web UI served locally with API proxy; Settings UI reorganized with shared settings components and sectioned navigation.

* **Tests / CI**
  * New CI gates: bundle-size budget, file-size ratchet + self-test, allowlist forward-only, and ban on new time.Sleep() in tests.
  * Added unit tests for picker and sidebar behavior; optional goroutine-leak check in tests.

* **Docs**
  * Added a “partial” status legend entry.

PR #558, opened 2026-05-01
fix(quality): enforce file-size allowlist ratchet

FranDias:
## Summary

Follow-up to #511 staff review findings.

- Enforce the file-size allowlist as a real forward-only ratchet in PR CI:
  - new allowlist entries fail against the PR base
  - existing allowlisted files fail if they grow against the PR base
- Skip documented fixture/vendor directories by path segment: `vendor/`, `node_modules/`, `dist/`, and `testdata/`.
- Run the file-size budget immediately after checkout in the lint job, with full history so the script can compare against the PR base.
- Define the `partial` status token used in `docs/surfaces.md`.

## Root Cause

#511 documented the allowlist as forward-only, but the script only checked whether an oversized file appeared in the current allowlist. A PR could add a new oversized file and a matching allowlist line, and CI would pass.

## Test Plan

- [x] `bash -n scripts/check-file-size.sh`
- [x] `shellcheck scripts/check-file-size.sh`
- [x] `bash scripts/check-file-size.sh`
- [x] Temporary local verification: adding a fake allowlist line fails with `file-size allowlist grew relative to origin/main`
- [x] `git diff --check`
- [x] pre-push hook: build, file-size, smoke, vhs

## Summary by CodeRabbit

* **Documentation**
  * Added a "partial" status to the surface support legend to clarify reduced or asymmetric support across surfaces.

* **Chores**
  * Strengthened CI file-size validation to compare changes against the PR/base reference, tightened scanning exclusions for generated/vendor outputs, and made the check fail on budget breaches or when allowlist entries or allowed-file sizes increase.

* **Tests**
  * Added a self-test that verifies the file-size check behavior using temporary fixture repositories.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,8 @@ jobs:
           # The file-size ratchet compares the current allowlist to the PR
           # base. Keep enough history for `git merge-base HEAD origin/<base>`.
           fetch-depth: 0
+      - name: File-size ratchet self-test
+        run: bash scripts/check-file-size-self-test.sh
       - name: File-size budget (CONTRIBUTING.md)
         # Warn at 800 LOC, fail at 1500 LOC unless allowlisted in
         # scripts/file-size-allowlist.txt. The allowlist is forward-only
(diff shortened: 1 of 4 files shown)


FranDias:
Flattened into #552.

PR #619, opened 2026-05-04
[codex] test(web): add route migration phase 0 coverage

FranDias:
## Summary

Adds Phase 0 coverage for the planned TanStack Router migration without mounting the router in product code yet.

- Extracts the current hash parser/serializer into pure legacy route helpers and pins current channel, DM, app, wiki, notebook, and review URL behavior with tests.
- Adds a route registry plus TanStack route-tree tests so the migration has a concrete inventory and route contract guardrails.
- Expands the unmounted TanStack Router skeleton to include the planned route surface with hash-history compatibility.
- Updates the Telegram wizard E2E spec to match the current token -> mode -> group/DM flow surfaced by the full shell run.
- Documents the web path-forward plan and Phase 0 findings.

## Impact

This PR is intended to make the router migration safer before product routing changes. It gives us tests that verify current behavior, detect route contract drift, and document migration-specific decisions such as wiki splats and hash-history hrefs.

## Validation

- `bash scripts/test-web.sh web/src/routes/legacyHash.test.ts web/src/routes/routeRegistry.test.ts` passed, 44 tests.
- `bunx tsc --noEmit` passed.
- `bunx biome check web/src/routes/legacyHash.ts web/src/routes/routeRegistry.test.ts web/e2e/tests/telegram-connect.spec.ts` passed.
- `bash scripts/test-web.sh` passed, 110 files / 799 tests.
- `bun run build` passed with the existing large-chunk warning.
- `web/e2e/run-local.sh shell` passed, 31 tests.
- Pre-push hook passed: build, complexity-baseline, file-size, Go unit, smoke, VHS, and web-unit.

## Summary by CodeRabbit

* **New Features**
  * Enhanced navigation routing infrastructure with improved URL state management and support for legacy hash-based URLs.

* **Tests**
  * Added comprehensive test coverage for legacy hash routing patterns and route registry validation.
  * Refactored Telegram integration E2E tests with reusable helper functions.

* **Documentation**
  * Added detailed migration plan for web app navigation architecture.

PR #621, opened 2026-05-04
feat(web): mount TanStack Router (step 2 of route migration)

FranDias:
## Summary

Step 2 of the route-migration plan in `docs/experiments/2026-05-04-web-path-forward.md`, stacked on top of PR #619.

- `main.tsx` mounts `<RouterProvider/>` with `createHashHistory()` instead of `<App/>`. The root route's `component` is attached at startup to keep `lib/router.ts` free of route-component imports.
- New `web/src/routes/RootRoute.tsx` owns the boot lifecycle (api init, onboarding gate, theme css), the app shell, and the global host components (Toast, Confirm, ProviderSwitcher, TelegramConnect, ErrorBoundary, UpgradeBanner). Two effect-only siblings live inside it:
  - **UrlToStoreHydrator** runs in `useLayoutEffect`. It reads `useMatches()` and dispatches to `applyMatchToStore` from `routeSync.ts`.
  - **StoreToRouterBridge** subscribes directly to Zustand and reads `router.state.location` at fire time, calling `useNavigate` only when the URL doesn't already match the derived target.
- New `web/src/routes/routeSync.ts`: the URL↔store glue, independently testable. Each route maps to a patcher returning `Partial<AppStore>`; `applyMatchToStore` commits via a single `useAppStore.setState` so the bridge can never observe an intermediate slice (e.g. `wikiLookupQuery` written, `currentApp` still stale).
- New `NotFoundSurface` + `RoutedBody` wrapper: unmatched routes (`/console`, `/threads`, any typo) now render a small not-found screen instead of leaving `MainContent` rendering whatever store state was last written.
- `useHashRouter` and the legacyHash phase-0 helpers are deleted along with `App.tsx`.
- `/console` and `/threads` legacy aliases drop out of `router.ts` and `routeRegistry.ts` per the plan's open question #8.
- 20 new adapter tests in `web/src/routes/routeSync.test.ts` covering URL→store hydration for every route, store→URL derivation, atomic write notification count, and unmatched-route detection. Replaces the behavior coverage the deleted Phase 0 `legacyHash` tests carried.
- Bundle ceiling raised: `scripts/check-bundle-size.sh` `FAIL_KB` 1300 → 1400, `WARN_KB` 950 → 1350. The TanStack Router runtime is intentional product weight; step 5 of the migration will lazy-load heavy panels and ratchet back down.

This is the temporary-adapter step in the plan: existing navigation call sites still mutate the store, and the bridge mirrors those changes into the URL. **Step 3** of the migration converts those call sites to `router.navigate(...)` / `<Link>` and deletes `StoreToRouterBridge` entirely, after which the store fields it mirrored become read-only derived state until step 4 drains them out.

## Open questions resolved by this PR

- (#1) The phase-0 router stub is replaced wholesale: hash history, complete route table, real root component.
- (#5) E2E scope: the PR mounts the router in product code, so the full shell run is required.
- (#8) No legacy URL aliases — `/console` and `/threads` no longer match.

Open questions deferred:

- (#2) The temporary store-write bridge stays in place with a deadline tied to step 3 of the same PR train.
- (#3) DM `channelMeta` hydration now lives in the atomic patch returned from `patchDm` rather than calling `enterDM`; the cleaner cross-store shape is part of step 4.

## Test plan

- [x] `cd web && bunx tsc --noEmit`
- [x] `cd web && bun run build`
- [x] `bash scripts/check-bundle-size.sh` — 1313 KB ≤ 1400 KB
- [x] `bash scripts/test-web.sh` — 110 files, 793 tests (incl. 20 new adapter tests)
- [x] `web/e2e/run-local.sh shell` — 31 tests
- [x] `bunx biome check web/src/routes/{RootRoute,routeSync,routeSync.test}.ts`


najmuzzaman-mohammad (approved):

FranDias:
Superseded by #634 — flattened the stack into a single PR for end-to-end review.

PR #658, opened 2026-05-05
feat(web): UI review — onboarding wizard, splash screen, kbd styles

plyuto-lgtm:
## Summary

Cosmetic-only UI pass across the onboarding wizard and splash screen.

**Splash screen**
- Replaced italic logo text with bold Inter + inline SVG app icon
- Removed subtitle, border, and yellow color — WUPHF now uses `#612A92`
- Background set to `#f9f9fb` (matches wizard gradient)
- Spinner repositioned above logo group
- Auto-dismiss extended to 3s; smooth 600ms fade-in / 500ms fade-out transitions added

**Onboarding wizard**
- All step 1–5 CTA buttons unified to "Continue"
- Selected template cards: background fill only, no stroke (`border-color: transparent`)
- Enter key on focused template cards now advances the step (was toggling selection)
- `NexSignupPanel`: "Register an email…" rendered as a proper title (not uppercase label); title + description grouped with 8px gap
- `Step6Nex`: Register CTA + Skip button added, matching task step pattern
- `Step6Task`: removed `wizard-panel` wrapper; textarea has shadow appearance; suggestions rendered as plain list with staggered entrance animation, 16px left indent, muted gray text

**Kbd styles**
- `.kbd-sm` updated: 20px height, `--bg-subtle` background, sans-serif font

## Test plan
- [ ] Walk the full onboarding wizard from step 1 → splash → app
- [ ] Verify Enter on a focused template card advances (not toggles)
- [ ] Verify splash fades in/out smoothly and dismisses after ~3s
- [ ] Verify no regressions in main app shell styles

## Summary by CodeRabbit

* **New Features**
  * Added Nex signup step during onboarding
  * Unified provider and runtime selection in setup

* **UI/UX Improvements**
  * Redesigned onboarding wizard with updated flow and copy
  * Enhanced splash screen with new logo graphics
  * Improved workspace rail styling and interactions
  * Updated interview bar action layout for responsive behavior
  * Refined team selection and provider configuration interfaces
  * Enhanced local LLM picker visibility and controls


FranDias (dismissed):

FranDias (dismissed):

FranDias (dismissed):

FranDias (dismissed):

FranDias (approved):

web/src/components/onboarding/SplashScreen.tsx
```diff
@@ -1,14 +1,18 @@
...
   onDone: () => void;
 }
 
 export function SplashScreen({ onDone }: SplashScreenProps) {
+  const [dismissing, setDismissing] = useState(false);
+
   const dismiss = useCallback(() => {
-    onDone();
+    setDismissing(true);
+    setTimeout(onDone, FADE_OUT_MS);
```
FranDias: **🟠 IMPORTANT:** `setTimeout(onDone, FADE_OUT_MS)` fires on every `dismiss()` call but is never tracked or cleared. Two consequences:

1. The auto-dismiss timer in the effect fires `dismiss()` at t=3s. If a user also clicks at e.g. t=2.9s, `dismiss()` runs immediately — then the auto-dismiss still fires at t=3s, scheduling a second `onDone` call 500ms later.
2. If `SplashScreen` unmounts during the 500 ms fade window, the pending `onDone` fires into an unrelated parent state slot.

Fix: track and guard the timer:
```ts
const doneTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
const dismiss = useCallback(() => {
  if (doneTimerRef.current) return; // idempotent
  setDismissing(true);
  doneTimerRef.current = setTimeout(onDone, FADE_OUT_MS);
}, [onDone]);
useEffect(() => () => {
  if (doneTimerRef.current) clearTimeout(doneTimerRef.current);
}, []);
```

web/src/components/onboarding/Wizard.tsx
```diff
@@ -754,6 +763,10 @@ export function Wizard({ onComplete }: WizardProps) {
             nextStep();
           }
           return;
+        case "nex":
```
FranDias, line 809: **🟠 IMPORTANT:** Escape on the `nex` step calls `closeNexSignup()` which sets `nexSignupStatus = "hidden"`, but this produces zero observable change. `NexSignupPanel` has no `"hidden"` branch — it falls through to the email input form, identical to `"open"`. The step stays visible; CTAs behave exactly as before.

`"hidden"` is also effectively unreachable in normal flow: initial state is `"open"`, success → `"ok"`, nex-cli missing → `"fallback"`, error → back to `"open"`. Escape is the only path to `"hidden"` and it does nothing useful.

Fix: either (a) drop the `"hidden"` branch from `NexSignupStatus`, `NexSignupPanel`, and `closeNexSignup` entirely, or (b) have Escape on the `nex` step call `nextStep()` to skip the optional step, which matches user expectation.

web/src/components/onboarding/wizard/Step6Nex.tsx
```diff
@@ -0,0 +1,122 @@
+import { ONBOARDING_COPY } from "../../../lib/constants";
```
FranDias, line 1: **🟡 SUGGESTION:** No unit tests for this new step. The tests that previously exercised `NexSignupPanel` were deleted in this PR but not replaced — the four render states (`open`, `submitting`, `ok`, `fallback`) have no coverage. The e2e test only covers the Skip path.

Please add a `Step6Nex.test.tsx` covering at minimum: (1) primary button enabled/disabled by email trim, (2) `ok` state shows inbox confirmation + Continue, (3) `fallback` state shows the external link + Skip-only, (4) `submitting` disables both CTAs.

web/src/components/onboarding/wizard/Step6Nex.tsx
```diff
@@ -0,0 +1,122 @@
...
+        <h1
+          className="wizard-headline wizard-headline-sm"
+          style={{
+            display: "flex",
+            alignItems: "center",
+            justifyContent: "center",
+            gap: 10,
+          }}
+        >
+          Power up with
```
FranDias, line 39: **🟡 SUGGESTION:** Headline "Power up with" is hardcoded here while `ONBOARDING_COPY.step6_headline = "Power up with Nex"` (defined in `web/src/lib/constants.ts`) is never read. The two will silently drift on the next copy edit. Either reference the constant (splitting off the ` Nex` suffix before the SVG) or drop the unused constant.

web/src/components/onboarding/Wizard.tsx
```diff
@@ -167,7 +168,7 @@ export function Wizard({ onComplete }: WizardProps) {
   // nex.ai/register, key pasted on the Setup step).
   const [nexEmail, setNexEmail] = useState("");
   const [nexSignupStatus, setNexSignupStatus] =
-    useState<NexSignupStatus>("hidden");
+    useState<NexSignupStatus>("open");
```
FranDias, line 201: **🟡 SUGGESTION:** With Step6Nex now its own dedicated wizard step, `"hidden"` in `NexSignupStatus` is effectively dead — the only code that sets it is `closeNexSignup()` (called on Escape), which produces no observable UI change. Consider tightening the union to `"open" | "submitting" | "ok" | "fallback"` and removing `closeNexSignup`. This would simplify `NexSignupPanel`'s render branches and the Escape handler.

FranDias (commented):
Cosmetic-only UI pass over the onboarding wizard, splash screen, and kbd styles. Adds a dedicated Nex registration step (Step6Nex), unifies CTA labels to "Continue", refreshes the splash with an inline SVG logo and 600/500ms fade transitions, restyles `.kbd-sm`, and rebuilds Step5Setup around a typed UNIFIED_PROVIDERS list with auth-mode-aware tabs. Tests in the diff are mostly relabel updates plus an e2e Skip click for the new nex step. Mostly safe, but there is one real timer leak in the splash, orphaned `"hidden"` status state, and missing test coverage for the new NexStep.

[No local checkout — diff only]

---
Findings: 🔴 0 critical · 🟠 2 important · 🟡 8 suggestions

web/src/components/onboarding/wizard/RuntimeLogos.tsx
```diff
@@ -0,0 +1,147 @@
...
+    </svg>
+  );
+}
+
+const LOGO_MAP: Record<string, () => React.ReactElement> = {
+  "Claude Code": ClaudeLogo,
+  Codex: OpenAILogo,
+  Opencode: OpencodeLogo,
+  Cursor: CursorLogo,
+  Windsurf: WindsurfLogo,
```
FranDias: **🟡 SUGGESTION:** `CursorLogo` and `WindsurfLogo` are registered in `LOGO_MAP` but neither `"Cursor"` nor `"Windsurf"` appear in `UNIFIED_PROVIDERS` label values (Step5Setup.tsx). These logos are dead code in the new setup UI. Either remove them and their `LOGO_MAP` entries, or wire them into `UNIFIED_PROVIDERS`. Leaving them as ghosts means they'll silently rot when icon styling changes.

web/src/styles/kbd.css
```diff
@@ -75,16 +83,42 @@
...
+  border-radius: 4px;
+  box-sizing: border-box;
+}
+.kbd-hint-mod {
+  padding: 2px 4px;
+  border: 1px solid rgba(255, 255, 255, 0.25);
+  border-bottom-width: 2px;
+  border-radius: 4px;
+}
+.kbd-hint-mod {
```
FranDias: **🟡 SUGGESTION:** Two separate `.kbd-hint-mod` rule blocks for the same selector appear back-to-back in this hunk (lines ~111–116 and ~117–122). Cascade works fine but this is noise and a future-edit footgun. Merge into a single block.

web/src/styles/messages.css
```diff
@@ -580,6 +590,17 @@
   flex: 1;
   min-width: 0;
 }
+.interview-bar-actions-inline {
+  display: flex;
+  grid-template-columns: unset;
```
FranDias: **🟡 SUGGESTION:** `grid-template-columns: unset` has no effect here — the parent rule on the same selector sets `display: flex`, so this is a flex container, not a grid container. The property is silently ignored. Drop the line, or add a comment if the intent is to override the parent's `grid-template-columns` for specificity reasons.

web/src/styles/layout.css
```diff
@@ -301,7 +301,7 @@ html[data-theme="nex"]
   --top-panel-h: 56px;
   --sidebar-pad-x: 12px;
   --bottom-panel-h: 36px;
-  --rail-popover-left: 64px;
+  --rail-popover-left: 120px;
```
FranDias, line 312: **🟡 SUGGESTION:** `--rail-popover-left` jumped from 64px to 120px. The rail is 56px wide, so the old value (56 + 8px gap = 64px) kept popovers snug to the rail edge. At 120px the anchor is 64px past the rail edge — rail tooltips/context menus will appear visually disconnected. If this is to clear a sidebar between the rail and the popover target, derive it from layout tokens instead of hardcoding: e.g. `calc(var(--workspace-rail-w, 56px) + var(--sidebar-w, 56px) + 8px)`, so it doesn't silently break when rail sizing changes.

web/src/components/onboarding/wizard/Step5Setup.tsx
```diff
@@ -46,133 +39,149 @@ interface SetupStepProps {
...
-  prereqsError,
-  version,
 }: {
-  spec: RuntimeSpec;
+  provider: (typeof UNIFIED_PROVIDERS)[number];
+  apiKeyValue: string;
+  onApiKeyChange: (v: string) => void;
+  showInput: boolean;
+  onToggleInput: (show: boolean) => void;
   installed: boolean;
```
FranDias: **🟡 SUGGESTION:** Google has `runtimeLabel: null`, so it never enters the `runtimePriority` array. The priority badge for Google always shows empty even after a user adds a `GOOGLE_API_KEY`. The key is correctly persisted (Wizard.tsx l.638), but Google is invisible in the fallback-chain UI. If intentional (Gemini is API-key-only, not a wuphf CLI runtime), add a one-line comment in `UNIFIED_PROVIDERS` explaining why `runtimeLabel` is `null` and what that means for the priority UX.

FranDias (commented):
Cosmetic-only UI pass over the onboarding wizard, splash screen, and kbd styles. Adds a dedicated Nex registration step (Step6Nex), unifies CTA labels to "Continue", refreshes the splash with an inline SVG logo and 600/500ms fade transitions, restyles `.kbd-sm`, and rebuilds Step5Setup around a typed UNIFIED_PROVIDERS list with auth-mode-aware tabs. Tests in the diff are mostly relabel updates plus an e2e Skip click for the new nex step. Mostly safe, but there is one real timer leak in the splash, orphaned `"hidden"` status state, and missing test coverage for the new NexStep.

[No local checkout — diff only]

---
Findings: 🔴 0 critical · 🟠 2 important · 🟡 8 suggestions

web/src/components/onboarding/SplashScreen.tsx
FranDias: **🟠 IMPORTANT:** `setTimeout(onDone, FADE_OUT_MS)` fires on every `dismiss()` call but is never tracked or cleared. Two consequences:

1. The auto-dismiss timer in the effect fires `dismiss()` at t=3s. If a user also clicks at e.g. t=2.9s, `dismiss()` runs immediately — then the auto-dismiss still fires at t=3s, scheduling a second `onDone` call 500ms later.
2. If `SplashScreen` unmounts during the 500 ms fade window, the pending `onDone` fires into an unrelated parent state slot.

Fix: track and guard the timer:
```ts
const doneTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
const dismiss = useCallback(() => {
  if (doneTimerRef.current) return; // idempotent
  setDismissing(true);
  doneTimerRef.current = setTimeout(onDone, FADE_OUT_MS);
}, [onDone]);
useEffect(() => () => {
  if (doneTimerRef.current) clearTimeout(doneTimerRef.current);
}, []);
```

web/src/components/onboarding/Wizard.tsx
FranDias, line 809: **🟠 IMPORTANT:** Escape on the `nex` step calls `closeNexSignup()` which sets `nexSignupStatus = "hidden"`, but this produces zero observable change. `NexSignupPanel` has no `"hidden"` branch — it falls through to the email input form, identical to `"open"`. The step stays visible; CTAs behave exactly as before.

`"hidden"` is also effectively unreachable in normal flow: initial state is `"open"`, success → `"ok"`, nex-cli missing → `"fallback"`, error → back to `"open"`. Escape is the only path to `"hidden"` and it does nothing useful.

Fix: either (a) drop the `"hidden"` branch from `NexSignupStatus`, `NexSignupPanel`, and `closeNexSignup` entirely, or (b) have Escape on the `nex` step call `nextStep()` to skip the optional step, which matches user expectation.

web/src/components/onboarding/wizard/Step6Nex.tsx
FranDias, line 1: **🟡 SUGGESTION:** No unit tests for this new step. The tests that previously exercised `NexSignupPanel` were deleted in this PR but not replaced — the four render states (`open`, `submitting`, `ok`, `fallback`) have no coverage. The e2e test only covers the Skip path.

Please add a `Step6Nex.test.tsx` covering at minimum: (1) primary button enabled/disabled by email trim, (2) `ok` state shows inbox confirmation + Continue, (3) `fallback` state shows the external link + Skip-only, (4) `submitting` disables both CTAs.

web/src/components/onboarding/wizard/Step6Nex.tsx
FranDias, line 39: **🟡 SUGGESTION:** Headline "Power up with" is hardcoded here while `ONBOARDING_COPY.step6_headline = "Power up with Nex"` (defined in `web/src/lib/constants.ts`) is never read. The two will silently drift on the next copy edit. Either reference the constant (splitting off the ` Nex` suffix before the SVG) or drop the unused constant.

web/src/components/onboarding/Wizard.tsx
FranDias, line 201: **🟡 SUGGESTION:** With Step6Nex now its own dedicated wizard step, `"hidden"` in `NexSignupStatus` is effectively dead — the only code that sets it is `closeNexSignup()` (called on Escape), which produces no observable UI change. Consider tightening the union to `"open" | "submitting" | "ok" | "fallback"` and removing `closeNexSignup`. This would simplify `NexSignupPanel`'s render branches and the Escape handler.

web/src/components/onboarding/wizard/RuntimeLogos.tsx
FranDias: **🟡 SUGGESTION:** `CursorLogo` and `WindsurfLogo` are registered in `LOGO_MAP` but neither `"Cursor"` nor `"Windsurf"` appear in `UNIFIED_PROVIDERS` label values (Step5Setup.tsx). These logos are dead code in the new setup UI. Either remove them and their `LOGO_MAP` entries, or wire them into `UNIFIED_PROVIDERS`. Leaving them as ghosts means they'll silently rot when icon styling changes.

web/src/styles/kbd.css
FranDias: **🟡 SUGGESTION:** Two separate `.kbd-hint-mod` rule blocks for the same selector appear back-to-back in this hunk (lines ~111–116 and ~117–122). Cascade works fine but this is noise and a future-edit footgun. Merge into a single block.

web/src/styles/messages.css
FranDias: **🟡 SUGGESTION:** `grid-template-columns: unset` has no effect here — the parent rule on the same selector sets `display: flex`, so this is a flex container, not a grid container. The property is silently ignored. Drop the line, or add a comment if the intent is to override the parent's `grid-template-columns` for specificity reasons.

web/src/styles/layout.css
FranDias, line 312: **🟡 SUGGESTION:** `--rail-popover-left` jumped from 64px to 120px. The rail is 56px wide, so the old value (56 + 8px gap = 64px) kept popovers snug to the rail edge. At 120px the anchor is 64px past the rail edge — rail tooltips/context menus will appear visually disconnected. If this is to clear a sidebar between the rail and the popover target, derive it from layout tokens instead of hardcoding: e.g. `calc(var(--workspace-rail-w, 56px) + var(--sidebar-w, 56px) + 8px)`, so it doesn't silently break when rail sizing changes.

web/src/components/onboarding/wizard/Step5Setup.tsx
FranDias: **🟡 SUGGESTION:** Google has `runtimeLabel: null`, so it never enters the `runtimePriority` array. The priority badge for Google always shows empty even after a user adds a `GOOGLE_API_KEY`. The key is correctly persisted (Wizard.tsx l.638), but Google is invisible in the fallback-chain UI. If intentional (Gemini is API-key-only, not a wuphf CLI runtime), add a one-line comment in `UNIFIED_PROVIDERS` explaining why `runtimeLabel` is `null` and what that means for the priority UX.

FranDias (dismissed):

FranDias (approved):

web/src/components/onboarding/wizard/constants.ts
```diff
@@ -19,6 +19,7 @@ export const STEP_ORDER: readonly WizardStep[] = [
   "templates",
   "team",
   "setup",
+  "nex",
```
FranDias, line 22: **🔴 CRITICAL:** Adding `"nex"` to `STEP_ORDER` without updating the resume-draft validator silently corrupts the resume flow. `web/src/components/onboarding/wizard/onboardingDraft.ts` keeps an explicit `VALID_STEPS` allow-list (lines 49–57) that still reads:

```ts
const VALID_STEPS: ReadonlySet<WizardStep> = new Set<WizardStep>([
  "welcome", "templates", "identity", "team", "setup", "task", "ready",
]);
```

`useOnboardingDraftSync` now persists `step: "nex"` whenever a user is on this step. On the next mount, `loadDraft` → `isValidDraft` rejects it (`!VALID_STEPS.has("nex")`), `clearDraft()` runs, and the user loses **everything** — office name, description, blueprint pick, runtime priority — with no resume banner.

This is a real, reproducible data-loss bug: any user who reaches the Nex step and closes/refreshes will be returned to the welcome screen with no draft.

**Fix:** Add `"nex"` to `VALID_STEPS` in `onboardingDraft.ts`, or better yet derive it directly from `STEP_ORDER` so the two can never drift:

```ts
const VALID_STEPS: ReadonlySet<WizardStep> = new Set<WizardStep>(STEP_ORDER);
```

web/src/components/onboarding/wizard/Step6Nex.tsx
```diff
@@ -0,0 +1,125 @@
...
+              Skip
+            </button>
+          ) : null}
+          {status === "ok" ? (
+            <button className="btn btn-primary" onClick={onNext} type="button">
+              Continue
+            </button>
+          ) : status === "fallback" ? (
+            <button className="btn btn-primary" onClick={onNext} type="button">
+              Skip
```
FranDias: **🟠 IMPORTANT:** When `status === "fallback"` the only CTA in the nav row is a `btn-primary`-styled button labelled `"Skip"`. The `status !== "ok" && status !== "fallback"` guard above hides the secondary `task-skip` button, leaving a primary-styled "Skip" as the sole action.

A primary CTA labelled "Skip" is contradictory — primary buttons signal the recommended action, but here the user is abandoning registration. The keyboard handler in `Wizard.tsx` (`case "nex"`) consistently routes Enter in fallback to `nextStep()`, so the logic is correct; only the visual weight is off.

**Fix:** Either (a) relabel to `Continue` in the fallback path (the user already triggered the external link — advancing is the expected next step), or (b) use `btn-secondary`/`task-skip` styling to match the visual hierarchy of a skip action.

web/src/components/messages/InterviewBar.tsx
```diff
@@ -85,13 +86,15 @@ export function InterviewBar() {
...
+  // compareOpen before state from one card leaks into the next.
   useEffect(() => {
+    const currentId = current?.id ?? null;
+    if (activeRequestIdRef.current === currentId) return;
+    activeRequestIdRef.current = currentId;
     setTextMode(null);
     setCustomText("");
     setCompareOpen(false);
-  }, []);
+  });
```
FranDias: **🟡 SUGGESTION:** Removing the `[]` dep array so the effect runs after every render, guarded by a ref, is functionally correct but reads as a forgotten dep array to most reviewers.

The canonical equivalent:

```ts
useEffect(() => {
  setTextMode(null);
  setCustomText("");
  setCompareOpen(false);
}, [current?.id]);
```

This fires exactly when the active request id changes, avoids per-render overhead, and is immediately readable.

web/src/components/onboarding/SplashScreen.tsx
```diff
@@ -1,34 +1,78 @@
...
+            <g clipPath="url(#lc0)">
+              <path d="M80 0H0V80H80V0Z" fill="#FFB3E6" />
+              <path d="M25 15H15V65H25V15Z" fill="#612A92" />
+              <path d="M65 15H55V65H65V15Z" fill="#612A92" />
+              <path d="M45 35H35V55H45V35Z" fill="#612A92" />
+              <path d="M35 55H25V65H35V55Z" fill="#612A92" />
+              <path d="M55 55H45V65H55V55Z" fill="#612A92" />
+            </g>
+            <defs>
+              <clipPath id="lc0">
```
FranDias: **🟡 SUGGESTION:** Hard-coded `id="lc0"` on the `<clipPath>` (and `url(#lc0)` three lines above). If `<SplashScreen>` is ever rendered twice in the same document — testing harnesses, HMR transient states, Storybook — both clipPaths share the same id and the SVG spec resolves to the first match, silently breaking the rounded-rect mask on one instance.

**Fix:** Use `useId()`:

```tsx
const clipId = useId();
// ...
<g clipPath={`url(#${clipId})`}> ... </g>
<clipPath id={clipId}> ... </clipPath>
```

web/src/components/onboarding/wizard/RuntimeLogos.tsx
```diff
@@ -0,0 +1,111 @@
...
+      strokeLinecap="round"
+      strokeLinejoin="round"
+      aria-hidden="true"
+    >
+      <path d="M3 7v10a2 2 0 002 2h14a2 2 0 002-2V9a2 2 0 00-2-2h-6l-2-2H5a2 2 0 00-2 2z" />
+    </svg>
+  );
+}
+
+const LOGO_MAP: Record<string, () => ReactElement> = {
```
FranDias: **🟡 SUGGESTION:** `LOGO_MAP: Record<string, () => ReactElement>` accepts arbitrary strings — a label typo or future rename in `UNIFIED_PROVIDERS` silently returns `null` with no compile-time warning.

Since the valid labels are a closed set, type the map:

```ts
type RuntimeLogoLabel =
  | "Claude Code" | "Codex" | "Opencode" | "Google" | "Run a local model";

const LOGO_MAP: Partial<Record<RuntimeLogoLabel, () => ReactElement>> = { ... };
```

A future rename then becomes a TS error instead of a missing icon in production.

FranDias (commented):
Cosmetic-leaning PR that splits Nex signup into its own wizard step, polishes the splash screen, and adjusts kbd/layout styles. The visual changes are clean and the new step's tests are solid — but **the addition of `"nex"` to `STEP_ORDER` was not propagated to `VALID_STEPS` in `onboardingDraft.ts`**, causing a silent data-loss bug: any user who reaches the Nex step and refreshes has their draft rejected and cleared on the next mount. Also flagged: a UX inconsistency in the fallback CTA, an unconventional `useEffect` pattern in `InterviewBar`, a static SVG clipPath id, and a non-type-safe runtime logo map.

---
Findings: 🔴 1 critical · 🟠 1 important · 🟡 3 suggestions

[No local checkout — diff only]

FranDias (dismissed):

FranDias (dismissed):

FranDias (dismissed):

FranDias (dismissed):

FranDias (dismissed):

FranDias (dismissed):

FranDias (approved):

PR #660, opened 2026-05-05
feat(notebook): auto-write per-agent notebook entries on broker events (PR 1)

najmuzzaman-mohammad:
## Summary

PR 1 of the notebook-wiki-promise design. Empty notebook shelves stop being empty: every roster-agent `Broker.PostMessage` and every task transition deterministically writes one markdown entry under `agents/{slug}/notebook/{YYYY-MM-DD-HHMMSS}-{kind}-{shortHash}.md`. No LLM on the hot path; no agent-judgment dependency.

- New `AutoNotebookWriter` (async buffered channel + dedicated goroutine, Start/Stop lifecycle that mirrors `WikiWorker`).
- Hooks attach **after** `b.saveLocked()` succeeds in `mutation_service.go`, `lifecycle.go`, `self_healing.go`, and on `Broker.PostMessage`.
- In-memory LRU dedupe per (slug, day) keyed by `sha256(content)`, ring buffer of 50.
- Pre-write secretlint regex scrub (AWS/GitHub/Stripe/OpenAI/Anthropic/SendGrid/Slack/Google/PEM private-key block).
- Roster-membership filter at the broker hook site (lock-free `isAgentMemberSlugLocked`) — required to avoid re-entering `b.mu` from `writer.Handle`.
- Errors and queue saturation are logged + counted + dropped. Never block the broker.

Source of truth: design doc at `~/.gstack/projects/nex-crm-wuphf/najmuzzaman-main-design-20260505-131620-notebook-wiki-promise.md` (13 locked decisions: OV2A, 2A, OV4A, 4A, OV3A, 6A, 7A, S1A, S2A, S3A, T1A, OV5A, OV6A, OV7A).

Memory-workflow gate (premise #3 closure) is **not** wired in this PR — deferred to TODO #19 per OV3A.

## Test plan

- [x] `go test -count=1 ./internal/team/...` — green
- [x] `go test -count=1 -race -timeout 600s ./internal/team/` — green (84s)
- [x] `bash scripts/test-go.sh ./internal/team` — all 1 packages green
- [x] `go vet ./internal/team/...` — clean
- [x] Unit tests in `auto_notebook_writer_test.go`: enqueue path, roster filter, dedupe within day, dedupe per-day buckets, redaction, task transition kind/path, no-op transition drop, queue saturation drop-not-block, Handle-after-Stop no-op, write-failure counted-not-propagated, secret regex set, UTF-8 truncate, markdown escape.
- [x] Integration tests in `broker_auto_notebook_writer_test.go` against a real `WikiWorker` + temp git repo: PostMessage lands on shelf, human message does not, duplicate post deduped, task mutation lands on owner shelf with `task-transitioned` filename.
- [x] Benchmarks in `broker_auto_notebook_writer_bench_test.go`: `BenchmarkPostMessage_WithAutoNotebook` vs baseline (run locally with `go test -bench .`).
- [x] `broker_messages_test.go` extended with `TestPostMessage_TriggersAutoNotebookWriter` to pin the hook seam.

## Summary by CodeRabbit

* **New Features**
  * Automatic per-agent notebook entries for agent messages and task transitions: dated notes, per-day dedupe, secret redaction, roster filtering, non‑blocking enqueue with saturation handling, lifecycle-aware start/stop, runtime counters/inspection, and transitions emitted after state persistence.

* **Tests**
  * Extensive unit and integration tests covering writes, dedupe, redaction, roster filtering, queue saturation, transition semantics, and shutdown behavior.

* **Benchmarks**
  * Benchmarks measuring message latency with and without the notebook writer attached.

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -188,6 +188,60 @@ Tracking work that is deliberately deferred from the current branch. Each item n
 
 ---
 
+### 18. Wiki worker queue depth monitoring + alert thresholds (post-PR 1 notebook auto-writer)
+
+**What:** Instrument `wiki_worker.go` to emit queue depth and time-in-queue metrics, plus a warn log at 80% of the 64-buffer capacity. Track the dropped-event counter from `auto_notebook_writer.go` alongside.
+
+**Why:** PR 1 of the notebook-wiki-promise effort adds a new write source (every agent message + every task transition) that competes with real wiki, artifact, fact, lint, playbook, and learning writes through the same shared queue. Codex flagged this as the most likely operational failure mode. Without observability we discover starvation only when a real wiki write fails or times out.
+
+**Pros:** Confirms PR 1 didn't quietly DoS real wiki writes. Gives a concrete signal for tuning the auto-writer's enqueue capacity or for triggering the OV7B event-log migration (TODO #20).
+
+**Cons:** Adds a small metrics surface; needs to land somewhere in the existing analytics path.
+
+**Context:** PR 1 design doc: `~/.gstack/projects/nex-crm-wuphf/najmuzzaman-main-design-20260505-131620-notebook-wiki-promise.md`. Eng review: same dir, dated 2026-05-05. Decision OV7A locked the per-event commit model with bounded enqueue; finding #8 from the Codex outside voice is the underlying concern.
+
+**Depends on / blocked by:** PR 1 must ship first (introduces the new write source).
+
+**Trigger to revisit:** First week of real PR 1 traffic, OR if any real wiki write ever times out.
+
+---
+
+### 19. Premise #3 closure design pass: memory workflow gate auto-satisfaction or removal
+
+**What:** Fresh `/office-hours` design pass on closing the memory workflow gate semantically. Today the gate requires `lookup + capture + promote` all satisfied to mark a task complete (`memory_workflow.go:398`). The eng review had originally planned to satisfy it inline from the auto-writer, but Codex flagged two killers: (a) the gate needs all three steps, not just capture; (b) calling `RecordTaskMemoryCapture` inline holds `b.mu` and would deadlock with the broker API path.
+
+**Why:** Premise #3 of the notebook-wiki-promise design says "kill the narrow `process_research` filter so the gate applies to all tasks, auto-satisfied by the writer." With OV3A we deferred this from PR 1 because the implementation path was wrong. Two possible exit ramps: (1) auto-satisfy all three steps deterministically with proper locking (raw helper holding existing lock + queueing the file write async); (2) kill the gate entirely — once writes are deterministic the gate is dead weight, but we lose its override-tracking and partial-error surfaces.
+
+**Pros:** Closes premise #3 honestly. Either direction simplifies the broker.
+
+**Cons:** A real design pass, not a one-line filter removal. Potentially touches `memory_workflow.go`, `broker_tasks_memory_workflow.go`, and `memory_workflow_reconciler.go`.
+
+**Context:** Notebook-wiki-promise design doc PR 8 row was originally "5-line filter removal at memory_workflow.go:212-227." Codex outside voice (2026-05-05) showed why that's wrong. Locked decisions OV3A, OV3B (rejected), OV3D (deadlock).
+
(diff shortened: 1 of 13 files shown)


FranDias (dismissed):

najmuzzaman-mohammad:
Pushed 99ad2b20 addressing all four review items:

1. **Stop-race panic (CodeRabbit, critical)** — replaced `close(w.queue)` shutdown with a dedicated `stopCh` signal channel. `Handle` now selects on `stopCh` before any send, so a concurrent caller that already cleared the running.Load() fast-path cannot panic on a closed-channel send. Run loop exits via stopCh; the queue is left intact and falls out via GC.

2. **Bench drifts into drop path (CodeRabbit)** — swapped the real `WikiWorker` for a `benchNotebookClient` no-op so the loop measures only the steady-state enqueue cost. Asserts `Counters().QueueSaturated == 0` after the run so any future regression that re-introduces drops fails the bench.

3. **Fixed sleep in negative-path test (CodeRabbit)** — removed every `time.Sleep` from the new tests (CONTRIBUTING.md "no time.Sleep in tests" lint was failing CI for this same reason). Added a `WaitForCondition(ctx, predicate)` primitive on the writer that fans out a sync.Cond signal after each event is processed; the human-message negative-path test now drains via an agent-sentinel `PostMessage` rather than a fixed wait.

4. **Dedupe collapses distinct transitions (CodeRabbit)** — folded `TaskID` plus `BeforeStatus`/`AfterStatus` into the writer's dedupe basis. Two distinct tasks owned by the same agent and sharing a title, or two consecutive transitions on the same task, can no longer collapse into one shelf entry.

`bash scripts/check-no-new-sleeps-in-tests.sh` is green locally; `go test -count=1 -race -timeout 600s ./internal/team/` passes in 85s.

FranDias (dismissed):

najmuzzaman-mohammad:
Pushed c392e3e0 addressing the outside-diff finding:

**`unblockDependentsLocked` missed emitting auto-notebook transitions (CodeRabbit, major).** The cascade path mutates dependent task statuses (`blocked` → `in_progress` / `open`) when a parent reaches a terminal status, but did not call `emitTaskTransitionAutoNotebook`. Every other status mutation in PR 1 emits one event; the dependency-cascade was the missing seam, which would have skewed PR 3's clustering signal once cross-agent dependency chains start firing.

Capture `beforeStatus` per-task before mutation and emit after the existing `appendActionLocked`, still under `b.mu` (caller's `saveLocked` persists state).

`go test -count=1 -race -timeout 600s ./internal/team/` — green (85.6s).

FranDias (dismissed):

najmuzzaman-mohammad:
Pushed da417361 addressing both new review items:

1. **Cascade emits leak on save failure (CodeRabbit, major).** Refactored `unblockDependentsLocked` to RETURN the pending transitions as `[]pendingTaskTransition` instead of calling `emitTaskTransitionAutoNotebook` directly. Production callers (`MutateTask` + the human-interview answer flow in `broker_requests_interviews.go`) now flush the batch via `flushPendingAutoNotebookTransitionsLocked` only after their own `saveLocked` succeeds. A failed save no longer leaves the writer holding entries for transitions that were rolled back.

2. **Unconditional emits on no-op transitions (CodeRabbit, minor).** Short-circuited at the source inside `emitTaskTransitionAutoNotebook` so every existing call site benefits with one change. The writer's `NoopTransition` counter still catches anything that slips through.

`go test -count=1 -race -timeout 600s ./internal/team/` — green (84.6s).

FranDias (dismissed):

najmuzzaman-mohammad:
Addressed in 102f7b60.

`unblockTasksForAnsweredRequestLocked` now returns `[]pendingTaskTransition`, capturing each unblocked task's `beforeStatus` before the `blocked → in_progress|open` flip. The caller appends these onto `pendingCascade`, so request-answer-driven transitions get flushed by `flushPendingAutoNotebookTransitionsLocked` after `saveLocked()` succeeds (same defer-until-after-save guarantee as the dependency cascade).

`emitTaskTransitionAutoNotebook`'s no-op guard handles the case where the task was Blocked but not status="blocked" (e.g. blocked-and-done), so we don't emit phantom transitions.

Tests: `go test -count=1 -race -timeout 600s ./internal/team/` → pass (84.9s).

FranDias (approved):

PR #662, opened 2026-05-05
[codex] Redact secrets from agent output

FranDias:
## Summary

- Add scanner-backed secret redaction for channel messages, human interviews, agent issues, stream output, actions, ledger signals/decisions, persistence, logs, and prompt/context surfaces.
- Persist and expose redaction metadata (`redacted`, `redaction_count`, `redaction_reasons`) through broker/TUI/web API types, with a compact web badge for redacted messages.
- Tighten entropy detection to avoid path/URL false positives and add prompt guidance so agents do not disclose credentials.

## Impact

Agent-visible and user-visible chat surfaces now redact detected API keys, tokens, passwords, private keys, and similar credentials before they reach storage, pubsub, external transports, logs, or rendered UI. Existing persisted state is sanitized on load/save.

## Validation

- `bash scripts/test-go.sh ./internal/scanner ./internal/team`
- `go test ./internal/team -run TestWriteHeadlessOpencodeMCPConfigLogsBaseConfigParseFailure`
- `go test ./internal/scanner`
- `bunx biome check web/src/api/client.ts web/src/components/messages/MessageBubble.tsx`
- `bunx tsc --noEmit` from `web/`
- `bash scripts/test-web.sh web/src/components/messages/MessageFeed.test.tsx`
- `bash scripts/test-web.sh`
- `bash scripts/test-go.sh ./cmd/wuphf ./cmd/wuphf/channelui`
- `git diff --check`
- `bash scripts/test-go.sh`
- `go vet ./...`
- `golangci-lint run ./...`
- `bunx --bun secretlint` on changed files
- Pre-push hook passed: build, complexity-baseline, file-size, go-unit, smoke, vhs, web-unit

## Summary by CodeRabbit

* **New Features**
  * Messages now redact detected secrets with a “[REDACTED]” marker and include redaction metadata (flag, count, reasons) shown in the UI.
  * UI shows a redacted badge with reason tooltip on messages, requests, and interview dialogs.
  * Last-message previews, streams, logs, requests, and persisted surfaces are sanitized and truncated for privacy.
  * Agent prompts include explicit credential-handling guidance.

* **Tests**
  * Added tests validating redaction, metadata, and subscriber/stream behavior.


internal/team/broker_publish.go
```diff
@@ -40,6 +43,7 @@ func (b *Broker) publishActionLocked(action officeActionLog) {
 }
 
 func (b *Broker) publishActivityLocked(activity agentActivitySnapshot) {
+	activity.Detail = redactSecretsInText(activity.Detail)
```
FranDias, line 59: **🟠 IMPORTANT:** `publishActivityLocked` only redacts `activity.Detail`, leaving `activity.Activity` untouched. The same gap is in `UpdateAgentActivity` in `broker_streams.go` — `update.Activity` is copied verbatim into `current.Activity` while `update.Detail` is redacted. In practice today's callers pass a closed enum (`"text"`, `"tool"`, `"thinking"`, `"idle"`, etc.) for `Activity`, so the risk is theoretical — but `UpdateAgentActivity` is an exported broker method with no callsite restriction, and the type contract makes no closed-enum guarantee. A future caller (or a round-trip through JSON) can pass user/model-supplied text in `Activity` and it will reach SSE subscribers, the TUI, and the web Members API uncensored.

**Fix:** redact symmetrically:
```go
activity.Activity = redactSecretsInText(activity.Activity)
activity.Detail   = redactSecretsInText(activity.Detail)
```

internal/team/broker_streams_test.go
```diff
@@ -45,6 +45,35 @@ func TestAgentStreamBuffer_RecentReturnsBoundedHistory(t *testing.T) {
...
+
+	select {
+	case got := <-out:
+		if strings.Contains(got, secret) {
+			t.Fatalf("subscriber leaked secret: %q", got)
+		}
+		if !strings.Contains(got, "[REDACTED]") {
+			t.Fatalf("subscriber missing redaction marker: %q", got)
+		}
+	case <-time.After(time.Second):
```
FranDias: **🟠 IMPORTANT:** `time.After(time.Second)` here (and `time.After(2 * time.Second)` in `TestBrokerMessageSubscribersReceiveRedactedMessages`) are time-bounded test guards. The project rule (`feedback_no_sleeps_in_tests`) says tests must not gate on wall-clock waits — find a deterministic signal or rewrite inputs.

Both subscriber channels are already buffered, and the publish loops run under `b.mu`, so by the time `PostMessage` / `PushTask` returns the value is already enqueued. A bare `got := <-msgs` is deterministic and eliminates the flake window.

**Fix:** drop the `select` / `time.After` pair entirely; receive directly:
```go
got := <-out
// assert on got
```
If you want a regression guard against a channel that's never written, let the test hang into the package-level `-timeout` rather than embedding an arbitrary 1–2 s budget in every redaction test.

internal/team/broker_human.go
```diff
@@ -201,6 +201,7 @@ func (b *Broker) Requests(channel string, includeResolved bool) []humanInterview
 }
 
 func cloneHumanInterview(req humanInterview) humanInterview {
+	req = sanitizeHumanInterview(req)
```
FranDias, line 204: **🟠 IMPORTANT:** `sanitizeHumanInterview` redacts secrets in-place but emits no metadata — `humanInterview` has no `Redacted`/`RedactionCount`/`RedactionReasons` fields, so the UI cannot warn humans that the question, context, options, or answer they're reading had something stripped.

This is asymmetric with `sanitizeChannelMessageSecrets`, which sets `msg.Redacted = true` and surfaces a badge in `MessageBubble.tsx`. For the much higher-stakes interview surface (the human is about to make a decision based on this text) the scrub is silent. Worst case: an agent pastes a credential into the interview context, the redactor strips it, and the human sees `[REDACTED]` with no explanation.

**Fix:** add `Redacted bool`, `RedactionCount int`, `RedactionReasons []string` to `humanInterview` and accumulate hits the same way `sanitizeChannelMessageSecrets` does, then plumb a badge through the requests view to mirror the `MessageBubble` treatment.

internal/team/broker_persistence.go
```diff
@@ -205,20 +223,44 @@ func (b *Broker) prepareBrokerStateWriteLocked() (brokerStateWrite, error) {
...
 		Tasks:             b.tasks,
-		Requests:          b.requests,
-		Actions:           b.actions,
-		Signals:           b.signals,
-		Decisions:         b.decisions,
+		Requests:          requests,
+		Actions:           actions,
+		Signals:           signals,
+		Decisions:         decisions,
 		Watchdogs:         b.watchdogs,
```
FranDias: **🟠 IMPORTANT:** `Watchdogs: b.watchdogs` is persisted and loaded raw — neither `loadState` nor this write path runs any redaction over `alert.Summary`. Only the at-creation site in `ledger.go` covers forward writes; legacy state files written before this PR keep raw watchdog summaries through every load/save cycle.

Same gap for `teamTask.Title` / `teamTask.Details` (line 167–169 in `loadState`, migrated for DM slugs only) and any free-text fields on `schedulerJob` — all are user/agent-controlled strings included in TUI task views and the web TaskDetailModal, returned raw from `b.tasks`/`b.scheduler`.

The PR description claims "Existing persisted state is sanitized on load" — that's true for messages, agentIssues, requests, pendingInterview, actions, signals, and decisions, but not for watchdogs, tasks, or the scheduler queue.

**Fix:** add sanitize passes for `watchdogAlert.Summary`, `teamTask.Title`/`Details`, and relevant `schedulerJob` fields in both `loadState` and `prepareBrokerStateWriteLocked`.

FranDias (commented):
Secret-redaction scope and approach are solid. The HTTP boundary is well-covered, tests are thorough, and the at-creation redaction in `ledger.go` is the right level to intercept secrets. Four in-hunk issues and one critical out-of-hunk omission below.

---
Findings: 🔴 0 critical (in-hunk) · 🟠 4 important · 🟡 0 suggestions

**Findings dropped (out of diff hunks):**

🔴 **CRITICAL** `internal/team/broker_queues.go:38` — `Broker.Signals()` and `Broker.Decisions()` still use `copy(out, b.x)` with no sanitization, even though `Broker.Actions()` directly above was updated and the three HTTP handlers (`handleSignals`, `handleDecisions`, `handleActions`) were all updated. `QueueSnapshot()` (lines 75–86) also builds `Actions`, `Signals`, and `Decisions` directly from the raw in-memory slices without calling any sanitize helper. Every in-process Go caller — TUI, launcher, any test that calls `b.Signals()` or `b.QueueSnapshot()` — gets unsanitized records. Fix: mirror the `Actions()` loop for `Signals()` and `Decisions()`, and replace the three raw `append([]T(nil), b.x...)` calls in `QueueSnapshot()` with per-record sanitize loops.

internal/team/broker_publish.go
najmuzzaman-mohammad: Fixed in commit 55c99727 — `publishActivityLocked` now calls `activity.Activity = redactSecretsInText(activity.Activity)` before publishing (broker_publish.go:58), matching the Detail redaction symmetrically.

internal/team/broker_streams_test.go
najmuzzaman-mohammad: Fixed in commit 55c99727 — the select/time.After guard was removed from `TestBrokerMessageSubscribersReceiveRedactedMessages` (and the companion test). Both now use a direct `got := <-out` receive, which is deterministic since `PostMessage`/`PushTask` return only after the value is enqueued in the buffered subscriber channel.

internal/team/broker_human.go
najmuzzaman-mohammad: Fixed in commit d63c3190 — `humanInterview` now carries `Redacted bool`, `RedactionCount int`, and `RedactionReasons []string`. `sanitizeHumanInterview` accumulates redaction counts from each field through a shared counter and sets all three metadata fields before returning. The UI receives the same redaction badge the message bubbles already show.

internal/team/broker_persistence.go
najmuzzaman-mohammad: Fixed in commit 55c99727 — `loadState` now sanitizes watchdogs (`sanitizeWatchdogAlert`), tasks (`sanitizeTeamTask`), and scheduler jobs (`sanitizeSchedulerJob`) when loading from disk, parallel to the existing message/request/action/signal/decision sanitization. Both the read path (loadState) and write path (saveLocked) now run all sanitizers.

internal/team/broker_persistence.go
```diff
@@ -159,12 +159,37 @@ func (b *Broker) loadState() error {
...
 	for i := range b.tasks {
 		b.tasks[i].Channel = channel.MigrateDMSlugString(b.tasks[i].Channel)
+		b.tasks[i] = sanitizeTeamTask(b.tasks[i])
+	}
+	for i := range b.watchdogs {
+		b.watchdogs[i] = sanitizeWatchdogAlert(b.watchdogs[i])
+	}
+	for i := range b.scheduler {
+		b.scheduler[i] = sanitizeSchedulerJob(b.scheduler[i])
 	}
```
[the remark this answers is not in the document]
najmuzzaman-mohammad: Fixed in commit 348cba50 — `sanitizeSchedulerJob` no longer touches `Payload` since it contains JSON that must remain parseable at runtime (`json.Unmarshal` in scheduler.go). The commit also extracts a `RedactedBadge` helper to consolidate the redaction reasoning.

najmuzzaman-mohammad (approved):
All CodeRabbit threads addressed. CI is green on 455e7bab. Ready to merge.

PR #674, opened 2026-05-05
follow-ups: agent-event-bubbles deferred TODOs from PR #664

najmuzzaman-mohammad:
## Summary

Closes 4 of the 5 deferred TODOs from PR #664 (agent event bubbles). The 5th (Tier 2 hover peek card) remains design-deferred and is intentionally not in scope.

- **fix(broker):** mirror the empty-From guard from `appendMessageLocked` into `bootstrapHumanHasPostedLocked` so a corrupted/legacy persisted message with empty `From` cannot falsely flip `humanHasPosted=true` on broker restart. Caught while writing the restart test below — it would have shipped silently.
- **test(broker):** pin watchdog → `agentActivitySnapshot.Kind="stuck"` end-to-end, including the concurrent-alert guard (resolving one alert leaves Kind=stuck while another alert on the same owner is still active).
- **test(broker):** pin `bootstrapHumanHasPostedLocked` durable-restart behavior across close + reopen, with explicit empty-string and whitespace-only From cases.
- **refactor(web):** move the `fallbackTask` whitespace-trim defense from `AgentEventPill` to the `useOfficeMembers` `select` callback. Single source of truth; consumer guard removed.
- **test(web):** add `computePillState` dim-state coverage for routine + milestone hold boundaries and the tick-interval edge at exact hold expiry. CSS handles `prefers-reduced-motion` itself; the JS guarantee tested here is that `data-state="dim"` is emitted at the correct wall-clock offset for the media query to act on.

Per the PR #664 ship plan, the **Tier 2 hover peek card** TODO stays deferred — it needs a `/design-consultation` pass before any code lands.

## Test plan

- [x] `bash scripts/test-go.sh ./internal/team` (full package, 168s, all green)
- [x] Watchdog/bootstrap tests pass with `-race -count=2`
- [x] `bunx tsc --noEmit` clean
- [x] `bash scripts/test-web.sh` for touched files (30/30 passed across `agentEventTimer.test.ts`, `officeIdleDictionary.test.ts`, `AgentEventPill.test.tsx`)
- [x] `bunx biome check src/` — no new warnings on touched files (pre-existing baseline noise on unrelated files unchanged)
- [ ] CI green
- [ ] CodeRabbit review addressed

## Notes for reviewer

- The bootstrap test had to call `bootstrapHumanHasPostedLocked()` explicitly after `loadState()` because `NewBrokerAt` skips `loadState` in test mode (`skipBrokerStateLoadOnConstruct`), so the in-constructor bootstrap call sees an empty slice. Worth a separate cleanup at some point — out of scope here.
- Bisectable commit ordering: production fix → its test → next test → web refactor → its smoke. Each commit is independently revertable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect human-activity detection after restarts when messages had empty or whitespace-only sender fields.
  * Prevented whitespace-only fallback tasks from showing as idle copy.

* **Improvements**
  * Normalize member task values so blank/whitespace tasks are treated as empty.
  * Refined UI idle/dim transition timing to better respect reduced-motion preferences.

* **Tests**
  * Added test coverage for restart/persistence, watchdog activity state transitions, and UI timer state boundaries.

### DIFF
--- a/internal/team/broker_bootstrap_human_posted_test.go
+++ b/internal/team/broker_bootstrap_human_posted_test.go
@@ -1,0 +1,201 @@
+package team
+
+// Tests for bootstrapHumanHasPostedLocked — the function that re-seeds
+// b.humanHasPosted from the persisted message log when a broker restarts.
+// Four cases:
+//
+//  1. Empty log → restart → humanHasPosted == false.
+//  2. Human message in log → restart → humanHasPosted == true.
+//  3. Empty/whitespace From → restart → humanHasPosted stays false
+//     (pins the adversarial-fix: isHumanMessageSender("") returns true,
+//     but the empty-From guard in appendMessageLocked and bootstrap must
+//     prevent a false positive).
+//  4. Agent-only message in log → restart → humanHasPosted == false.
+//
+// Each test uses the standard newTestBroker/saveLocked/reloadedBroker
+// pattern (same as TestBrokerSurfaceMetadataPersists and friends) so
+// there is no need to touch production code.
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+// metaHumanHasPosted GETs /office-members on the given broker via the
+// real handler and decodes meta.humanHasPosted from the JSON response.
+// It is a strict helper: any unexpected status or decode failure is
+// fatal, not just an error.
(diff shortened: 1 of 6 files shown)


FranDias (approved):

PR #675, opened 2026-05-05
fix(wiki): retryable wiki worker init — transient failures no longer permanently disable notebook/review

najmuzzaman-mohammad:
## Problem

A transient `repo.Init` failure at broker startup consumed `sync.Once` and left `wikiWorker` permanently `nil`. Every `/notebook/*` and `/review/*` endpoint returned 503 for the lifetime of the process — even after the underlying issue (git missing from PATH, parent dir permission flap, fsck+backup double-fault) was resolved. Operators had to restart the broker to recover.

Two additional symptoms from the same root:
- `/health` reported `memory_backend_ready: true` even when `wikiWorker == nil`, masking the failure from the status bar and operators
- 503 responses had no diagnostic detail — operators had to grep broker stdout to find the real error

## Solution

Replace `sync.Once` with a `sync.Mutex` + `wikiInitErr error` so any caller can retry:

| Before | After |
|---|---|
| `wikiOnce.Do(b.initWikiWorker)` | `wikiInitMu.Lock(); if worker == nil { b.initWikiWorker() }` |
| First failure → permanent nil | First failure → retries on next request |
| `/health` always reports ready | `/health` gates on `wikiWorker != nil` |
| 503 body: generic string | 503 body: includes underlying init error |

New helpers:
- `requireWikiWorker(w, label)` — retry + 503 helper for notebook handlers
- `requireReviewLog(w)` — retry + 503 helper for review handlers
- `WikiInitErr()` / `ReviewInitErr()` — accessors for last init error (used by \`/health\` and 503 bodies)

## CodeRabbit follow-ups (round 2)

- **Retry rate-limit guard.** \`wikiInitRetryInterval\` (1s) prevents request-storm thrash when init fails on a slow path (NFS hang, DNS-blocked git subprocess) — concurrent requests within the window see the cached error instead of piling onto the broken filesystem under \`wikiInitMu\` for 30s each.
- **\`reviewInitErr\` separation.** \`ensureReviewLog\` now returns its underlying error and stores it in a new \`reviewInitErr\` field. \`requireReviewLog\` attributes wiki failures under the \"wiki\" label and review-log failures under \"review\" so 503 bodies point operators at the correct directory.
- **\`MemoryBackendError\` on /health.** Adds an \`omitempty\` field surfacing \`WikiInitErr()\` so operators see the failure root cause without grepping broker stdout.
- **\`plantWikiInitBlocker\` helper.** Extracts the regular-file-as-\`.wuphf\` blocker setup into one shared test helper used by both lifecycle and notebook tests.
- **\`resetWikiInitRetryGuardForTest\`.** Test-only helper that clears \`lastWikiInitAttempt\` so retry tests can drive the synchronous self-heal path without sleeping past the rate-limit window (no test sleeps; CI lint passes).

## Tests

Regression tests in \`broker_wiki_lifecycle_test.go\`:

1. **\`TestEnsureWikiWorkerRetriesAfterInitFailure\`** — plants a blocker file, verifies worker stays nil + error is set, removes blocker, verifies retry brings worker up and clears error, verifies idempotency
2. **\`TestNotebookHandler503SelfHealsAfterInitFailure\`** — full HTTP round-trip: first request 503 with error in body, clear blocker, second request 200
3. **\`TestHealthMemoryBackendReadyReflectsWikiWorker\`** — verifies \`/health\` reports \`memory_backend_ready: false\` AND surfaces \`memory_backend_error\` when markdown is active but worker init failed
4. **\`TestBrokerNotebookServiceUnavailable\`** rebased onto \`plantWikiInitBlocker\` so it actively blocks init now that handlers retry

Full \`internal/team\` package green race-on (148s). gofmt + go vet + golangci-lint clean.

## Test plan
- [ ] \`go test -race -run 'TestEnsureWikiWorker|TestNotebookHandler503|TestHealthMemoryBackend' ./internal/team/\`
- [ ] \`go test -race -run 'TestNotebook|TestReview|TestHealth' ./internal/team/\` (no regressions)
- [ ] Manual: start broker without git in PATH, confirm 503 with error in body; install git, retry, confirm 200 without restart

## Release Notes

* **Bug Fixes**
  * Health endpoint now accurately reflects actual runtime readiness for markdown-backed storage.
  * Improved error handling for unavailable backend services with clearer error messages.
  * Enhanced initialization robustness with retry capability and persistent error tracking.

* **Tests**
  * Added regression tests for wiki initialization recovery and health status verification.


FranDias (approved):

PR #722, opened 2026-05-07
fix(web): tasks page flickers and loses done tasks on refresh

### LINKED ISSUE #718: Tasks page blank on page refresh due to API race condition
On browser refresh, the Tasks page shows "No tasks yet" briefly before tasks appear (or stays blank until navigating away and back). The root cause is that React Query fires GET /tasks before initApi() has finished fetching the auth token. The request fails silently, React Query caches the empty/error result, and the 10-second refetch interval eventually recovers.

This affects all data-fetching hooks that mount before apiReady is true, not just tasks. The useBrokerEvents hook is gated on apiReady, but useOfficeTasks is not.

abeeb-rai:
## Summary

The Tasks page had two issues: it showed blank on page refresh, and done tasks would disappear and reappear in a ~10-second flicker loop.

### Root cause: React Query cache key collision
`RuntimeStrip`, `WorkspaceSummary`, and `TasksApp` all shared the same `["office-tasks"]` cache key but fetched with different `includeDone` params. `RuntimeStrip` fetched with `includeDone: false` (active tasks only) and its result overwrote `TasksApp`'s `includeDone: true` cache — wiping all done tasks. Every refetch cycle, they'd fight: TasksApp shows 170 tasks → RuntimeStrip overwrites → drops to 9 → TasksApp refetches → back to 170.

**Fix**: `RuntimeStrip` and `WorkspaceSummary` now use their own cache key `["office-tasks-active"]`.

### Secondary: nil slice serialization
`Broker.Reset()` and `loadState()` could set `b.tasks` to `nil`. Go serializes nil slices as JSON `null`, which could cause the frontend to receive an empty task list on edge cases.

**Fix**: initialize `b.tasks` to `[]teamTask{}` instead of `nil`.

### Secondary: upgrade-changelog 400 on dev builds
The `UpgradeBanner` component fetched `/upgrade-changelog?from=dev&to=X` on dev builds. The broker rejects `"dev"` as an invalid semver range, returning a 400 that showed as a console error on every page refresh.

**Fix**: skip the changelog fetch when either version is a dev build.

Fixes #718

## Test plan
- [ ] Start wuphf, let agents create tasks (both active and done)
- [ ] Navigate to Tasks page — all tasks should display immediately
- [ ] Refresh browser — tasks should appear without delay or flicker
- [ ] Done tasks should remain visible and not disappear/reappear
- [ ] No `upgrade-changelog` 400 error in browser console
- [ ] `curl /tasks?viewer_slug=human&all_channels=true&include_done=true` returns `"tasks": []` not `"tasks": null`

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented rare errors by ensuring task lists are always initialized (no nil task state).
  * Fixed changelog display during development-version upgrades.

* **Improvements**
  * Unified and optimized task fetching across dashboard components with dedicated cache keys and shorter refresh intervals for fresher status.
  * Task queries now respect connection state to avoid unnecessary requests when disconnected.

* **Tests**
  * Updated tests to control connection state for reliable task-fetching behavior.


FranDias (dismissed):

FranDias (approved):

PR #708, opened 2026-05-06
feat(team): public-tunnel hardening — passcode + rate limit + audit (Phase 2)

najmuzzaman-mohammad:
Stacks on **#706 (Phase 1)**. Targets that branch so the diff here is
just the Phase 2 surface; flip the base to `main` once #706 merges.

## Why

Phase 1 made the tunnel one click and bundled cloudflared, but the
trycloudflare URL was still the only secret protecting the invite. A URL
that ends up in a public Slack channel, a wayback crawl, or an
over-the-shoulder screenshot is enough to redeem the invite within the
24h window. Phase 2 adds a second factor and a brake.

### Server (`cmd/wuphf`)

- **`share_join_guard.go`** — new file with the gate primitives:
  - `generatePasscode()` mints a 6-digit numeric secret with `crypto/rand`.
  - `constantTimeCompare` wraps `subtle.ConstantTimeCompare` and equalises
    the wall-clock cost of a length-mismatch path.
  - `joinRateLimiter` — per-source-IP token bucket (5-burst, refill 1
    every 12s, lazy GC). No new dep.
  - `extractJoinSourceIP` — trusts `Cf-Connecting-Ip` ONLY when the
    immediate RemoteAddr is loopback (i.e. the tunnel-mode share server).
    On the network-share path the header is ignored so a LAN attacker
    can't spoof their own bucket key.
- **`share.go`** — `newShareHandler` now takes `shareHandlerConfig` with
  optional `JoinGate` + `RateLimiter`. `nil` = Phase 1 byte-for-byte.
  Failed gate attempts emit one audit-log line with the **truncated**
  token prefix + source IP — no full token, so logs don't become an
  oracle for stolen tokens.
- **`tunnel.go`** — `webTunnelController` owns a `passcodes` map keyed by
  minted invite token. Each `mintInviteLocked` draws a fresh passcode;
  `stop()` zeroes the whole map. `joinGate` enforces "token must be one
  this tunnel issued" — an attacker with a network-share token cannot
  ride the tunnel even if they reach it. Adds tunnel-lifecycle audit
  logs.
- **`internal/team/broker_web_tunnel.go`** — `WebTunnelStatus.Passcode`
  added (`omitempty` so network-share status responses are unchanged).

### Client (`web/src`)

- **`api/joinInvite.ts`** — `passcode` is an optional submit field; new
  closed-set error codes `passcode_required` and `rate_limited`.
- **`components/join/JoinPage.tsx`** — passcode field is hidden by default
[...]

PR #723, opened 2026-05-07
refactor(transport): host-driven outbound dispatcher for telegram

najmuzzaman-mohammad:
## Summary

Closes leftover transport-contract item (3): `telegram.go:333` polled `Broker.ExternalQueue("telegram")` from inside its own goroutine, contradicting the Transport.Send contract doc which states *"The Host calls Send from a per-transport worker goroutine (not the broker's hot path)."* This PR moves the polling lifecycle into a generic Host-side dispatcher the launcher owns, while keeping all telegram-specific bits (formatting, slug→chat lookup, typing indicator) on the adapter.

## What changed

**`internal/team/broker_outbound_dispatch.go`** (new) — `runOutboundDispatcher(ctx, broker, name, formatter, sender)`. Polls `broker.ExternalQueue(name)` every 2s, runs each msg through the adapter's `formatter` (returns `ok=false` to skip), then calls `sender`. Logs send errors and continues — at-least-once-with-drop matches the prior in-adapter semantics, since `ExternalQueue` already marks messages delivered on dequeue. Future channel-bound adapters wire in by passing two thin functions instead of writing their own polling loop.

**`internal/team/telegram.go`**
- `Run`: drops the `drainOutbound` goroutine. The errCh size shrinks from 2 to 1 (only `pollInbound` remains).
- `Send`: sends a typing indicator before the message body so chats see the same UX (brief typing bubble → message). Typing failures are silent — UX nicety, not a delivery prerequisite.
- `FormatOutbound` (new): converts `channelMessage` → `transport.Outbound`. Returns `ok=false` when no chat is mapped (routine skip, not an error). No side effects.
- `typingLoop` stays adapter-side because it polls the broker's tagged-agent state, not the outbound queue — leaking that semantic to the Host would couple the generic dispatcher to telegram-specific "is anyone tagged" logic.

**`internal/team/launcher_transports.go`** — starts `runOutboundDispatcher` in its own goroutine alongside `Run`, with proper teardown ordering: `cancel()` → wait `Run` → wait dispatcher → `broker.Stop()`.

## Tests

`internal/team/broker_outbound_dispatch_test.go` (new), 3 cases:
- `TestOutboundDispatcherFormatsAndSendsEachQueuedMessage` — happy path, including `ok=false` skip semantics. Channel-rendezvous on `sendCalls.Add(1) == 2` so the assertion is deterministic (no sleep).
- `TestOutboundDispatcherLogsButContinuesOnSendError` — a transient sender error does not stop the loop; subsequent messages on later ticks still get delivered.
- `TestOutboundDispatcherNilBrokerIsNoop` — defensive guard for misconfigured callers.

## Why this slice (and what's NOT in it)

After PR #699 closed presence end-to-end, two transport-contract leftovers remained:
1. ✅ **This PR** — telegram outbound dispatch contract honored. Pure refactor; no user-visible behavior change.
2. **`ShareTransport.Send` push channel** — separate work, will stack on #708 (which stacks on #706) since they touch `cmd/wuphf/share.go` heavily and need to settle first.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run ./internal/team/...` (0 issues)
- [x] `go test ./internal/team` (229s, ok)
- [x] All 3 new dispatcher tests pass
- [ ] Manual: launch `wuphf-dev` with a connected Telegram channel; tag an agent; observe typing indicator → reply lands in the chat exactly as before. No log noise from dropped messages.

**Note on race-mode flake:** `bash scripts/test-go.sh` (race enabled) intermittently flags a data race in `internal/team/headless_codex_queue.go:566` (`Launcher.beginHeadlessCodexTurn`) and a slow `TestExtractionSurvivesReboot` timeout. Both reproduce on a clean origin/main worktree (with this PR's changes stashed), so they are pre-existing flakes — not introduced by this refactor. Will file separately.

## Summary by CodeRabbit

* **Refactor**
  * Improved outbound message delivery architecture for enhanced reliability and consistency in message handling.
  * Enhanced typing indicator behavior to provide more timely user feedback during message transmission.

### DIFF
--- a/internal/team/broker_outbound_dispatch.go
+++ b/internal/team/broker_outbound_dispatch.go
@@ -1,0 +1,87 @@
+package team
+
+// broker_outbound_dispatch.go owns the per-transport outbound dispatcher loop
+// that the launcher runs alongside Run for adapters that publish to an
+// external surface (Telegram today, future channel-bound adapters tomorrow).
+//
+// This honors the transport.Transport contract intent — "The Host calls Send
+// from a per-transport worker goroutine (not the broker's hot path)" — by
+// moving the polling lifecycle out of each adapter and into a shared loop the
+// Host owns. Adapters supply two thin functions:
+//
+//   formatter(channelMessage) (transport.Outbound, bool)
+//       Adapter-specific conversion. Returns ok=false to skip a message
+//       (e.g. the channel slug has no chat mapped). No side effects.
+//
+//   sender(ctx, transport.Outbound) error
+//       The adapter's Transport.Send method (passed by value as a method
+//       expression: t.Send) — actually delivers to the wire.
+//
+// The dispatcher polls broker.ExternalQueue(name) every interval, formats
+// each message, and calls sender. Send errors are logged and the loop
+// continues to the next message — at-least-once delivery semantics already
+// live in the broker's ExternalQueue (which marks messages as delivered on
+// dequeue), so a transient send failure drops the message rather than
+// retrying. This matches the prior in-adapter behavior.
+
+import (
+	"context"
+	"log"
+	"time"
+
+	"github.com/nex-crm/wuphf/internal/team/transport"
+)
+
+// outboundDispatchInterval is the polling cadence. Matches the prior
+// in-adapter loop (2s) so behavior is byte-identical at the SLO level for
+// the existing telegram path. If we ever need pushed delivery (lower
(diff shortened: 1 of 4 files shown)


FranDias (dismissed):

FranDias (dismissed):

FranDias (approved):

PR #728, opened 2026-05-07
fix(scheduler): strip stale rate-limit marker from task Details on resume

najmuzzaman-mohammad:
## Summary

Fixes an infinite-resume notification loop on `task-1` / `client-loop` (33+ observed occurrences in production logs).

`Broker.ResumeTask` cleared `task.Blocked` but left the prior `Retry after <ts>` line in `task.Details`. On the next live-external attempt the CEO agent got 429'd again, re-blocking the task with a fresh `RESOURCE_EXHAUSTED` line. The watchdog's `externalWorkflowRetryAfter` check then re-detected the **stale** timestamp once it expired and called `ResumeTask` again — producing a self-driving loop that re-delivered approval prompts to the CEO.

## Fix

- New `stripExternalRetryMarker` helper in `internal/team/broker_studio.go`, plus the `externalRetryAfterLinePattern` regex it uses (anchored to lines containing `429` or an RFC3339 retry timestamp).
- `Broker.ResumeTask` now calls `stripExternalRetryMarker(task.Details)` whenever it transitions `task.Blocked` from true to false, so a subsequent re-block writes a fresh marker rather than racing the stale one.

## Test plan

- [x] `go test -race -run TestBrokerResumeTaskStripsRateLimitMarker ./internal/team/` — passes (1.7s)
- [x] `gofmt -l` clean on all 3 changed files
- [x] `go vet ./internal/team/...` clean
- [x] `go build ./cmd/wuphf` clean

## Files

- `internal/team/broker_studio.go` — pattern + helper
- `internal/team/broker_tasks_lifecycle.go` — call site in `ResumeTask`
- `internal/team/broker_tasks_test.go` — `TestBrokerResumeTaskStripsRateLimitMarker` regression coverage

## Summary by CodeRabbit

* **Bug Fixes**
  * Clear stale external rate-limit marker lines when resuming blocked tasks to prevent immediate re-triggering of watchdog resume loops and improve scheduler reliability.

* **Tests**
  * Added regression and unit tests ensuring resume strips only standalone retry-marker lines (preserving other content) and that behavior is correct across various multiline scenarios.

### DIFF
--- a/internal/team/broker_studio.go
+++ b/internal/team/broker_studio.go
@@ -31,7 +31,11 @@ var externalRetryAfterPattern = regexp.MustCompile(`(?i)retry after ([0-9]{4}-[0
 // resumed. Without stripping, a stale timestamp left in Details is detected
 // by externalWorkflowRetryAfter on the next scheduler tick, re-resuming the
 // task and re-delivering it to the CEO — producing an infinite loop.
-var externalRetryAfterLinePattern = regexp.MustCompile(`(?im)^[^\n]*(?:429[^\n]*|retry after [0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9:.+-]+Z?)[^\n]*$\n?`)
+//
+// The `\b429\b` word boundary keeps the 429-status branch from matching
+// substrings like "4290 items" or "v4.29" that legitimate task Details may
+// contain. The retry-after branch is already tightly anchored to RFC3339.
+var externalRetryAfterLinePattern = regexp.MustCompile(`(?im)^[^\n]*(?:\b429\b[^\n]*|retry after [0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9:.+-]+Z?)[^\n]*$\n?`)
 
 // stripExternalRetryMarker removes rate-limit lines from a task Details string
 // so stale retry-after timestamps cannot re-trigger the watchdog resume loop.
--- a/internal/team/broker_studio.go
+++ b/internal/team/broker_studio.go
@@ -26,6 +26,20 @@ import (
 // header back to the client. (?i) makes the prefix case-insensitive.
 var externalRetryAfterPattern = regexp.MustCompile(`(?i)retry after ([0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9:.+-]+Z?)`)
 
+// externalRetryAfterLinePattern matches a full line containing a rate-limit
+// marker so it can be stripped from task Details when a blocked task is
+// resumed. Without stripping, a stale timestamp left in Details is detected
+// by externalWorkflowRetryAfter on the next scheduler tick, re-resuming the
+// task and re-delivering it to the CEO — producing an infinite loop.
+var externalRetryAfterLinePattern = regexp.MustCompile(`(?im)^[^\n]*(?:429[^\n]*|retry after [0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9:.+-]+Z?)[^\n]*$\n?`)
+
+// stripExternalRetryMarker removes rate-limit lines from a task Details string
+// so stale retry-after timestamps cannot re-trigger the watchdog resume loop.
+func stripExternalRetryMarker(details string) string {
+	stripped := externalRetryAfterLinePattern.ReplaceAllString(details, "")
+	return strings.TrimSpace(stripped)
+}
+
 // externalWorkflowRetryAfter extracts a retry-at timestamp from an
 // external workflow provider's error message. Returns (zero, false)
 // when err is nil, when the pattern doesn't match, or when the
--- a/internal/team/broker_studio.go
+++ b/internal/team/broker_studio.go
@@ -26,22 +26,36 @@ import (
 // header back to the client. (?i) makes the prefix case-insensitive.
 var externalRetryAfterPattern = regexp.MustCompile(`(?i)retry after ([0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9:.+-]+Z?)`)
 
-// externalRetryAfterLinePattern matches a full line containing a rate-limit
-// marker so it can be stripped from task Details when a blocked task is
-// resumed. Without stripping, a stale timestamp left in Details is detected
-// by externalWorkflowRetryAfter on the next scheduler tick, re-resuming the
-// task and re-delivering it to the CEO — producing an infinite loop.
+// stripExternalRetryMarker removes rate-limit marker lines from a task Details
+// string so a stale "Retry after <ts>" timestamp cannot re-trigger the
+// watchdog resume loop on the next scheduler tick.
 //
-// The `\b429\b` word boundary keeps the 429-status branch from matching
-// substrings like "4290 items" or "v4.29" that legitimate task Details may
-// contain. The retry-after branch is already tightly anchored to RFC3339.
-var externalRetryAfterLinePattern = regexp.MustCompile(`(?im)^[^\n]*(?:\b429\b[^\n]*|retry after [0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9:.+-]+Z?)[^\n]*$\n?`)
-
-// stripExternalRetryMarker removes rate-limit lines from a task Details string
-// so stale retry-after timestamps cannot re-trigger the watchdog resume loop.
+// The predicate is externalRetryAfterPattern — the same regex the watchdog uses
+// in externalWorkflowRetryAfter to detect a re-resume condition. Reusing that
+// pattern keeps a single source of truth for marker grammar: if the watchdog
+// would not re-resume from a given line, we do not strip it. This narrows the
+// strip to lines that carry a parseable RFC3339 retry-after timestamp and
+// avoids over-stripping unrelated content (e.g. ticket refs like "SUP-4290",
+// version strings like "v4.29", line counts like "completed 4290 items").
+//
+// Known limitation: when an external provider packs both the rate-limit
+// timestamp and an operator note onto a single line (e.g. "429 ... Retry
+// after <ts>. Owner: refresh OAuth"), the whole line is dropped together. A
+// future refinement could substitute just the marker substring; the loop fix
+// does not require that today.
 func stripExternalRetryMarker(details string) string {
-	stripped := externalRetryAfterLinePattern.ReplaceAllString(details, "")
-	return strings.TrimSpace(stripped)
+	if details == "" {
+		return details
+	}
+	lines := strings.Split(details, "\n")
+	kept := make([]string, 0, len(lines))
+	for _, line := range lines {
+		if externalRetryAfterPattern.MatchString(line) {
+			continue
+		}
+		kept = append(kept, line)
+	}
+	return strings.TrimSpace(strings.Join(kept, "\n"))
 }
 
 // externalWorkflowRetryAfter extracts a retry-at timestamp from an
--- a/internal/team/broker_tasks_lifecycle.go
+++ b/internal/team/broker_tasks_lifecycle.go
@@ -92,6 +92,10 @@ func (b *Broker) ResumeTask(taskID, actor, reason string) (teamTask, bool, error
 		changed := false
 		if task.Blocked {
 			task.Blocked = false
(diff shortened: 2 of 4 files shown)


FranDias (dismissed):

internal/team/broker_studio.go
```diff
@@ -26,6 +26,20 @@ import (
 // header back to the client. (?i) makes the prefix case-insensitive.
 var externalRetryAfterPattern = regexp.MustCompile(`(?i)retry after ([0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9:.+-]+Z?)`)
 
+// externalRetryAfterLinePattern matches a full line containing a rate-limit
+// marker so it can be stripped from task Details when a blocked task is
+// resumed. Without stripping, a stale timestamp left in Details is detected
+// by externalWorkflowRetryAfter on the next scheduler tick, re-resuming the
+// task and re-delivering it to the CEO — producing an infinite loop.
+var externalRetryAfterLinePattern = regexp.MustCompile(`(?im)^[^\n]*(?:429[^\n]*|retry after [0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9:.+-]+Z?)[^\n]*$\n?`)
```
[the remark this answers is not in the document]
najmuzzaman-mohammad: Fixed in 842f7409 — anchored the `429` branch with `\b` word boundaries so it only matches the standalone token, not substrings like `4290`, `1429`, or `v4.29`. Added `TestStripExternalRetryMarkerScope` covering both true positives (standalone 429 line, RFC3339 retry-after line) and the four false-positive cases you flagged, so a future loosening of the regex has to confront the tradeoff explicitly.

FranDias (approved):

internal/team/broker_studio.go
FranDias: **🔴 CRITICAL:** Bare `429` substring match destroys unrelated task context.

This regex strips any line containing the literal substring `429` — ticket refs like `SUP-4290`, task IDs, dates, line counts — all silently deleted on any resume, regardless of whether the task was ever rate-limited.

Quick repro:
```
In:  "Investigate ticket SUP-4290 about login failures"
Out: ""
```

The watchdog re-enters the loop only when `externalRetryAfterPattern` matches a parseable RFC3339 timestamp — the bare `429` token is never the trigger. Drop the `429[^\n]*` alternation and use `externalRetryAfterPattern` as the line predicate:

```go
func stripExternalRetryMarker(details string) string {
    if details == "" { return details }
    var kept []string
    for _, line := range strings.Split(details, "\n") {
        if externalRetryAfterPattern.MatchString(line) { continue }
        kept = append(kept, line)
    }
    return strings.TrimSpace(strings.Join(kept, "\n"))
}
```

FranDias: **🔴 CRITICAL:** Operator context packed on the same line as the marker is destroyed.

External providers (One, Composio) and the headless codex recovery path commonly emit single-line errors with multiple semantic clauses:
```
"Cannot send: 429 RESOURCE_EXHAUSTED, provider Composio. Retry after 2026-04-15T22:00:29.610Z. Owner needs to refresh OAuth token."
```
The whole-line regex `^[^\n]*...[^\n]*$` strips this to `""`, losing the actionable operator note. The line-by-line approach (or stripping only the matching sub-expression) preserves the surrounding content.

internal/team/broker_tasks_lifecycle.go
```diff
@@ -92,6 +92,10 @@ func (b *Broker) ResumeTask(taskID, actor, reason string) (teamTask, bool, error
 		changed := false
 		if task.Blocked {
 			task.Blocked = false
+			// Strip the stale rate-limit marker so the watchdog's
+			// externalWorkflowRetryAfter check cannot re-enter the
+			// resume loop after the task is re-blocked for a new reason.
+			task.Details = stripExternalRetryMarker(task.Details)
```
FranDias: **🟠 IMPORTANT:** Strip runs on every `Blocked→false` transition, not just rate-limit recoveries.

`ResumeTask` is called from operator manual unblocks, dependency-resolved queue releases, and capability self-healing — not only the watchdog cooldown path. With the current broad `429` matcher, any task whose details happen to include `429` (e.g. `"queued behind active lane on task-4290"`) loses that context on resume.

Adopting `externalRetryAfterPattern` as the predicate (requires a parseable RFC3339 timestamp) narrows the strip to actual rate-limit markers automatically.

internal/team/broker_studio.go
```diff
@@ -26,6 +26,20 @@ import (
...
+
+// stripExternalRetryMarker removes rate-limit lines from a task Details string
```
FranDias: **🟠 IMPORTANT:** `externalRetryAfterLinePattern` duplicates the RFC3339 grammar from `externalRetryAfterPattern` and will drift.

When a provider starts emitting sub-second precision or `+00:00`-style offsets and one regex is updated, the other diverges — either re-introducing the loop (strip misses the marker) or over-stripping (strip catches things the watchdog does not). Single source of truth: use `externalRetryAfterPattern` as the predicate.

internal/team/broker_tasks_test.go
```diff
@@ -918,6 +918,48 @@ func TestBrokerResumeTaskUnblocksAndSchedulesOwnerLane(t *testing.T) {
 	}
 }
 
+// TestBrokerResumeTaskStripsRateLimitMarker guards against the phantom
```
FranDias, line 921: **🟠 IMPORTANT:** Test coverage is happy-path only. Two failure modes are unexercised:

1. **False-positive substring** — `Details = "Investigate ticket SUP-4290"` should NOT be stripped on resume. Add a test asserting `.Details` is unchanged after `ResumeTask`.
2. **Mixed-content line** — `Details = "429 RESOURCE_EXHAUSTED. Retry after <ts>. Owner: refresh OAuth."` should preserve the operator note post-strip (or at minimum document the current all-or-nothing behaviour).

Also missing: assert `resumed.Details` still contains pre-block context that was not part of the rate-limit marker, proving over-stripping didn't happen.

internal/team/broker_studio.go
FranDias: **🟡 SUGGESTION:** The RFC3339 charset `[0-9:.+-]+Z?` admits strings like `::::::` or `+++++`. `externalRetryAfterPattern` shares this looseness but is compensated by `time.Parse` rejecting garbage. The strip regex has no such fallback — it deletes first. If the fix moves to `externalRetryAfterPattern`-based line matching this is moot; otherwise tighten the charset.

FranDias (commented):
The fix addresses a real production bug — a self-driving resume loop on rate-limited live-external tasks — by stripping the stale rate-limit marker from `task.Details` on resume. The direction is correct, but the regex implementation has serious false-positive behaviour and test gaps that need fixing before merge.

**Root issues:**
1. `externalRetryAfterLinePattern` strips *any line containing the substring `429`* — ticket refs (`SUP-4290`), dates, counters — all silently deleted on resume. The watchdog only re-enters the loop on a parseable RFC3339 timestamp, so stripping bare `429` lines adds risk for zero benefit.
2. The strip runs on every `Blocked→false` transition, not just rate-limit cooldown recoveries.
3. The regex duplicates the RFC3339 grammar from `externalRetryAfterPattern` and will drift.
4. Tests only exercise the happy path; no coverage of false-positive substr matches or mixed-content lines.

**Suggested fix for the regex — use `externalRetryAfterPattern` as the predicate:**
```go
func stripExternalRetryMarker(details string) string {
    if details == "" { return details }
    var kept []string
    for _, line := range strings.Split(details, "\n") {
        if externalRetryAfterPattern.MatchString(line) { continue }
        kept = append(kept, line)
    }
    return strings.TrimSpace(strings.Join(kept, "\n"))
}
```

[No local checkout — diff only]

---
Findings: 🔴 2 critical · 🟠 3 important · 🟡 1 suggestion

FranDias (dismissed):

internal/team/broker_studio.go
najmuzzaman-mohammad: Fixed in c9199221 — refactored stripExternalRetryMarker to filter line-by-line using externalRetryAfterPattern as the predicate (single source of truth), and dropped externalRetryAfterLinePattern entirely. A line is now stripped only when the watchdog would actually re-resume from it. SUP-4290, task-4290, v4.29 etc. are no longer affected. Same-line operator context is still dropped together with the marker line — documented as a known limitation; refining to substring substitution can come later.

najmuzzaman-mohammad: Fixed in c9199221 — refactored stripExternalRetryMarker to filter line-by-line using externalRetryAfterPattern as the predicate (single source of truth), and dropped externalRetryAfterLinePattern entirely. A line is now stripped only when the watchdog would actually re-resume from it. SUP-4290, task-4290, v4.29 etc. are no longer affected. Same-line operator context is still dropped together with the marker line — documented as a known limitation; refining to substring substitution can come later.

internal/team/broker_tasks_lifecycle.go
najmuzzaman-mohammad: Fixed in c9199221 — refactored stripExternalRetryMarker to filter line-by-line using externalRetryAfterPattern as the predicate (single source of truth), and dropped externalRetryAfterLinePattern entirely. A line is now stripped only when the watchdog would actually re-resume from it. SUP-4290, task-4290, v4.29 etc. are no longer affected. Same-line operator context is still dropped together with the marker line — documented as a known limitation; refining to substring substitution can come later.

internal/team/broker_studio.go
najmuzzaman-mohammad: Fixed in c9199221 — refactored stripExternalRetryMarker to filter line-by-line using externalRetryAfterPattern as the predicate (single source of truth), and dropped externalRetryAfterLinePattern entirely. A line is now stripped only when the watchdog would actually re-resume from it. SUP-4290, task-4290, v4.29 etc. are no longer affected. Same-line operator context is still dropped together with the marker line — documented as a known limitation; refining to substring substitution can come later.

internal/team/broker_studio.go
najmuzzaman-mohammad: Fixed in c9199221 — refactored stripExternalRetryMarker to filter line-by-line using externalRetryAfterPattern as the predicate (single source of truth), and dropped externalRetryAfterLinePattern entirely. A line is now stripped only when the watchdog would actually re-resume from it. SUP-4290, task-4290, v4.29 etc. are no longer affected. Same-line operator context is still dropped together with the marker line — documented as a known limitation; refining to substring substitution can come later.

internal/team/broker_tasks_test.go
najmuzzaman-mohammad: Addressed in c9199221: TestStripExternalRetryMarkerScope now covers the SUP-4290 / task-4290 / v4.29 false-positives, the same-line operator-context drop (documented limitation), and the empty-input case. TestBrokerResumeTaskStripsRateLimitMarker asserts pre-block operator context survives the resume. New TestBrokerResumeTaskPreservesDetailsWithoutMarker covers the manual-unblock path — Details unchanged when no marker is present.

FranDias (dismissed):

internal/team/broker_tasks_lifecycle.go
```diff
@@ -92,6 +92,10 @@ func (b *Broker) ResumeTask(taskID, actor, reason string) (teamTask, bool, error
...
 			changed = true
 		}
```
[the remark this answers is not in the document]
najmuzzaman-mohammad: Fixed in c3c15c76 — defense in depth.

1. `unblockDependentsLocked` now strips the marker at the source whenever it transitions `Blocked→false` on a dependency completion. Closes the loop deterministically without depending on a follow-up `ResumeTask` call.

2. `ResumeTask` now strips unconditionally (outside the `if task.Blocked` guard) so any future code path that sets `Blocked=false` directly is also safe via the watchdog→`ResumeTask` fallback. The strip is a no-op when no marker is present.

New regression test `TestBrokerUnblockDependentsStripsRateLimitMarker` asserts the dependency-completion path strips the marker and preserves pre-block context.

FranDias (approved):

PR #734, opened 2026-05-07
refactor(web): extract shared AgentPulsePanel and ActiveTasksPanel

najmuzzaman-mohammad:
## Summary

- Extracts `normalizeStatus`, `classifyMember`, `isAgentActive`, and `taskMeta` into `web/src/lib/officeStatus.ts` — single source of truth instead of two diverged copies
- Adds `web/src/components/apps/shared/AgentPulsePanel.tsx` — unified agent status list used by both OfficeOverviewApp and ArtifactsApp
- Adds `web/src/components/apps/shared/ActiveTasksPanel.tsx` — unified task card list with configurable badge class, limit, click handler, and empty label
- Both apps now import the shared components; ~165 lines of duplicated logic removed

No behavior change. Follow-up to #711 (office overview dashboard).

## Test plan

- [ ] All 15 `OfficeOverviewApp.test.tsx` tests pass
- [ ] TypeScript clean on touched files
- [ ] Visually verify agent pulse and active task sections render identically in both Overview and Activity apps

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added compact shared panels for showing active tasks and active agents with consistent empty states and optional interaction.
  * Added shared status utilities to standardize task status labels, agent classification, and task metadata.

* **Refactor**
  * Replaced inline task/agent lists across overview screens with the new shared panels for consistent layout and behavior.
  * Unified active-agent detection so “Agents working now” and related sections show consistent results.

PR #712, opened 2026-05-06
feat(web): agent profile pages with skills and recent activity (phase 5 pr 5)

najmuzzaman-mohammad:
## Summary

- Adds `AgentProfilePanel` component — a right-rail slide-in panel that renders a full agent profile with 8 sections: status badge, role/description, runtime/provider, owned skills, channels the agent belongs to, recent runs (linked to activity app), recent tasks, and permissions summary
- Adds a **Profile** button to `AgentPanel`'s primary actions row; clicking it switches the panel in-place to the profile view (no navigation, no new route)
- Makes message **avatar** and **author name** clickable for agent messages — both now open the agent panel (which the user can then navigate to Profile from)
- All sections handle missing data gracefully with empty-state copy

## Entry points

| Surface | How |
|---|---|
| Roster/sidebar | Click agent → panel opens → click **Profile** |
| Message avatar | Click avatar on any agent message → panel opens |
| Message author name | Click author name on any agent message → panel opens |

## Sections rendered

- **Status** — active / idle / paused badge with coloured dot
- **Role** — full role description text
- **Current task** — live task text when agent is active
- **Runtime** — harness (claude-code/codex/…) and model
- **Skills** — skills owned by this agent (active + proposed), with pending badge
- **Channels** — clickable chips navigating to each channel
- **Recent runs** — up to 8 log tasks with tool-call count and error flag; "See all activity" links to the activity app
- **Recent tasks** — up to 5 tasks owned by this agent, linked to task detail
- **Permissions** — role classification, removable flag, built-in flag (read-only in this PR)

## Test plan

- [ ] Render full-data agent profile — all sections populated
- [ ] Render minimal agent (no role, no tasks, no channels, no skills) — empty states shown, no crash
- [ ] Click Profile in AgentPanel → profile replaces quick-panel view
- [ ] Click message avatar → agent panel opens
- [ ] Click message author name → agent panel opens
- [ ] "See all activity" navigates to `/apps/activity`
- [ ] Channel chip navigates to `/channels/:slug`
- [ ] Vitest: `bash scripts/test-web.sh web/src/components/agents` — 14/14 pass

## Checks

- [x] `bunx tsc --noEmit` — clean
- [x] `bunx biome check` — no errors on changed files
- [x] `bunx secretlint` — clean
- [x] `bash scripts/test-web.sh web/src/components/agents` — 14/14
- [x] `bash scripts/check-complexity-baseline.sh` — exit 0

[...]

PR #737, opened 2026-05-08
fix(web): profile panel run rows display-only + normalizeStatus null guard

najmuzzaman-mohammad:
## Summary

Follow-up fixes for two open CodeRabbit comments from PRs #712 and #734 that were missed before merge.

- **AgentProfilePanel** (`web/src/components/agents/AgentProfilePanel.tsx`): per-run buttons in `RecentRunsSection` navigated to the same `/apps/activity` URL as the "See all activity" button, making the per-row click misleading — it implied a specific run detail view that doesn't exist. Converts run rows to non-interactive `<div>` elements. Renames `.agent-profile-run-btn` → `.agent-profile-run-row` in CSS and splits interactive-only styles (`cursor`, `hover`, `transition`) to the task button selector only.

- **officeStatus** (`web/src/lib/officeStatus.ts`): `normalizeStatus` calls `raw.toLowerCase()` without a null/undefined guard. Adds `if (!raw) return ""` early return, matching the defensive pattern in `normalizeTaskStatus`.

## Test plan

- [ ] `bash scripts/test-web.sh web/src/components/agents/AgentProfilePanel.test.tsx` — 12/12 pass
- [ ] Verify agent profile panel renders run rows as non-clickable text (no pointer cursor, no hover color change)
- [ ] Verify "See all activity" button still navigates to `/apps/activity`
- [ ] Verify `normalizeStatus("")`, `normalizeStatus(null as any)` return `""` without throwing

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved status handling for empty or invalid input.

* **Refactor**
  * Agent profile recent runs are now displayed as non-interactive rows with task and error information instead of navigable buttons.

### DIFF
--- a/web/src/components/agents/AgentProfilePanel.tsx
+++ b/web/src/components/agents/AgentProfilePanel.tsx
@@ -150,22 +150,13 @@ function RecentRunsSection({ runs, loading }: RecentRunsSectionProps) {
               key={r.taskId}
               className="agent-profile-list-item agent-profile-run-item"
             >
-              <button
-                type="button"
-                className="agent-profile-run-btn"
-                onClick={() =>
-                  void router.navigate({
-                    to: "/apps/$appId",
-                    params: { appId: "activity" },
-                  })
-                }
-              >
+              <div className="agent-profile-run-row">
                 <span className="agent-profile-run-id">{r.taskId}</span>
                 <span className="agent-profile-run-meta">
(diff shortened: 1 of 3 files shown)


FranDias (approved):

PR #749, opened 2026-05-09
fix(tunnel): retry cloudflared bring-up on transient trycloudflare.com 1101

najmuzzaman-mohammad:
## Summary

When Cloudflare's free QuickTunnel API (`api.trycloudflare.com`) is unhealthy, it returns HTTP 500 with a non-JSON body (`error code: 1101`). `cloudflared` cannot unmarshal that, exits in under a second, and the user sees a hard failure on the first click of "Start public tunnel".

This was reproduced live during this session — `curl -X POST https://api.trycloudflare.com/tunnel` returned `500` + 16 bytes of `error code: 1101` plain text from the test machine.

The tunnel controller now retries the bring-up up to 3 times with 1.5s / 3s backoff when the failed attempt's tail matches a known-transient TryCloudflare signature.

## What changed

- **`cmd/wuphf/tunnel.go`**
  - New `cloudflaredMaxAttempts = 3` and `cloudflaredRetryBackoffStep = 1.5s`.
  - New `quickTunnelTransientPatterns` + `isTransientQuickTunnelFailure(tail)` matching the exact 1101/500 signature plus 502/503/504.
  - `start()` is restructured around a retry loop. The loopback listener and the share HTTP server are committed once and stay stable across attempts; only `cmd`/`ctx`/pipes cycle.
  - Concurrent `stop()` during the backoff sleep is detected by re-checking `c.server == server` after each backoff so the loop bails out instead of respawning into a torn-down session.
  - Non-transient failures (timeout, missing binary, ctx cancel, pipe error) skip the retry path entirely — no waiting through pointless respawns for permanent errors.
  - Final-failure error message gains a trailing hint when the signature is transient: `(after 3 attempts; trycloudflare.com appears to be having issues — try again in a minute)`.

- **`cmd/wuphf/tunnel_test.go`**
  - New `TestIsTransientQuickTunnelFailure` locks in the exact 1101/500 chain from the user's report plus 502, the benign banner, `connection refused`, and an empty tail.

## Scope notes

- This is a defense for transient blips (typically seconds). It does **not** paper over a sustained Cloudflare outage — three attempts at 1.5s/3s burns ~6s and surfaces a clearer error.
- For sustained outages, the right fix is a named tunnel or alternate provider; flagged for a follow-up if it becomes a recurring problem.
- No new packages. No public API change. Host-only code path.

## Test plan

- [x] `gofmt -l cmd/wuphf/tunnel.go cmd/wuphf/tunnel_test.go` clean
- [x] `go vet ./cmd/wuphf` clean
- [x] `golangci-lint run ./cmd/wuphf` — 0 issues
- [x] `bash scripts/test-go.sh ./cmd/wuphf` green (race + count=1 elsewhere also green)
- [x] `bunx secretlint cmd/wuphf/tunnel.go cmd/wuphf/tunnel_test.go` clean
- [ ] Reviewer: confirm the retry loop's stop()-during-backoff path (look for `c.server != server` re-check)
- [ ] Reviewer: confirm no double `cmd.Wait()` on any failure path

## Summary by CodeRabbit

* **Bug Fixes**
  * Public tunnel startup now retries on transient QuickTunnel/API failures with capped attempts and backoff while keeping the local sharing/listener stable.
  * Non-transient errors skip retries and are surfaced immediately.
  * Error messages include hints when upstream services may be unhealthy and append recent tunnel output.

* **Tests**
  * Added tests for transient-failure detection and that startup cancellation interrupts retry/backoff.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nex-crm/wuphf/pull/749)


FranDias (dismissed):

FranDias (dismissed):

FranDias (approved):

PR #789, opened 2026-05-10
feat(broker): @wuphf/broker — loopback listener with /api-token + bearer auth (rewrite branch 4)

FranDias:
## Summary

- Lands `packages/broker/` — pure-Node TS library exposing `createBroker(config) → BrokerHandle`. Binds `127.0.0.1:<port>`; emits `/api-token` (loopback + Origin + Sec-Fetch gated, no bearer), `/api/health` + `/api/events` SSE (structural default-deny bearer for the whole `/api` namespace), and a `WebSocket /terminal/agents/:slug?token=` upgrade gated on origin + token + protocol-branded slug (closes `1011 not_implemented` for branch-4). DNS-rebinding guard checks both `Host` and `RemoteAddr` via `@wuphf/protocol`. Bearer compare goes through `node:crypto.timingSafeEqual`. Static handler uses four-layer path-traversal defense including FD-locked reads (open via FileHandle, stat + stream from the handle so the FD's inode is locked against swap TOCTOU).
- Replaces the desktop hello-world `broker-stub-entry` with a real `broker-entry` that boots `@wuphf/broker`, reports `{ ready, brokerUrl }` to the supervisor, and emits `{ alive }` heartbeats + `{ broker_log, event, payload }` forwarded logs. `BrokerSupervisor` exposes `whenReady()` (BrokerUrl-branded), `subscribeReady` for restart-driven scoped window rebuild, a startup readiness watchdog with force-escalation, and sender-identity + stopping + timed-out gates on the message handler. In packaged mode the BrowserWindow loads `${brokerUrl}/` (broker serves the renderer bundle, same-origin loopback); dev mode plumbs the electron-vite origin through `WUPHF_DEV_RENDERER_ORIGIN` so the new Origin gate accepts the legitimate cross-origin bootstrap fetch.
- Wires the new package into root `package.json` scripts and `lefthook.yml` (biome, typecheck, test, custom invariants).

## Spec anchors

- `business-musings/wuphf-greenfield-rewrite-rfc-2026-05.md` §7.3 (IPC discipline) and §15 Stream A row `feat/broker-loopback-listener`.
- `docs/architecture/broker-contract.md` — the v0 broker contract carried forward unchanged.
- `@wuphf/protocol#ApiBootstrap` — `/api-token` wire shape stays `{ token, broker_url }` byte-for-byte compatible with v0; token regex narrowed to base64url-only post-triangulation; `BrokerUrl` brand added for structural proof at IPC boundaries.

## Architecture

```mermaid
[...]

PR #792, opened 2026-05-11
test(desktop): backfill broker.ts branch coverage to 100% (closes #790)

### LINKED ISSUE #790: test(desktop): bring broker.ts branch coverage back over 95

FranDias:
## Summary

Closes #790. Closes the branch-coverage gap in `apps/desktop/src/main/broker.ts` that the pass-1 + pass-2 broker hardening in #789 opened. Thresholds in `apps/desktop/vitest.config.ts` are back to **100/100/100/100**.

```mermaid
graph LR
  A["#789 merged<br/>broker.ts branches 86.24%<br/>threshold dropped 96 → 91"]
  B["#790 filed<br/>track returning to ≥95"]
  C["This PR<br/>+42 tests, 2 v8-ignores<br/>broker.ts 100%, all files 100/100/100/100"]
  D["threshold ratcheted<br/>91 → 100"]
  A --> B --> C --> D
```

### `tests/broker-helpers.spec.ts` — 25 new direct unit tests

Exported the pure helpers from `broker.ts`: `readReadyMessage`, `readBrokerLogMessage`, `sanitizeBrokerEventName`, `filterPayloadToSafeKeys`, `brokerUrlPort`, `errorMessage`, `errorCode`. One assertion per rejection arm of each helper's grammar (non-object, missing keys, wrong shape, banned keys, non-Error inputs, non-string `code`, etc.).

### `tests/broker-supervisor.spec.ts` — +13 integration tests

Each test pins a single supervisor-loop branch that the helper file can't reach:

- isAlive after first isAlive doesn't re-emit `broker_alive`.
- `whenReady()` rejects when `fatalReason` is set.
- `subscribeReady` unsubscribe is idempotent.
- `restart_skipped` with `reason: 'fatal'` when fatalReason set without stopping.
- Startup-timeout watchdog bails when conditions already settled.
- Force-timer bails when `brokerProcess` has cycled to a new fork.
- `clearStartupTimer` clears an armed force timer when `stop()` runs mid-timeout.
- `forwardBrokerLog` emits without `droppedKeys` when every key is safe.
- `handleExit` with null `startedAtMs` → `uptimeMs: null`.
- Startup-timeout with null `startedAtMs` → `uptimeMs: null`.
- Windows taskkill skipped when pid is null (both `requestProcessTermination` and `forceStop`).
- POSIX forceStop fallback when pid is null.

### `tests/main-window-security.spec.ts` — +1 test

`createSecureWindow` rejects a brokerUrl that parses cleanly but is not a loopback http URL (e.g., `http://example.com:8080`).

### Two genuinely unreachable arms wrapped in `/* v8 ignore */`

| File | Arm | Why unreachable |
|---|---|---|
| `logger.ts` rotation `existsSync(activePath)` | `false` branch | Rotation is invoked only from the size-trigger inside `appendToLogFile`, which has just written to `activePath`. The guard exists for an external delete racing rotation — not in-process reachable. |
| `window.ts` `isAllowedRendererNavigation` non-http/file fallthrough | else arm | `resolveRendererUrl` only returns http or file URLs after validation; the else-arm is belt-and-suspenders against a future change weakening that invariant. |

Each ignore site carries a comment explaining why the arm is unreachable, so future code-readers don't think coverage is just being dodged.

### Threshold bump in `apps/desktop/vitest.config.ts`

```ts
thresholds: {
  lines: 100,
  statements: 100,
  functions: 100,
  branches: 100,
},
```

A regression below this floor is now a meaningful signal: a new code path is either an untested defensive branch or an untested happy path.

## Verification

- protocol: 536 tests pass
- broker: 67 tests pass
- desktop: **193 tests pass** (was 151; +42 new tests across 3 files)
- desktop coverage: **100/100/100/100** (was 98.07/91.92/100/98.48)
- no type errors
- all 4 broker invariants pass

## Test plan

- [x] Local: `bun run test:coverage` from `apps/desktop/` — 193/193 pass at 100%.
- [x] Local: `bash scripts/test-protocol.sh` + broker suite + invariants.
- [ ] CI: protocol + broker + desktop + go-test + windows-smoke + release builds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added defensive guards for startup/restart/shutdown flows, safer log rotation, and stricter renderer navigation protocol checks.

* **Refactor**
  * Centralized IPC message parsing and log-sanitization into an internal helper to improve robustness.

* **Tests**
  * Expanded unit tests covering IPC parsing, supervisor behaviors, log handling, startup watchdogs, and secure-window validation.

* **Chores**
  * Increased test coverage thresholds and updated coverage configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nex-crm/wuphf/pull/792)


FranDias:
## Pass-3 5-lens triangulation review

Dispatched 5 orthogonal Codex agents (architecture, api, types, security, distsys) against the PR via `scripts/dispatch-triangulation.sh`. All 5 verdicts: **Hold for fixes**.

### Highest-confidence consensus (3+ lenses)

| # | File:line | Severity | Finding |
|---|---|---|---|
| 1 | `apps/desktop/tests/broker-supervisor.spec.ts:1649` + `broker.ts:574` | **BLOCK** (security, architecture, distsys) | The `restart_skipped reason='fatal'` test sets `fatalReason` via `Reflect.set` then directly invokes private `scheduleRestart`. No real flow reaches this state: `scheduleRestart` is never called after `fatalReason` is set (both `handleExit` and `handleRestartStartFailure` set fatalReason then `return` without scheduling). The branch at `broker.ts:574` (`fatalReason !== null` guard inside the restart-timer callback) is dead code. |
| 2 | `apps/desktop/tests/broker-supervisor.spec.ts:1811` + `broker.ts:453` | **BLOCK** (security) / **HIGH** (architecture, distsys, types) | The `handleExit null uptimeMs` test forges a state where `brokerProcess` is set but `startedAtMs` is null. `start()` sets `startedAtMs` BEFORE assigning `brokerProcess` and arming exit/timer paths, so when `handleExit` proceeds past the `brokerProcess !== exitedProcess` gate, `startedAtMs` is always non-null. The `uptimeMs` ternary's null arm is dead code. |
| 3 | `apps/desktop/tests/broker-supervisor.spec.ts:1833` + `broker.ts:660` | **BLOCK** (security) / **HIGH** (architecture, distsys, types) | The `startup-timeout null uptimeMs` test forges the same impossible state inside the watchdog body. The timer body's 4-condition early-return runs BEFORE reading `startedAtMs`, and any path that clears `startedAtMs` also changes `brokerProcess`, so the bail-out fires first. The `uptimeMs` ternary's null arm is dead code. |
| 4 | `broker.ts:841/868/889/893/924/968/972` | **MEDIUM** (architecture, api, security, types) | The 7 newly-exported helpers (`readReadyMessage`, `readBrokerLogMessage`, `sanitizeBrokerEventName`, `filterPayloadToSafeKeys`, `brokerUrlPort`, `errorMessage`, `errorCode`) widen `broker.ts`'s module surface. Future contributors will reasonably treat them as supported production imports. Move to `broker-internal.ts` or annotate `@internal`. |

### Second-tier (1-2 lenses)

- `tests/broker-supervisor.spec.ts:1680, 1708` — **HIGH (api) / MEDIUM (distsys)** — startup-bail and force-timer-cycle tests use `Reflect.set` to manufacture interleavings the natural cleanup paths prevent. The underlying defenses are correct; the tests don't model real races. distsys suggests rewriting around "old process exits before startup timeout fires."
- `tests/broker-supervisor.spec.ts:1757` — **MEDIUM (architecture)** — `clearStartupTimer` force-timer-clear test has no behavioral assertion. A broken implementation could still pass.
- `apps/desktop/src/main/window.ts:20/94/102` vs `packages/protocol/src/ipc.ts:293` — **MEDIUM (types)** — `CreateSecureWindowConfig.brokerUrl: string` loses the BrokerUrl brand. `isLocalHttpRendererUrl` checks protocol/host/port but not the full brand (no userinfo / no non-root path / no query / no fragment). Import `isBrokerUrl` and enforce.
- `tests/broker-helpers.spec.ts:84` + `broker.ts:863` — **MEDIUM (api)** — `readBrokerLogMessage` accepts non-object payloads and the test asserts a string payload passes through unchanged. Downstream `filterPayloadToSafeKeys` returns `{ safePayload: {}, droppedKeyCount: 0 }` for non-objects → docs imply keys are filtered/accounted, but a non-object payload yields no `droppedKeys`. Reject non-object payloads at the codec or update docs.
- `apps/desktop/docs/modules/broker-spawn.md` — **MEDIUM (api)** — IPC grammar is desktop-private but only lightly documented; a Go/Rust broker implementer would need explicit schemas, validation rules, version/upgrade behavior. Either mark exports `@internal` or promote the schemas to `@wuphf/protocol`.

### Direct disagreements

The 3 BLOCKs from security match HIGHs from architecture/distsys — security is the strictest lens because it reads "100% coverage via private-state mutation" as a security claim that's structurally false. Treating as BLOCKs.

## Disposition (this PR)

| # | Action |
|---|---|
| 1-3 | **Delete the 5 synthetic Reflect.set tests** (1649, 1680, 1708, 1811, 1833). Wrap the genuinely-dead branches at `broker.ts:453, 574, 660, 679` in `/* v8 ignore */` with invariant rationale. Rewrite the startup-timer bail test around a natural race (old process exits before startup timeout fires) per distsys's suggestion. |
| 4 | **Move 7 exported helpers** from `broker.ts` to `broker-internal.ts`. Update import in `broker-helpers.spec.ts`. |
| 5 | **Tighten `window.ts` brokerUrl** to enforce the full BrokerUrl brand via `isBrokerUrl`. Add negative tests for userinfo/path/query/fragment. |
| 6 | **Reject non-object payload in `readBrokerLogMessage`** so the IPC grammar stays map-shaped. Update tests. |
| 7 | **Add behavioral assertion** to the `clearStartupTimer` force-timer-clear test. |

Will re-run coverage after each change. Target: keep 100/100/100/100 with only reachable branches counted; treat the v8-ignored arms as documented defensive guards.

FranDias:
## Pass-3 triangulation fixes applied (commit `73d79aba`)

Addressed every consensus finding from the 5-lens review. All 5 lenses voted "Hold for fixes"; this commit closes the BLOCKs and the strongest MEDIUMs.

| # | Severity | Finding | Disposition |
|---|---|---|---|
| 1 | BLOCK (security, architecture, distsys) | `tests/broker-supervisor.spec.ts:1649` covers an unreachable `reason: "fatal"` ternary arm via `Reflect.set` + private-method invocation. | **FIXED** — deleted the test; wrapped `broker.ts:574` (fatalReason guard inside scheduleRestart's setTimeout) with `/* v8 ignore */` + invariant comment documenting why scheduleRestart is never called after fatalReason is set. |
| 2 | BLOCK (security) / HIGH (architecture, distsys, types) | `tests/broker-supervisor.spec.ts:1811` covers an unreachable `uptimeMs: null` arm in `handleExit`. | **FIXED** — deleted the test; wrapped `broker.ts:462` with `/* v8 ignore */`. `start()` sets `startedAtMs` before `brokerProcess` is assigned. |
| 3 | BLOCK (security) / HIGH (architecture, distsys, types) | `tests/broker-supervisor.spec.ts:1833` covers an unreachable `uptimeMs: null` arm in the startup-timeout body. | **FIXED** — deleted the test; wrapped `broker.ts:678` with `/* v8 ignore */`. Same invariant. |
| 4 | MEDIUM (api, distsys) | `tests/broker-supervisor.spec.ts:1680` (startup-timer stopping bail) bypasses real `stop()` cleanup. | **REWRITTEN** — now exercises the natural race (an early exit at `maxRestartRetries: 0` sets `fatalReason`, then the startup timer fires and hits the `fatalReason !== null` arm of its 4-condition bail). No `Reflect.set`. |
| 5 | MEDIUM (api, distsys) | `tests/broker-supervisor.spec.ts:1708` (force-timer cycle bail) bypasses `clearStartupTimer`. | **FIXED** — deleted the test; wrapped `broker.ts:703-711` (`brokerProcess === brokerProcess` equality check) with `/* v8 ignore */` + invariant about how `clearStartupTimer` always runs first. |
| 6 | HIGH (api) / various | `whenReady() fatalReason` test used `Reflect.set("fatalReason", "synthetic")`. | **REWRITTEN** — now uses the natural cap path: `maxRestartRetries: 0`, emit exit, `handleExit` hits cap and sets `fatalReason`, then `whenReady()` rejects. No `Reflect.set`. |
| 7 | MEDIUM (architecture) | `tests/broker-supervisor.spec.ts:1757` (`clearStartupTimer` force-timer clear) ran the path but didn't assert behavior. | **FIXED** — now mocks `killProcess`, advances past the original force deadline, and asserts no SIGKILL was issued. |
| 8 | MEDIUM (api, architecture, security, types) | The 7 newly-exported helpers in `broker.ts` widen the package's module surface. | **FIXED** — extracted into `apps/desktop/src/main/broker-internal.ts` with an `@internal` module comment. `broker-helpers.spec.ts` updated to import from the new path. `broker.ts` imports them back through the same module. |
| 9 | MEDIUM (types) | `window.ts` `brokerUrl: string` only checked protocol/host/port via `isLocalHttpRendererUrl`. The full BrokerUrl brand (no userinfo / no non-root path / no query / no fragment) wasn't enforced. | **FIXED** — `createSecureWindow` now calls `isBrokerUrl` from `@wuphf/protocol`. Added negative tests for userinfo, `/api-token` path, `?x=1` query, and `#frag` fragment. |
| 10 | MEDIUM (api) | `readBrokerLogMessage` accepted non-object payloads; downstream `filterPayloadToSafeKeys` silently dropped them with `droppedKeyCount: 0`, contradicting the documented `payload?: object` shape. | **FIXED** — `readBrokerLogMessage` now rejects non-object payloads (string/number/boolean/array/null) at the codec, returning `null`. Test updated to assert each rejection arm. |

### Deferred to follow-up

- The deeper API/architecture critique (api Finding 1) about the IPC grammar belonging in `@wuphf/protocol` rather than as desktop-private helpers is a real architectural question, but moving the codecs requires versioning + conformance fixtures and is out of scope for this PR. The `@internal` annotation on `broker-internal.ts` + the existing `broker-spawn.md` documentation are sufficient for now; the design question stays open for the next branch that introduces a non-TS broker implementation.

### Verification

- protocol: 536 tests pass
- broker: 67 tests pass
- desktop: **191 tests pass** (was 193; −4 synthetic tests deleted, +2 added; net −2)
- desktop coverage: **100/100/100/100** — covered branches all reachable through the public API; defensive arms marked `v8 ignore` with documented invariants
- no type errors
- all 4 broker invariants pass

Each `v8 ignore` site now carries an explanatory comment so a future contributor knows the arm is unreachable, not just being dodged.

FranDias:
## Staff-code-reviewer feedback addressed (commit `8f05f99c`)

Spawned the `staff-code-reviewer` agent on the cumulative diff. It surfaced one HIGH that was missed by the 5-lens triangulation:

### HIGH (FIXED in `8f05f99c`)

**`tests/broker-supervisor.spec.ts:1655-1682` — "startup-timeout watchdog bails when fatalReason was set by an early exit" passed for the wrong reason.**

The 4-condition OR-guard in the startup-timer body:

```typescript
if (
  this.brokerProcess !== brokerProcess ||  // arm 0
  this.brokerUrl !== null ||                // arm 1
  this.stopping ||                          // arm 2
  this.fatalReason !== null                 // arm 3 ← claimed target
) return;
```

short-circuits left-to-right on OR. The test's setup (exit at cap → handleExit nulls `brokerProcess` + sets `fatalReason`) makes arm 0 (`brokerProcess !== brokerProcess`) **true first**, so arm 3 is never evaluated. The test would still pass if arm 3 were deleted. v8 coverage reports 100% only because each sub-clause is EVALUATED when all are false (the natural timeout case), not because each is ever truthy.

Trace of TRUE-side reachability for each arm:
- arm 0: reachable via the test's exit-before-timeout race ✓
- arm 1 (`brokerUrl !== null`): reachable only via Node timer-microtask-already-queued race; `clearStartupTimer` runs first in normal flow.
- arm 2 (`stopping`): same race-only.
- arm 3 (`fatalReason !== null`): genuinely unreachable — every path that sets `fatalReason` also nulls `brokerProcess` first.

**Fix**: collapsed the 4-OR to its single reachable arm (`this.brokerProcess !== brokerProcess`). Updated the test comment to reflect that it exercises the sender-identity check, not `fatalReason`.

### MEDIUM (FIXED)

**`tests/broker-supervisor.spec.ts:1684-1696` — "clearStartupTimer is a no-op when no timer is armed" was redundant Reflect-based work.**

The null-arm of `clearStartupTimer` is exercised by every supervisor.start() call (start → armStartupTimer → clearStartupTimer runs with both timer fields null on the first invocation). The Reflect-based test added no coverage the suite didn't already provide AND re-introduced the private-method reflection the pass-3 review otherwise eliminated. Deleted.

### Deferred to follow-up

The reviewer flagged a few more items applicable to PR #791 (receipt write path), addressed in #791 directly: observability logging on 4xx receipt-post paths, `ReceiptStore.put` atomicity contract on the interface, and a streaming-overflow test pattern (tracked as #793).

### Verification

- 190 desktop tests pass (was 191; the redundant test was deleted)
- coverage holds at 100/100/100/100 — the OR collapse reduced the branch count from 352 → 348 (the 3 dead arms removed) but every counted branch is covered.
- bun run lint + bun run typecheck: clean

FranDias:
## Update: removed the 3 `biome-ignore lint/complexity/useLiteralKeys` suppressions

Per the repo quality bar ("Do not suppress lint warnings with ignore comments" — root `CLAUDE.md`), I dispatched a codex agent to deal with the root cause behind the three `useLiteralKeys` suppressions this PR introduced. They were papering over a contradiction between two strict rules:

- tsconfig `noPropertyAccessFromIndexSignature` (forces bracket access on `Record<string, T>`)
- biome `useLiteralKeys` (wants dot access for static string keys)

**Root fix — `bb2ea053`**:

- `apps/desktop/src/main/broker.ts`: moved `droppedKeys` into the payload literal initializer, no bracket assignment:
  ```ts
  const finalPayload: Record<string, LogPayloadValue> = {
    ...safePayload,
    ...(droppedKeyCount > 0 ? { droppedKeys: droppedKeyCount } : {}),
  };
  ```
- `apps/desktop/src/renderer/main.ts`: tightened the `parseBootstrap` cast from `Record<string, unknown>` to a known-optional-keys interface `{ readonly token?: unknown; readonly broker_url?: unknown }` — `Object.keys` still enumerates runtime keys (the unknown-key rejection loop is unchanged), and dot access now satisfies both rules.

**Verification:**
- `bunx tsc --noEmit` green
- `bunx biome check` green on the two changed files
- `bunx vitest run` → 190 tests pass
- `bunx vitest run --coverage` → 100/100/100/100 (unchanged)
- Pre-push hooks: desktop-test, desktop-typecheck, desktop-check-invariants, desktop-check-ipc-allowlist, complexity-baseline, file-size — all green

| # | Finding | Status | Notes |
|---|---------|--------|-------|
| 1 | broker.ts droppedKeys bracket access | FIXED | `bb2ea053`, literal initializer |
| 2 | renderer/main.ts `record["token"]` | FIXED | `bb2ea053`, narrowed cast + dot access |
| 3 | renderer/main.ts `record["broker_url"]` | FIXED | `bb2ea053`, same fix |

No behavior change. Pure type-narrowing refactor.

FranDias:
## Pass-4 triangulation synthesis (5 lenses: architecture, types, security, sre, electron)

Ran read-only codex agents in parallel. Reports + raw outputs at `/tmp/triangulation-pr792-pass4/`. The **security** lens timed out (status 124, no report); the other four returned.

### Convergent HIGH (3 lenses) — pass-3.1 OR-collapse re-introduced a race

**`apps/desktop/src/main/broker.ts:660-697`** — flagged independently by **electron, sre, architecture**.

The pass-3.1 collapse reduced the startup-timeout watchdog bail from a 4-OR
```ts
this.brokerProcess !== brokerProcess
  || this.brokerUrl !== null
  || this.stopping
  || this.fatalReason !== null
```
down to its first arm only. The justification (echoing the staff-reviewer's HIGH on the prior pass) was that arms 1/2/3 were "race-only or unreachable". The triangulation disagrees:

- The `{ready}` IPC handler at `broker.ts:241-244` sets `brokerUrl` **without nulling `brokerProcess`**.
- `stop()` at `broker.ts:368-385` sets `stopping` **while the process is still current**.
- Node `clearTimeout` cannot recall a callback that's already been queued onto the event loop.

So under event-loop stalls (which Electron packaged apps see in the wild — main-process I/O contention, signed-binary launch overhead, OS context switches), a startup-timeout callback queued at T-1ms can fire AFTER a `{ready}` or `stop()` win flips the secondary state but BEFORE the process handle gets nulled. The watchdog will then log `broker_ready_timeout`, terminate a broker that's already ready or stopping, and trigger spurious restart churn.

The PR explicitly **defends this exact race in the restart timer** at `broker.ts:575-591`. Stripping the same defense out of the startup watchdog is inconsistent and load-bearing.

**Verdict**: revert the pass-3.1 collapse. The 4-OR arms are not redundant; they're race-tolerant guards. The right answer to "the test was passing for the wrong reason" is a queued-timer race test — not deleting the arms. Suggested follow-up: model a `nextTick`-delivered timeout callback against `{ready}` and `stop()` racing in front of it; assert no termination occurs.

### Other HIGH (1 lens each, strong evidence)

**SRE Finding 1 — `utilityProcess.error` event has no listener (`broker.ts:275`).**
Electron documents `utilityProcess`'s `"error"` as carrying `type`, `location`, and `report` for V8 fatals **before** `"exit"` fires. We subscribe only to `"exit"`. On a 3am V8 fatal, on-call sees a generic `broker_exited` with no diagnostic shape. Suggestion: add `brokerProcess.on("error", ...)` with a safe-allowlisted log line, let `"exit"` continue to drive restart.

**Architecture + Electron Finding 2 — startup force-timer v8-ignore is reachable (`broker.ts:703-717`).**
The `/* v8 ignore */` on the `brokerProcess !== brokerProcess` arm of the force-timer claims structural unreachability. But the normal "SIGTERM worked before SIGKILL" path lands here: timeout → `requestProcessTermination` → fast exit → `handleExit` nulls `brokerProcess` → force timer fires, finds the handle has changed. That's not unreachable; that's the happy-path race. Either clear `startupForceTimer` in `handleExit`, or remove the v8-ignore and cover the path.

### MEDIUMs

**Types Finding 1 — `BrokerUrl` brand admits encoded path traversal (`packages/protocol/src/ipc.ts:293-303`).**
`isBrokerUrl("http://127.0.0.1:54321/%2e%2e")` returns `true`. `assertApiBootstrapBrokerUrl` reads `parsed.pathname` after `new URL()` normalization, so `%2e%2e` looks like an empty path to the check, but `asBrokerUrl` and `apiBootstrapToJson` round-trip the raw string. The brand is partly decorative for percent-encoded edges. Suggestion: reject unless raw input equals `parsed.origin` or `${parsed.origin}/`; add tests for `%2e`, `%2e%2e`, uppercase variants.

**Architecture Finding 3 — restart-start failure reentrancy comment is wrong (`broker.ts:635`).**
The comment says the reentrant-restart race ("forkProcess synchronously calls `stop()` then throws") is infeasible to cover. `BrokerSupervisorConfig.forkProcess` is public injection surface; the existing fake-fork pattern can make it deterministic. Cover it or remove the claim.

**Architecture + Types — renderer bootstrap parity drift (`renderer/main.ts:167`, `protocol/src/ipc.ts:229`).**
This round's `parseBootstrap` cast doesn't enforce protocol's data-property guard. The renderer reads `record.token` / `record.broker_url` directly, while `apiBootstrapFromJson` checks `Object.getOwnPropertyDescriptor` to reject accessor/Proxy fields. Production `fetch().json()` returns plain data so this isn't a hot path, but the renderer claims byte-for-byte parity that doesn't hold. Suggestion: mirror the descriptor check, or extract a browser-safe codec from `@wuphf/protocol`.

**Types Finding 3 — `droppedKeys` can be forged by subprocess payloads (`logger.ts:47`, `broker-internal.ts:116-127`, `broker.ts:747-749`).**
`droppedKeys` is in the safe-key allowlist. A subprocess can post `{ broker_log: "info", event: "X", payload: { droppedKeys: 999 } }` with no unsafe keys; `filterPayloadToSafeKeys` keeps it; `forwardBrokerLog` only appends a computed counter when `droppedKeyCount > 0`. The payload-derived value passes through, forging the redaction-counter's semantics. Suggestion: drop `droppedKeys` from `safePayload` before constructing `finalPayload` (treat it as supervisor-reserved).

### LOWs

**Architecture Finding 6 — coverage tests re-introduce `Reflect`-private access (`broker-supervisor.spec.ts:1788/1812/1836`).** Pass-3 removed synthetic `Reflect.set` tests; these `Reflect.get`/`apply` reads of private methods to cover pid-null branches re-introduce the same coupling.

**Logger rotation v8-ignore (`logger.ts:248-258`)** — 3 lenses flagged, severity divergence: architecture=MEDIUM, electron=LOW, sre=LOW. The comment names the exact reachable case (external delete races rotation). Recommend removing the ignore + injecting a filesystem helper to cover.

### Direct disagreements

- `logger.ts:254`: architecture=MEDIUM vs electron=LOW (and sre=LOW). Severity-only divergence on whether the v8-ignore is a real problem; all three agree the branch is reachable.
- `protocol/ipc.ts:229`: architecture=MEDIUM vs types=LOW. Both agree there's parity drift; disagreement is over real-world reachability.

### Security lens — re-run needed

Status 124 (timeout, 300s/agent). No report produced. The other lenses incidentally covered the likely top security concerns (encoded path traversal in brand, accessor-descriptor bypass in bootstrap, payload-forgery of `droppedKeys`), but we're missing the independent security frame. Recommend re-dispatching the security lens alone with the same problem statement.

### Aggregate verdict

**Hold for fixes.** All 4 returning lenses returned "Hold for fixes". The blocking finding is the convergent watchdog OR-collapse; the rest are MEDIUM-or-below and can roll forward incrementally. Suggested fix order:

1. **Revert the pass-3.1 collapse on `broker.ts:660-697`**; add a queued-timer race test that holds the watchdog timer past `{ready}` / `stop()`.
2. Add `utilityProcess.on("error", ...)` diagnostic listener.
3. Re-examine `startupForceTimer` v8-ignore; either clear in `handleExit` or cover.
4. Treat `droppedKeys` as supervisor-reserved.
5. Tighten `BrokerUrl` brand against encoded path-segment input; add `%2e%2e` tests.
6. Re-dispatch security lens.

Per-lens reports at `/tmp/triangulation-pr792-pass4/{architecture,types,security,sre,electron}-report.md`.

FranDias:
## Pass-4 fix — restored startup watchdog OR-arms (`8ac105d4`)

Addressed the convergent HIGH from the pass-4 triangulation. Pass-3.1's collapse to a single sender-identity arm regressed the queued-callback defense the restart timer at `broker.ts:575` explicitly relies on. Restored the 3 race-tolerant arms with comments tying back to each racing handler.

### Production change (`broker.ts`)

```ts
if (
  this.brokerProcess !== brokerProcess ||
  this.brokerUrl !== null ||
  this.stopping ||
  /* v8 ignore next */ this.fatalReason !== null
) {
  return;
}
```

- **`brokerProcess !== brokerProcess`** (sender-identity): fresh fork has cycled.
- **`brokerUrl !== null`**: `{ready}` won the race; without this arm, queued callback kills a ready broker.
- **`stopping`**: `stop()` won the race; without this arm, queued callback double-kills a process already being torn down.
- **`fatalReason !== null`**: defensive only. No current architecture path sets `fatalReason` while a same-handle startup timer is still armed (cap-paths either run from inside this callback or null `brokerProcess` first via `handleExit`). True-arm v8-ignored as belt-and-suspenders, mirroring the restart timer's `fatalReason` defense at line 591.

### Test changes (`broker-supervisor.spec.ts`)

Added 2 tests that specifically exercise the `brokerUrl` and `stopping` arms via the same `vi.spyOn(globalThis, "clearTimeout").mockImplementation(...)` pattern the restart-timer race test uses at line 1078:

- `startup-timeout watchdog bails when {ready} won the race (brokerUrl arm, queued callback)` — sends ready, no-ops clearTimeout so the queued startup callback still fires; asserts no `broker_ready_timeout` log and no kill.
- `startup-timeout watchdog bails when stop() won the race (stopping arm, queued callback)` — calls `stop()`, no-ops clearTimeout, advances the timer queue, then settles stop via exit; asserts no `broker_ready_timeout` log (kill is expected from stop itself).

These address the staff-reviewer's pass-3 complaint that the prior test was passing for the wrong reason — each new test sets up state where the sender-identity arm is FALSE and a later arm must catch the race.

### Verification

- `bunx tsc --noEmit` ✓
- `bunx biome check` ✓ on changed files
- `bunx vitest run` → **192 passed** (was 190 + 2 new)
- `bunx vitest run --coverage` → **100/100/100/100** (352 branches, +4 vs pass-3.1)
- Pre-push hooks: complexity-baseline, desktop-check-invariants, desktop-check-ipc-allowlist, desktop-test, desktop-typecheck, file-size — all green

### Triangulation findings status

| # | Finding | Severity | Status |
|---|---|---|---|
| 1 | Watchdog OR-collapse race (electron+sre+arch) | HIGH | **FIXED** in `8ac105d4` |
| 2 | `utilityProcess.on("error", ...)` missing (sre) | HIGH | DEFERRED — open follow-up: subscribe to V8-fatal `"error"` event with `type`/`location`/`report` diagnostic shape, let `"exit"` continue to drive restart |
| 3 | `startupForceTimer` v8-ignore reachable (arch+electron) | HIGH | DEFERRED — open follow-up: clear `startupForceTimer` in `handleExit` for the exited handle, then remove v8-ignore |
| 4 | `BrokerUrl` admits `%2e%2e` (types) | MEDIUM | DEFERRED — protocol-package change; out of this PR's scope |
| 5 | `forkProcess` reentrancy "infeasible" comment wrong (arch) | MEDIUM | DEFERRED |
| 6 | renderer parseBootstrap missing descriptor guard (arch+types) | MEDIUM | DEFERRED — see #2 / fold into protocol parity sweep |
| 7 | `droppedKeys` forgeable by subprocess (types) | MEDIUM | DEFERRED — supervisor-reserved-key refactor |
| 8 | Coverage tests re-introduce private reflection (arch) | LOW | DEFERRED — pid-null paths |
| 9 | logger rotation v8-ignore reachable (arch+electron+sre) | LOW–MEDIUM | DEFERRED — inject filesystem helper |
| 10 | Security lens timeout | gap | DEFERRED — re-dispatch lens alone |

This PR ships the BLOCKING fix from the triangulation. The remaining 9 are all DEFERRABLE per the pass-4 verdict ("hold for fixes" was driven by finding #1). Suggest landing this and filing follow-up issues for #2-#10.

FranDias:
## Pass-5 triangulation synthesis (5 lenses: architecture, types, security, perf, distsys)

Read-only codex agents in parallel on the flattened PR (`41ace602`). Reports + raw outputs at `/tmp/triangulation-pr792-pass5/`. All 5 lenses returned (status 0); no timeouts this round.

### HIGH (1 lens, real bug)

**distsys Finding 1 — `probeInFlight` latch never resets** (`apps/desktop/src/renderer/main.ts:93-117`, `:286-293`)

The bootstrap probe sets `probeInFlight = true` on entry but never clears it. After the first probe completes (success OR failure), every later `refreshBrokerStatus()` sees `!probeInFlight === false` and skips bootstrap. This breaks the cross-process recovery path the surrounding comment describes — after a broker restart or initial transient `/api-token` failure, the renderer never re-probes. Single-flight has become one-shot. **Fix**: wrap the probe body in `try { ... } finally { probeInFlight = false; }`. Add a focused test for `url1 → null/error → url2`.

This is a regression from PR #791's bootstrap probe simplification (the prior code had different ordering). Pass-3/4 didn't catch it because all the prior tests only exercised a single probe.

### MEDIUM (4 distinct, 1 cluster)

**types Finding 1 — `BrokerUrl` brand admits root-path form, renderer string-concatenates** (`packages/protocol/src/ipc.ts:287/295`, `apps/desktop/src/renderer/main.ts:104`)

The brand invariant says "BrokerUrl IS the broker origin," but `assertApiBootstrapBrokerUrl` now accepts both `parsed.origin` AND `${parsed.origin}/`. The renderer then does:
```ts
fetch(`${bootstrap.brokerUrl}/api/health`, ...)
```
For a brand-accepted `http://127.0.0.1:54321/`, this builds `http://127.0.0.1:54321//api/health` (double slash). The brand failed its documented invariant. **Fix**: either (a) reject the trailing-slash form (mirror BrowserWindow's `loadURL` to add slash at the load site) or (b) replace string concat with `new URL("/api/health", bootstrap.brokerUrl)` everywhere. (a) preserves the brand semantics more cleanly.

**security Finding 1 — V8 diagnostic `report` allowlisted, can leak** (`broker.ts:505`, `logger.ts:70`, `logger.ts:231`)

Pass-4 added `type`/`location`/`report` to `SAFE_PAYLOAD_KEYS` to surface V8-fatal diagnostics. Electron's `utilityProcess.error.report` field carries the full Node diagnostic report — local paths, cwd, command line, network interfaces, loaded libraries, environment variables. Even with the broker's env allowlist, on V8 fatal we'd write user/system details to structured logs. **Fix**: don't allowlist raw `report`. Log `type`, `location`, `pid`, `restartCount`, plus a `reportBytes`/hash; OR parse the report and emit an explicit redacted summary.

**distsys Finding 2 — startup force-stop hangs on partial kill failure** (`broker.ts:752-760`, `broker.ts:831-843`)

The watchdog escalates to `forceStop(brokerProcess)` but doesn't settle/mark fatal/schedule fallback if no `exit` event follows. Windows path: `forceStop` starts async `taskkill /F` and returns immediately; if taskkill rejects, the code only logs via `.catch()` and leaves the timed-out process as current. With `startupTimedOutHandles` fencing ready, no restart, and no fatal — `whenReady()` waiters hang indefinitely. **Fix**: treat force-stop failure/no-exit as bounded terminal state — fallback to `utilityProcess.kill()` on taskkill rejection + arm a final exit grace, or mark fatal/reject waiters after force escalation fails to produce exit.

**Convergent MEDIUM (arch+security, severity split) — renderer accepts arrays** (`renderer/main.ts:188`, `protocol/src/ipc.ts:240`, `protocol/src/receipt-utils.ts:6`)

`parseBootstrap` accepts any non-null object (`typeof value === "object" && value !== null`) which includes arrays. Protocol's `apiBootstrapFromJson` uses `requireRecord` → `isRecord` which explicitly rejects arrays. The "byte-for-byte parity" claim is false:

```js
parseBootstrap(Object.assign([], { token: "A".repeat(16), broker_url: "http://127.0.0.1:54321" }))
// renderer: passes
// protocol: throws
```

Unreachable from real `fetch().json()` (JSON arrays can't carry named properties), but parity is the contract. **Fix**: add `Array.isArray(value)` rejection in `parseBootstrap`; add a parity test for `Object.assign([], { token, broker_url })`.

Severity disagreement: architecture=MEDIUM (drift risk for future contributors), security=LOW (no live exploit path via `fetch().json()`). Both findings agree on substance.

**perf Finding 1 — broker log forwarding stalls main thread under storm** (`broker.ts:783-797`, `logger.ts:132-148`)

Every accepted `{ broker_log }` IPC message → synchronous `mkdirSync` + `statSync` + `appendFileSync` on the Electron main thread. Normal traffic is fine; at 100× volume (e.g., a request flood hitting per-request rejection logs in the broker) this stalls UI. **Fix**: rate-limit/coalesce forwarded broker logs, OR amortize the file sink (cache directory existence + in-memory size counter + batch/async append).

### LOW

- **distsys Finding 3**: pre-start `whenReady()` waiters leak if initial fork throws — `start()` catch path doesn't reject them. (`broker.ts:335-367`, `:190-198`)
- **architecture Finding 2**: `StructuredLoggerConfig.fs` is misleadingly named — only `rotateLogs` uses it; writes still go through raw fs. Rename to `rotationFs` before merge.
- **architecture Finding 3**: `LOG_NAME_PATTERN` is duplicated in `broker-internal.ts:97-104` vs `logger.ts:45/238-240`. Future drift waiting to happen — export `isValidLogName` from logger.
- **architecture Finding 4**: `vitest.config.ts:32-34` comment claims one v8-ignore remains, but the actual code has ignored arms at `broker.ts:475/617/620/720/734` + `window.ts:130-134`. Stale inventory.
- **perf Finding 2**: `droppedKeys` reservation does 2 extra object clones per log forward. Mutate-in-place instead.

### Direct disagreements (severity-only)

| File:line | Disagreement |
|---|---|
| `renderer/main.ts:188` (array rejection) | architecture=MEDIUM, security=LOW |
| `protocol/ipc.ts:287/295` (trailing-slash) | types=MEDIUM, security=LOW |

Substance agrees in both cases.

### Aggregate verdict

**Hold for the HIGH and at least the MEDIUMs that affect availability/security** (probeInFlight latch, V8-report leak, force-stop hang, BrokerUrl double-slash). The arch/perf MEDIUMs and all LOWs can roll forward as follow-ups.

### Suggested fix order

1. **`probeInFlight` reset** — single-line fix in renderer, regression-test the recovery path.
2. **Remove `report` from `SAFE_PAYLOAD_KEYS`** — log `reportBytes`/hash instead, or extract specific subfields.
3. **Force-stop terminal state** — Windows taskkill rejection fallback + bounded final grace.
4. **`BrokerUrl` trailing-slash semantics** — pick (a) brand rejects trailing slash + load-site adds it, or (b) caller uses `new URL`.
5. **`parseBootstrap` `Array.isArray` rejection** — one line + one test.
6. **Pre-start waiters leak** — reject in `start()`'s initial fork catch path.
7. **Logger field rename** (`fs → rotationFs`).
8. **Export `isValidLogName` from logger** (de-dup).
9. **Coverage config v8-ignore inventory** — update comment.
10. **Broker log forwarding perf** — defer to follow-up issue if PR doesn't already need a respin.

Per-lens reports at `/tmp/triangulation-pr792-pass5/{architecture,types,security,perf,distsys}-report.md`.

FranDias:
## Pass-5 fix #1 — BrokerUrl is bare canonical origin (`f58cb7b3`)

Addresses pass-5's MEDIUM types finding (`packages/protocol/src/ipc.ts:287/295`, `apps/desktop/src/renderer/main.ts:104`). Tightened the brand to a single canonical wire form after tracing every producer and consumer.

### Why the back-and-forth resolves with tightening

| Site | Behavior | Form |
|---|---|---|
| Producer (`packages/broker/src/listener.ts:269`) | `asBrokerUrl(\`http://\${normalizeLoopback}:\${port}\`)` | **Bare origin only** |
| Consumer 1 (`window.ts:110`) | `parseUrl(brokerUrl, ...).toString()` | `URL.toString()` adds `/` AT the load site (where `will-navigate` needs it) |
| Consumer 2 (`renderer/main.ts:104`) | `fetch(\`${bootstrap.brokerUrl}/api/health\`)` | String concat — **only safe with bare input** |

In real production, the broker emits bare and every consumer is happy. The brand previously accepted both bare and trailing-slash forms — that was the drift the doc didn't match. The tightening makes the brand agree with the broker's actual emission, with **zero production compatibility cost**.

### Diff

- `packages/protocol/src/ipc.ts`: drop `brokerUrl === ${parsed.origin}/` OR-arm and the redundant `(parsed.pathname !== "" && parsed.pathname !== "/")` legacy gate. Single check: `brokerUrl !== parsed.origin`.
- `apps/desktop/src/renderer/main.ts`: mirror the same single-check rule.
- `packages/protocol/tests/ipc.spec.ts`: flip `isBrokerUrl("http://127.0.0.1:54321/")` from `true → false`; add explicit trailing-slash throws to the rejection assertion set + the `it.each` negative-case table.
- `apps/desktop/tests/main-window-security.spec.ts`: add `http://127.0.0.1:54321/` to the loopback-with-extras rejection list.

### Why not the alternative (`new URL("/api/health", base)` everywhere)?

That would keep accept-both and require all consumers to use the URL constructor. Worse because:
1. Two valid wire forms = ambiguous contract; contributors have to know not to string-concat.
2. Brand doc says origin — should match.
3. Much larger change surface, more drift risk, vs. one validator tightening.

### Verification

- `packages/protocol`: 541 tests pass (+1 vs prior — new trailing-slash negative case), typecheck + biome clean, demo 70/70 pass
- `apps/desktop`: 200 tests pass, 100/100/100/100 coverage maintained, typecheck + biome clean
- Pre-push hooks: complexity-baseline, desktop-check-invariants, desktop-check-ipc-allowlist, desktop-test, desktop-typecheck, file-size, protocol-demo, protocol-invariants — all green

### Status of remaining pass-5 findings

Still open in this PR:

- HIGH: `probeInFlight` latch never resets in renderer
- MEDIUM: V8 `report` allowlisted (potential leak)
- MEDIUM: Windows force-stop hangs on taskkill rejection
- MEDIUM: `parseBootstrap` accepts arrays (parity gap)
- LOWs: pre-start waiter leak, `fs`-field misname, `LOG_NAME_PATTERN` dup, stale v8-ignore comment, log-forward clone churn

FranDias:
## Pass-5 follow-ups landed — 9 findings addressed across 4 commits

Three parallel codex agents in isolated worktrees, then cherry-picked onto the PR branch in dependency order (renderer → logger → broker). One auto-merge on `logger.ts` resolved cleanly without conflict. PR head: `f5641b3a`.

### Per-finding disposition

| # | Finding | Severity | Status | Commit |
|---|---|---|---|---|
| probeInFlight latch never resets | distsys | **HIGH** | FIXED | `c9a911f4` — `finally { probeInFlight = false }` + regression test for `url1 → null → url2` proving second probe runs |
| V8 `report` allowlisted, leaks paths/env | security | MEDIUM | FIXED | `f5641b3a` — removed `report` from `SAFE_PAYLOAD_KEYS`, swapped to `reportBytes` derivation in broker.ts |
| Windows force-stop hangs on partial kill | distsys | MEDIUM | FIXED option (b) | `f5641b3a` — added final fatal deadline; if no exit by then, reject waiters + transition to dead |
| `parseBootstrap` accepts arrays | arch+security | MEDIUM | FIXED | `8a3209d2` — `Array.isArray(value)` guard + parity test |
| broker log forwarding stalls main thread | perf | MEDIUM | FIXED (amortization only) | `b1fa79ed` — mkdir-once-per-instance + in-memory `currentFileBytes` cache; eliminates `mkdirSync`+`statSync` from hot path |
| pre-start `whenReady()` waiters leak on initial fork throw | distsys | LOW | FIXED | `f5641b3a` — `rejectReadyWaiters(error)` in `start()`'s initial catch path |
| `droppedKeys` reservation does 2 extra clones | perf | LOW | FIXED | `f5641b3a` — `delete safePayload.droppedKeys` in place; eliminates `safePayloadWithoutReservedCounter` clone |
| `fs` config field misleadingly named | architecture | LOW | FIXED option (a) | `b1fa79ed` — renamed `StructuredLoggerConfig.fs` → `rotationFs` |
| `LOG_NAME_PATTERN` duplicated in `broker-internal.ts` | architecture | LOW | FIXED | `b1fa79ed` — exported from `logger.ts`, imported in `broker-internal.ts` |
| vitest config v8-ignore inventory stale | architecture | LOW | FIXED option (a) | `b1fa79ed` — enumerated intentional ignore classes accurately |

### Verification on integration branch

- `bunx tsc --noEmit` ✓ (desktop + protocol)
- `bunx biome check` ✓ on all changed files
- `bunx vitest run` → **208 tests passed** (+8 vs `f58cb7b3`'s 200)
- `bunx vitest run --coverage` → **100/100/100/100** (378 branches, +16 net)
- Pre-push hooks all green: complexity-baseline, desktop-check-invariants, desktop-check-ipc-allowlist, desktop-test, desktop-typecheck, file-size, protocol-demo, protocol-invariants

### Process

Followed CLAUDE.md "Worktree-based parallelism":
1. 3 parallel codex dispatches in isolated worktrees (renderer, broker cluster, logger hygiene).
2. Scope boundaries pre-declared so file overlap surfaced as predictable auto-merges (broker batch touched only `SAFE_PAYLOAD_KEYS` Set in `logger.ts`; logger batch touched config interface + rotate/append + `LOG_NAME_PATTERN` export; disjoint regions).
3. Cherry-pick onto integration branch in dependency order (renderer first as isolated, broker last as it auto-merged with logger's `logger.ts` changes).
4. Single auto-merge resolved cleanly; no manual conflict resolution needed.

### Pass-5 status: all findings resolved

The HIGH (`probeInFlight`) + 4 MEDIUMs + 5 LOWs are all FIXED. The pass-5 perf finding's "batch/async append" suggestion was descoped from this round (the amortization shipped covers the main-thread-stall part of the finding; full async I/O is a separate redesign and can ship as a follow-up issue if needed).

CodeRabbit's 3 inline comments from the earlier review pass are all auto-marked addressed via prior commits (`bb2ea05` and `8ac105d`).

FranDias:
## Pass-6 triangulation synthesis (5 lenses: api, sre, electron, perf, types)

Read-only codex agents in parallel on the now-16-commit PR (`77b77bfe`). Reports + raw outputs at `/tmp/triangulation-pr792-pass6/`. All 5 lenses returned (status 0); lens picks rotated for fresh signal (api/sre/electron not run since pass-3/4, perf re-runs after the amortization landed).

### Convergent MEDIUM (api + sre agree)

**Malformed `{ ready, brokerUrl }` becomes an opaque startup timeout** (`broker-internal.ts:40-53`, `broker.ts:246-279`)

After pass-5's BrokerUrl tightening (`f58cb7b3`: reject trailing-slash + non-canonical forms), a future broker emitting a non-canonical URL now gets silently dropped at `readReadyMessage` returning `null`. The message handler falls through and waits for the startup watchdog. On-call only sees `broker_ready_timeout` — the actual cause is "broker sent ready, wire shape failed validation."

This is the wire-drift case the brand tightening was supposed to protect against; the desktop side just doesn't surface the diagnostic. The protocol drift you'd want to debug is invisible until the watchdog ceiling fires.

**Both lenses suggested the same fix**: when a message has ready-shape keys (`ready: true` + `brokerUrl` present) but `isBrokerUrl(brokerUrl) === false`, log a structured `broker_ready_invalid` event with a sanitized reason (never the raw URL — strip to scheme/host shape, not content). Continue dropping the message; let the watchdog handle the lifecycle.

### Self-introduced regression (sre MEDIUM)

**Deleting the log directory permanently disables file logs** (`logger.ts:141-144`, `:155-158`)

This is a regression from pass-5b's amortization (`b1fa79ed`). After first successful write, `initializedLogDirectory` is set; subsequent appends skip `mkdirSync`. If the directory is deleted later (user cleanup, profile reset, AV, packaged-app dock-icon "Clear Logs" UI), every subsequent append throws `ENOENT` and the swallowed `try/catch` makes the file logs silently disappear forever — recovery requires app restart. Console logging still works, masking the failure.

**Fix**: on `ENOENT` in `appendFileSync`, clear `initializedLogDirectory`, retry `mkdirSync` + `appendFileSync` once. Add a regression test: write once, `rmSync` the log directory, verify the next log recreates the dir.

### Other MEDIUMs (1 lens each)

**api Finding 2 — `BrokerUrl` contract isn't documented for cross-language implementers** (`packages/protocol/docs/modules/ipc.md:19/125`)

The actual rule (`raw === parsed.origin`) is encoded in tests + the validator, but the protocol module docs still describe the loose v0 shape. A Go/Rust implementer reading the docs would emit `${origin}/` and get silently dropped by the desktop (see convergent finding above). Suggestion: add an accepted/rejected matrix to `ipc.md` + cross-language fixture guidance.

**perf Finding 1 — `Buffer.byteLength` scans full payload before truncation** (`logger.ts:253/257`, `broker.ts:931`)

`normalizePayloadValue` and `diagnosticReportBytes` both compute `Buffer.byteLength(string, "utf8")` over the whole input string before applying truncation. On the broker-log forwarding hot path (`payload.error` or `.reason` fields), 100x larger inputs become O(N) main-thread CPU work. The 8 KiB output cap doesn't help because the scan happens BEFORE truncation.

**Fix**: cap input size BEFORE counting — e.g., `if (value.length > maxChars * 2) return TRUNCATION_SENTINEL`. For `diagnosticReportBytes`, log `reportBytesAtLeast: <bound>` + `reportTruncated: true` when input exceeds a conservative ceiling.

This finding is severity-split with electron's: electron flagged the same `broker.ts:931` site at LOW (rare failure-path code), perf flagged it at MEDIUM (logger.ts hot-path code). The logger.ts hit is the bigger concern.

### LOWs (defer)

- **api Finding 3** — `ApiBootstrapWire` v0 shape is closed (rejects unknown keys); future additive changes need an explicit upgrade path. Document or accept.
- **electron Finding 1** — `broker-spawn.md:107-120` env allowlist docs omit `WUPHF_RENDERER_DIST` + `WUPHF_DEV_RENDERER_ORIGIN`. Docs-only fix.
- **electron Finding 2** — `broker.ts:931` `Buffer.byteLength` (severity-split with perf above).
- **perf Finding 2** — `Object.entries` allocations in `forwardBrokerLog` (`broker.ts:821`, `broker-internal.ts:110`, `logger.ts:224`). `for...in` + `Object.hasOwn` avoids the entries-array allocation per log.
- **perf Finding 3** — pass-5b amortization comment claims bounded drift, but another local writer mutating the file invalidates the claim. Either narrow the comment (single-writer assumption) or re-stat near threshold.
- **types Finding 1** — broker-internal.ts casts message payloads to `Readonly<Record<string, unknown>>` and `Object.entries` enumerates accessor-backed records. Production `utilityProcess` structured clone makes this LOW risk, but the helper is exported and the renderer-side `parseBootstrap` already moved to descriptor-based reads for this exact shape class. Mirror the descriptor pattern in `filterPayloadToSafeKeys`.

### Direct disagreements (severity-only)

| File:line | Disagreement |
|---|---|
| `broker.ts:931` (`diagnosticReportBytes` full-string scan) | electron=LOW, perf=MEDIUM |

Substance agrees.

### Aggregate verdict

- **api**: Hold for fixes (convergent ready-invalid logging + docs)
- **sre**: Hold for fixes (convergent ready-invalid + log-dir delete regression)
- **electron**: Ship with follow-ups
- **perf**: Ship with follow-ups
- **types**: Ship with follow-ups

The two Hold-for-fixes lenses converge on the same finding (#1 ready-invalid logging) + the sre lens flags a real regression I introduced. Those two are blocking. Suggested fix order:

1. **Log `broker_ready_invalid` when ready-shape message fails validation** (`broker-internal.ts` + `broker.ts`)
2. **ENOENT recovery in logger append path** (`logger.ts`) — regression fix
3. **Cap input size before `Buffer.byteLength`** (`logger.ts:253/257`, `broker.ts:931`)
4. **Protocol docs: accepted/rejected matrix for BrokerUrl** (`ipc.md`)
5. **`for...in` + `Object.hasOwn` in `filterPayloadToSafeKeys`** + descriptor-style read
6. **Env allowlist docs** (`broker-spawn.md`)
7. **Pass-5b amortization comment** narrow to single-writer

Per-lens reports at `/tmp/triangulation-pr792-pass6/{api,sre,electron,perf,types}-report.md`.

FranDias:
## Pass-6 follow-ups landed — 10 findings addressed across 7 commits

Three parallel codex agents in isolated worktrees (broker, logger, docs), then cherry-picked onto the PR branch in dependency order (docs → logger → broker). Pre-declared scope boundaries → zero conflicts. PR head: `00cd10f7`.

### Per-finding disposition

| Finding | Severity | Status | Commit |
|---|---|---|---|
| Convergent: malformed `{ ready, brokerUrl }` becomes opaque startup timeout | MEDIUM (api+sre) | FIXED | `00cd10f7` — `readReadyMessage` returns discriminated `{ kind: "ok"\|"invalid", reason }`; handler logs `broker_ready_invalid` with shape-only reason (never raw URL) |
| Self-introduced regression: log dir delete permanently disables file logs | MEDIUM (sre) | FIXED | `560c7a47` — ENOENT recovery retries `mkdirSync` + `appendFileSync` once before swallowing |
| `Buffer.byteLength` scans full payload before truncation (logger normalize) | MEDIUM (perf) | FIXED | `e10f71dd` — pre-cap input by char-length before UTF-8 byte walk |
| `diagnosticReportBytes` scans full V8 report unbounded | MEDIUM (electron LOW / perf MED) | FIXED | `00cd10f7` — capped at 64 KiB before `Buffer.byteLength` |
| `BrokerUrl` contract not documented for cross-language implementers | MEDIUM (api) | FIXED | `735d07db` — accepted/rejected matrix added to `packages/protocol/docs/modules/ipc.md` |
| `Object.entries` allocations on broker-log forward + logger validation | LOW (perf) | FIXED | `00cd10f7` + `8acf391b` — `for...in` + `Object.hasOwn` |
| `broker-internal.ts` casts trust accessor-backed records | LOW (types) | FIXED | `00cd10f7` — descriptor-based reads in `filterPayloadToSafeKeys` (mirrors `parseBootstrap` pattern) |
| `ApiBootstrapWire` closed-shape upgrade path undocumented | LOW (api) | FIXED | `735d07db` — explicit "Upgrade path" subsection in ipc.md |
| Broker spawn env allowlist docs stale (missing `WUPHF_RENDERER_DIST`, `WUPHF_DEV_RENDERER_ORIGIN`) | LOW (electron) | FIXED | `148a03b8` — updated `apps/desktop/docs/modules/broker-spawn.md` |
| Amortization comment claims unbounded drift bounded | LOW (perf) | FIXED option (a) | `b73a1792` — narrowed to single-writer assumption |

### Verification on integration branch

- `bunx tsc --noEmit` ✓ (desktop + protocol)
- `bunx biome check` ✓ on all changed files
- `bunx vitest run` → **220 desktop tests passed** (+12 vs prior 208)
- `bunx vitest run --coverage` → **100/100/100/100** (419 branches, +41 net)
- `bash scripts/test-protocol.sh` → **541 protocol tests passed**
- `bun run packages/protocol/scripts/demo.ts` → 70/70 scenarios pass
- Pre-push hooks all green

### Process

Followed CLAUDE.md "Worktree-based parallelism":
1. 3 parallel codex dispatches in isolated worktrees with pre-declared disjoint file ownership.
2. Cherry-pick onto integration branch in dependency order: docs (isolated) → logger (logger.ts only) → broker (broker.ts + broker-internal.ts + broker tests, biggest scope last to absorb hypothetical overlap).
3. Zero conflicts at integration thanks to clean scope partitioning.

The broker agent's commit was blocked by sandbox ref-lock write permissions (a known limitation when codex runs in sandbox-write mode); committed from outside the sandbox after the agent's hooks verified the staged tree.

### Pass-6 fully resolved

The HIGH + 4 MEDIUMs + 5 LOWs are all FIXED. PR has gone through 7 review rounds total (pass-1 through pass-6, plus a CR fix) and is now in a hardened state with:

- Brand-tight wire shape (BrokerUrl bare canonical origin only, surfaced as `broker_ready_invalid` on drift)
- Diagnostic data bounded (V8 reports capped at 64 KiB, payload strings capped before byte scan)
- Failure paths terminal (force-stop deadline, ENOENT log-dir recovery, pre-start waiter rejection)
- Performance amortized (mkdir-once, in-memory size cache, `for...in` over `Object.entries`, descriptor reads)
- Test surface real (no synthetic private-method reflection; queued-callback races covered)
- Docs aligned with code (BrokerUrl matrix, env allowlist, upgrade path)

CodeRabbit's 3 inline comments + 1 outside-diff comment are all addressed in PR history.

FranDias:
## Pass-7 triangulation synthesis (5 lenses: architecture, distsys, security, api, sre)

Read-only codex agents on the 23-commit PR (`00cd10f7`). Reports + raw outputs at `/tmp/triangulation-pr792-pass7/`. All 5 lenses returned (status 0); lens picks rotated for fresh signal against pass-6 (which used api/sre/electron/perf/types).

### Convergent MEDIUM (5-lens agreement) — byte caps are actually UTF-16 code-unit caps

Five lenses independently flagged the same root cause: `MAX_LOG_STRING_BYTES` and `MAX_DIAGNOSTIC_REPORT_BYTES` are enforced via `value.length` (UTF-16 code units), not encoded byte count.

**`apps/desktop/src/main/logger.ts:273-281`** (`normalizePayloadValue`) — MEDIUM by 4-of-5 lenses (api LOW):

```ts
if (value.length > MAX_LOG_STRING_BYTES) {
  return value.slice(0, MAX_LOG_STRING_BYTES) + "...";
}
// Falls through with no byte-length check otherwise.
```

A broker-controlled safe-key payload string `"€".repeat(8192)` has `.length === 8192` (within the cap) but encodes to ~24 KiB UTF-8. The early-return path doesn't fire; the function returns the full 24 KiB string. The reviewed log-size budget is bypassed and the main-thread sync write is 3x larger than the advertised cap. The test at `logger.spec.ts:384-397` codifies this behavior — locking the bug in.

**`apps/desktop/src/main/broker.ts:936-939`** (`diagnosticReportBytes`) — LOW by 4-of-5 lenses (architecture MED):

```ts
if (report.length > MAX_DIAGNOSTIC_REPORT_BYTES) {
  return MAX_DIAGNOSTIC_REPORT_BYTES;
}
return Buffer.byteLength(report, "utf8");
```

A 65,536-code-unit `€` string passes the length gate and `Buffer.byteLength` returns 196,608 — 3x the advertised "64 KiB" cap. Doesn't leak the raw report, but the `reportBytes` log field violates the cap and can mislead dashboard alerts.

**Fix**: byte-aware truncation in `normalizePayloadValue` (encode → trim to byte budget → decode-safe boundary). For `diagnosticReportBytes`, clamp with `Math.min(Buffer.byteLength(slice, "utf8"), MAX_DIAGNOSTIC_REPORT_BYTES)` after slicing a bounded prefix.

### Convergent (arch + api, MEDIUM in api lens) — BrokerUrl validation hand-mirrored in 3 places

**`packages/protocol/src/ipc.ts:274`** owns the invariant. **`apps/desktop/src/main/broker-internal.ts:71` + `:163`** has the classify-reason helper (added by pass-6). **`apps/desktop/src/renderer/main.ts:155`** has another full mirror.

A future BrokerUrl tweak must update all 3 sites in lockstep. Pass-7 finds them aligned today, but the drift hazard is real for a stable wire shape. The protocol package can't be pulled into the renderer bundle (depends on `node:crypto`), so the right fix is shared conformance vectors in `packages/protocol/testdata/broker-url-vectors.json` consumed by all 3 site test suites.

### Single-lens MEDIUMs (real)

**sre Finding 1 — broker payload can forge `eventLsn`** (`logger.ts:56/118-125`, `broker.ts:826-833`)

`eventLsn` is in `SAFE_PAYLOAD_KEYS` and the broker forwards subprocess payloads through `forwardBrokerLog`. A `{ broker_log }` with `{ eventLsn: 999999 }` overwrites the structured logger's sequence number. At 3am, ordering/correlation by `eventLsn` becomes untrustworthy. **Fix**: remove `eventLsn` from the safe-key allowlist (it's logger-envelope-only, never a legitimate subprocess payload field) OR spread payload BEFORE logger-owned fields so the envelope wins.

**security Finding 1 — broker log event names are unbounded** (`broker-internal.ts:115`, `broker.ts:833`)

`sanitizeBrokerEventName` checks `LOG_NAME_PATTERN` (regex shape) but no max length. A hostile broker can send `event: "a".repeat(50_000_000)` — passes the regex, then the main process synchronously writes a 50 MB JSON log record. Real DoS vector. **Fix**: add a small length cap (256? 512?) checked before regex.

**distsys Finding 2 — invalid ready URL classifier reparses unbounded input** (`broker-internal.ts:71/166-181`)

`readReadyMessage` calls `isBrokerUrl(value)` (parses via `new URL`), then on `kind: "invalid"` the classifier `classifyBrokerUrlReason` parses `new URL(value)` AGAIN to derive the sanitized reason. The input is unbounded subprocess data — two parses of arbitrary-size input → avoidable main-process stall during startup failure recovery. **Fix**: cap input length before parsing in the classifier; return `url_too_long` sanitized reason if over the cap.

### LOWs

- **api Finding 1** — bootstrap codec isn't single-source-of-truth (renderer hand-mirrors). Same root as the arch hand-mirror finding above; suggested fix: conformance fixtures.
- **api Finding 2** — BrokerUrl spec is JS-WHATWG-specific; non-TS implementers need portable vectors. Same fix.
- **api Finding 4 / arch Finding 3** — `packages/protocol/docs/OVERVIEW.md:266-267` still says "accepts any string URL"; `ipc.md:11-13` public-API list omits `BrokerUrl` + has stale line numbers. Pure docs.
- **architecture Finding 4** — `renderer/main.ts` exports `parseBootstrap` + `refreshBrokerStatus` from a side-effectful module entrypoint. Tests inherit DOM/window init. Refactor to a side-effect-free helper module.

### Direct disagreements (severity-only, substance agrees)

| Site | Disagreement |
|---|---|
| `broker.ts:936` (diagnosticReportBytes cap) | api=LOW, architecture=MEDIUM, distsys=LOW, security=LOW, sre=LOW → leaning LOW |
| `logger.ts:273` (normalize cap) | api=LOW, architecture=MEDIUM, distsys=MEDIUM, security=MEDIUM, sre=MEDIUM → MEDIUM |
| `broker-internal.ts:71` (hand-mirrored validation) | architecture=LOW, distsys=MEDIUM |
| `renderer/main.ts:155/190` (parseBootstrap mirror) | api=MEDIUM, architecture=LOW |

### Aggregate verdict

- **distsys**: Hold for fixes (byte-cap + invalid-URL reparse + diagnostic-cap)
- **security**: Hold for fixes (byte-cap + event-name DoS + diagnostic-cap)
- **sre**: Hold for fixes (byte-cap + eventLsn forgery + diagnostic-cap)
- **architecture**: Ship with follow-ups
- **api**: Ship with follow-ups

3-of-5 lenses Hold-for-fixes. The blocking items converge on:

1. **Byte-cap bug** (`logger.ts:273`) — user-visible output exceeds advertised cap; ALL 5 lenses flagged.
2. **eventLsn forgery** — broker subprocess can overwrite logger sequence ordering.
3. **Event-name DoS** — broker can write multi-MB log records via inflated event names.

Suggested fix order:

1. **Byte-aware truncation** in `normalizePayloadValue` + clamp in `diagnosticReportBytes`. Same pattern in both sites. Update existing tests to enforce byte boundaries (the current test at `logger.spec.ts:384-397` codifies the bug — must be flipped).
2. **Remove `eventLsn` from broker-safe keys** OR rearrange payload spread so envelope wins.
3. **Cap event-name length** in `sanitizeBrokerEventName` (single line + regex precedence).
4. **Cap URL length in `classifyBrokerUrlReason`** before second `new URL` call.
5. **Shared conformance vectors** in `packages/protocol/testdata/broker-url-vectors.json` (deferrable — addresses arch + api LOWs about hand-mirroring).
6. **Docs refresh** (OVERVIEW.md, ipc.md public API list) — deferrable.
7. **Renderer parseBootstrap extraction** — deferrable refactor.

Per-lens reports at `/tmp/triangulation-pr792-pass7/{architecture,distsys,security,api,sre}-report.md`.

FranDias:
## Pass-7 follow-ups landed — 9 findings addressed across 7 commits

Four parallel codex agents in isolated worktrees (broker, logger, protocol-docs, renderer), then cherry-picked onto the PR branch in dependency order (protocol → renderer → logger → broker). Pre-declared scope boundaries → zero conflicts. PR head: `182277ef`.

### Per-finding disposition

| Finding | Severity | Status | Commit |
|---|---|---|---|
| `normalizePayloadValue` byte cap was a UTF-16 code-unit cap | MED (5-lens convergent) | FIXED | `77d74c65` — byte-aware truncation via TextEncoder/Decoder; flips the previously-locked-in test |
| `diagnosticReportBytes` reports values above the 64 KiB cap | MED (5-lens convergent, LOW per most lenses) | FIXED | `2253049a` — `Math.min(byteLength, MAX_DIAGNOSTIC_REPORT_BYTES)` clamp |
| Broker payload can forge `eventLsn` (logger envelope override) | MED (sre) | FIXED option (a) | `f1b6f579` — dropped `eventLsn` from `SAFE_PAYLOAD_KEYS` (envelope-only field) |
| Broker log event names are unbounded → 50 MB JSON DoS | MED (security) | FIXED | `c3459427` — `MAX_LOG_EVENT_NAME_LENGTH = 128` checked before regex |
| `classifyBrokerUrlReason` re-parses unbounded subprocess input | MED (distsys) | FIXED | `182277ef` — `MAX_BROKER_URL_LENGTH = 2048` pre-cap returns `url_too_long` sanitized reason |
| BrokerUrl validation hand-mirrored in 3 places | LOW (arch + api) | FIXED | `44b02263` — `packages/protocol/testdata/broker-url-vectors.json` conformance fixtures + protocol-side consumption (cross-package adoption is a follow-up) |
| `packages/protocol/docs/OVERVIEW.md` stale ("accepts any string URL") | LOW (api) | FIXED | `44b02263` — refreshed with cross-link to ipc.md matrix |
| `ipc.md` public types list omits `BrokerUrl` + stale line numbers | LOW (arch) | FIXED | `44b02263` — added BrokerUrl + removed brittle line-number citations |
| Renderer `parseBootstrap` exported from side-effectful entrypoint | LOW (arch) | FIXED | `c41dab44` — extracted to new `renderer/bootstrap.ts`; `refreshBrokerStatus` stays in main.ts (state-dependent) |

### Verification on integration branch

- `bunx tsc --noEmit` ✓ (desktop + protocol)
- `bunx biome check` ✓ on all changed files
- `bunx vitest run` (desktop) → **228 tests passed** (+8 vs prior 220)
- `bunx vitest run --coverage` → **100/100/100/100** (431 branches, +12 net)
- `bash scripts/test-protocol.sh` → **542 protocol tests passed** (+1 — new conformance vector test)
- `bun run packages/protocol/scripts/demo.ts` → 70/70 scenarios pass
- Pre-push hooks all green (complexity, invariants, IPC allowlist, file-size, protocol-demo, protocol-invariants, protocol-wire-contract)

### Process

4 parallel codex dispatches in isolated worktrees with pre-declared disjoint file ownership:
- A (broker): `broker.ts`, `broker-internal.ts`, broker tests
- B (logger): `logger.ts`, `logger.spec.ts`
- C (protocol-docs): `packages/protocol/**` only
- D (renderer): `renderer/main.ts`, new `renderer/bootstrap.ts`, security spec

Cherry-pick order: protocol (isolated) → renderer (renderer-only) → logger (logger-only) → broker (biggest scope last). Zero conflicts.

Broker agent commit ran into the recurring sandbox ref-lock-write limitation; recovered by repairing the branch ref from outside the sandbox after hook verification confirmed the staged tree was clean.

### Cumulative state of PR #792 after 7 review rounds

- **30 commits**, 14 files / +2469 / −272 vs merge-base
- Brand-tight wire shape (BrokerUrl bare canonical origin, conformance vectors)
- Byte-bounded log paths (all length caps now enforce real UTF-8 bytes)
- Forgery-resistant logger envelope (eventLsn no longer overridable)
- DoS-resistant subprocess inputs (event-name + URL length caps)
- Failure paths terminal (force-stop deadline, ENOENT log-dir recovery, pre-start waiter rejection, broker_ready_invalid surfacing)
- Performance amortized (mkdir-once, in-memory size cache, `for...in` over `Object.entries`, descriptor reads, pre-cap before byte walks)
- Test surface real (no synthetic private-method reflection; queued-callback races covered; conformance vectors consumed)
- Docs aligned (BrokerUrl matrix + upgrade path + env allowlist + public types list)

CodeRabbit's 3 inline + 1 outside-diff comments all addressed across PR history.

PR #800, opened 2026-05-11
fix(review): refreshTick dep — SSE state_change now triggers re-fetch

najmuzzaman-mohammad:
## Summary

- Renames `_refreshTick` → `refreshTick` (removes false "unused" signal)
- Adds `refreshTick` to the fetch effect's dependency array so `review:state_change` SSE events actually trigger re-fetches
- Also fixes the rollback path in `handleStateChange` — it increments the tick to re-fetch, but the empty dep array meant that increment was silently swallowed

## Changes

- `web/src/components/review/ReviewQueueKanban.tsx` — two-line fix

## Test plan

- [ ] Trigger a `review:state_change` SSE event; confirm ReviewQueue re-fetches
- [ ] Cause an `updateReviewState` API failure; confirm optimistic update rolls back via re-fetch
- [ ] No regression in Kanban column rendering or DnD behavior


FranDias (approved):

PR #855, opened 2026-05-14
fix(facts): deterministic fact IDs and append-time dedup

abeeb-rai:
## Summary

- Replace random UUID fact IDs with deterministic SHA-256 hash of `(kind, slug, text, recorded_by)`
- Add JSONL append-time dedup: `factExistsInJSONL` scans existing lines before writing, skipping if a matching ID is present
- Enables dedup at both the JSONL layer and the SQLite `UpsertFact` layer (same ID = `INSERT OR REPLACE`)

## Problem

`FactLog.Append` generated a random `uuid.NewString()` for every call. The same observation recorded twice produced different IDs, bypassing the SQLite upsert dedup and causing duplicate facts in entity briefs.

## Test plan
- [x] `TestFactLog_DeterministicID` — same inputs produce same ID; different inputs produce different IDs
- [x] `TestFactLog_DedupSameFactTwice` — second append of identical fact is silently deduped; only 1 fact in file
- [x] `TestFactLog_DifferentTextNotDeduped` — different fact text still creates separate entries
- [x] All existing fact log tests pass (append, validation, malformed recovery, concurrent appends)

## Summary by CodeRabbit

* **Improvements**
  * Fact records now use deterministic IDs based on immutable inputs, enabling reliable identity across writes.
  * Appends perform on-disk deduplication: identical facts are not stored twice and original creation timestamps are preserved.

* **Tests**
  * Added unit tests verifying deterministic ID generation, deduplication behavior, and that distinct text entries are not deduplicated.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nex-crm/wuphf/pull/855)

### DIFF
--- a/internal/team/entity_facts.go
+++ b/internal/team/entity_facts.go
@@ -19,6 +19,8 @@ package team
 import (
 	"bufio"
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -30,8 +32,6 @@ import (
 	"strings"
 	"sync"
 	"time"
-
-	"github.com/google/uuid"
 )
 
 // EntityKind is the narrow set of wiki subtrees we treat as "entities" for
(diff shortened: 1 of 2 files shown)


FranDias (dismissed):

FranDias (approved):

PR #862, opened 2026-05-14
fix(synth): close queue saturation race in EnqueueSynthesis

abeeb-rai:
## Summary

Move `s.queued[key] = true` after the successful channel send in `EnqueueSynthesis`.

## Problem

`queued[key]` was set before the channel send, then undone on failure. Between the unlock and cleanup, `IsInflightOrQueued` could return true for a job that never entered the channel.

## Fix

Set `queued[key]` only after send succeeds. Failure path needs no cleanup since the key was never set.

## Test plan
- [x] All 13 synthesizer tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed entity synthesis queueing so entities are marked as queued only after their job is successfully enqueued. When the queue is full, the operation now fails cleanly without leaving entities in a reserved/incorrect queued state, preventing prior timing-related inconsistencies.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nex-crm/wuphf/pull/862)

### DIFF
--- a/internal/team/entity_synthesizer.go
+++ b/internal/team/entity_synthesizer.go
@@ -266,17 +266,17 @@ func (s *EntitySynthesizer) EnqueueSynthesis(kind EntityKind, slug, requestBy st
 		EnqueuedAt: time.Now().UTC(),
 		ID:         id,
 	}
-	s.queued[key] = true
 	s.mu.Unlock()
 
 	select {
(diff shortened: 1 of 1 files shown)


FranDias (dismissed):

FranDias (approved):

PR #863, opened 2026-05-14
fix(broker): close entity graph write race and add launcher panic recovery

abeeb-rai:
## Summary

Two high-severity fixes in one PR:

**H4 — Entity graph write race** (`entity_graph.go`)
- `RecordFactRefs` now holds `g.mu` through the wiki worker enqueue
- Previously, the lock was released before enqueue, allowing concurrent writers to read stale content and lose each other's edges
- Safe: wiki worker never calls back into EntityGraph (no circular lock)

**H5 — Launcher goroutines without panic recovery** (`launcher_boot.go`)
- All 6 background goroutines wrapped in `defer recoverPanicTo()`
- A panic in any loop is now caught and logged to `~/.wuphf/logs/panics.log` instead of crashing the process
- Reuses the existing pattern from `notifier_loops.go`

## Test plan
- [x] Behavior-preserving — same logic, just safer concurrency

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved thread safety for concurrent data operations to prevent race conditions.
  * Enhanced system resilience by adding panic recovery mechanisms to long-lived background processes, preventing unexpected shutdowns.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nex-crm/wuphf/pull/863)

### DIFF
--- a/internal/team/entity_graph.go
+++ b/internal/team/entity_graph.go
@@ -214,12 +214,14 @@ func (g *EntityGraph) RecordFactRefs(ctx context.Context, fact Fact) ([]EntityRe
 		return nil, nil
 	}
 
-	// Build the new file contents under the lock — a concurrent RecordFactRefs
-	// must see this write as atomic for the read-then-append contract to hold.
-	// Release the lock BEFORE enqueuing, so a read path that takes g.mu in the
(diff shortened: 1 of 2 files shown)


FranDias (approved):

PR #871, opened 2026-05-15
fix(test): eliminate data race in TestDecisionPacketWikiPromotionOnMerged

abeeb-rai:
## Summary

Fix flaky `TestDecisionPacketWikiPromotionOnMerged` DATA RACE that fails ~1 in 10 runs with `-race`.

## Root cause

`writeWikiPromotionLocked` spawns a goroutine that calls `WikiWorker.Enqueue()` (channel send). The test calls `worker.Stop()` (channel receive/close) which races with the goroutine's send at the Go runtime channel level.

## Fix

1. **Production**: capture `b.wikiWorker` into a local `worker` before spawning the goroutine
2. **Test**: replace `worker.Stop()` with `cancel()` + `<-worker.Done()` — cancel signals the drain loop to exit cleanly after processing in-flight requests, avoiding the Stop/Enqueue channel race. Poll for the promotion file first to ensure the goroutine has finished.

Verified: 20/20 passes with `-race` (previously ~1/10 failure rate).

## Test plan
- [x] `go test -race -count=20 -run TestDecisionPacketWikiPromotionOnMerged ./internal/team/` — 20/20 pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition in the broker's wiki promotion process that could occur during concurrent operations or shutdown sequences.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nex-crm/wuphf/pull/871)

### DIFF
--- a/internal/team/broker_decision_packet.go
+++ b/internal/team/broker_decision_packet.go
@@ -987,11 +987,14 @@ func (b *Broker) writeWikiPromotionLocked(taskID string, packet DecisionPacket) 
 	}
 	commitMsg := fmt.Sprintf("decision: promote %s on merge", taskID)
 	ctx := b.wikiPromotionContext()
+	// Capture the worker locally so the goroutine doesn't race with
+	// concurrent nil-assignment on b.wikiWorker (e.g. broker shutdown).
+	worker := b.wikiWorker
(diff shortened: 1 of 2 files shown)


FranDias (approved):

PR #898, opened 2026-05-18
fix(nex): degrade gracefully when only the @nex-ai/nex shim is installed

FranDias:
## Problem

A user running `npx wuphf` hit this in the onboarding **"Power up with Nex"** step — a raw JSON error blob rendered straight at them:

```
{"error":"nex-cli setup hi@mustafa.li: nex-cli binary not found.\n\n Install it with:\n curl -fsSL https://...install.sh | sh\n\n Then run your command again.","status":"error"}
```

### Root cause chain

```mermaid
flowchart TD
  A["Onboarding → POST /nex/register {email}"] --> B["nex.Register → nex.Run(setup, email)"]
  B --> C["BinaryPath() accepts 'nex-cli' OR 'nex'"]
  C --> D["Finds 'nex' — the @nex-ai/nex npm SHIM, not the real binary"]
  D --> E["Runs 'nex --cmd setup ...'"]
  E --> F["Shim can't find real nex-cli → prints install blurb to stderr, exits 1"]
  F --> G["Run() wraps stderr → broker 502 {status:error, error:...}"]
  G --> H["UI fallback never triggers → raw JSON shown"]
```

The shim (`@nex-ai/nex`) ships only a delegator that needs a separately-installed binary. When a user has the shim but not the binary, WUPHF's `BinaryPath()` accepts the `nex` alias, considers Nex installed, runs `setup`, and the shim exits non-zero. The UI's fallback check only matched `"502"` / `"not installed"` — but the error message is the JSON body (no `"502"`) saying `"binary not found"` — so it fell through to the raw-error display.

> The packaging root cause is fixed separately in nex-crm/nex-as-a-skill#161 (npm postinstall now downloads the real binary). This PR makes WUPHF degrade gracefully regardless.

## Fix

**`internal/nex/cli.go`** — `Run()` detects the shim's `nex-cli binary not found` stderr signature on a failed shell-out and returns `ErrNotInstalled`. A shim-without-binary now takes the same path as a bare missing binary, instead of surfacing the shim's raw install blurb.

**`web/.../NexConnectPanel.tsx`** — parses the broker's JSON error body so the panel shows the human-readable message (never a raw blob), and broadens the fallback trigger (`not installed` / `binary not found` / `502`) so an unusable nex-cli reliably flips the card to the register-externally affordance.

## Verification

- `internal/nex` — new `TestRun_ShimWithoutBinary`; full package green. `go build ./...`, `go vet`, `golangci-lint` clean.
- `internal/team` — existing `broker_nex_register` tests still green (502 contract unchanged).
- `web` — new `NexConnectPanel.test.tsx` covers success, shim-only, not-installed, and genuine-error paths (4 pass). `tsc --noEmit` + Biome clean.

## Release Notes

* **New Features**
  * Enhanced Nex integration setup with fallback UI displaying installation instructions when nex-cli is unavailable.

* **Bug Fixes**
  * Improved error handling for missing nex-cli binary, detecting npm shim-only cases and showing clear, user-friendly messages instead of technical errors.

* **Refactor**
  * Extracted reusable command display component with clipboard copy functionality.

* **Tests**
  * Added E2E and unit test coverage for Nex Connect scenarios and missing binary detection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nex-crm/wuphf/pull/898?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -852,6 +852,7 @@ jobs:
             tests/route-matrix.spec.ts \
             tests/route-regressions.spec.ts \
             tests/sidebar-scroll.spec.ts \
+            tests/nex-connect.spec.ts \
             tests/chat.spec.ts \
             tests/slash-commands.spec.ts \
             tests/dm.spec.ts \
--- a/internal/nex/cli.go
+++ b/internal/nex/cli.go
@@ -89,12 +89,24 @@ func quoteArg(s string) string {
 	return `"` + strings.ReplaceAll(strings.ReplaceAll(s, `\`, `\\`), `"`, `\"`) + `"`
 }
 
+// shimNotInstalledMarker is the phrase the @nex-ai/nex npm shim prints to
+// stderr when it runs but cannot locate the real nex-cli binary. BinaryPath()
+// accepts the `nex` alias, which may resolve to that shim; a shim with no
+// binary behind it is functionally "not installed", so Run() translates this
+// signature back to ErrNotInstalled instead of surfacing the shim's raw
+// install blurb to callers (and, ultimately, to the onboarding UI).
+const shimNotInstalledMarker = "nex-cli binary not found"
+
 // Run executes `nex-cli --cmd "<args...>"` with the given context/timeout and
 // returns stdout (trimmed) on success. nex-cli refuses to start without a TTY
 // unless --cmd/--script is used, so every shell-out goes through --cmd with
 // the args joined (and quoted where necessary) into a single command string.
 // A non-zero exit code, missing binary, or context timeout are all returned
 // as errors. The caller is responsible for logging and falling back.
+//
+// A shell-out that fails because the resolved binary is really the npm shim
+// without its backing binary is reported as ErrNotInstalled, so callers can
+// errors.Is-check a single "not installed" condition.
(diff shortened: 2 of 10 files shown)


FranDias:
### Update — Connect Nex panel now offers the fix, not just an escape hatch

Per follow-up: instead of only pointing users to `nex.ai/register`, the not-installed fallback now leads with the **copy-paste install command** for nex-cli.

**Decision — copy-paste, not one-click.** wuphf already states the policy in the adjacent Local LLMs panel: *"install commands are copy-paste only (we never run shell commands for you)."* NexConnectPanel sits in the same Settings → Integrations area, so it follows suit. Copy-paste also means zero telemetry, and no npm-on-PATH / global-write-permission assumptions.

The fallback state now shows:
- `curl … install.sh | sh` in a `CommandRow` (extracted from `SettingsApp` into `web/src/components/ui/CommandRow.tsx` so both panels share one implementation);
- an **"I've installed it — try again"** button that returns to the email form with no page reload;
- the `nex.ai/register` escape hatch, kept for machines with no terminal access (e.g. Windows — nex-cli ships no Windows binary).

Verified: `NexConnectPanel.test` (5 pass) + `nex-connect.spec.ts` (3 pass, real Chromium) updated for the install-command and retry paths.

PR #922, opened 2026-05-20
feat(protocol): add browser route codec subpath

FranDias:
## Summary

Adds a `@wuphf/protocol/browser` subpath whose entire transitive import
graph contains zero `node:*` imports. Enables the Electron renderer
(sandboxed, `nodeIntegration: false`) to import the route-envelope
codecs without dragging `node:crypto` / `node:buffer` into its bundle.

The R1 enabler for the Stream B renderer build.

## Mechanism

Pure code-motion + small isomorphic swaps — **zero wire-behaviour change**.
Conformance vectors and the Go cross-language oracle (214 vectors) stay
byte-identical.

- **`Buffer.byteLength(s, "utf8") -> new TextEncoder().encode(s).length`**
  in `receipt.ts` and `sanitized-string.ts`. `TextEncoder` is a global in
  both Node and browsers; UTF-8 byte count is identical. Drops two
  `node:buffer` imports.
- **`sha256.ts` split**: pure brand/validators (`Sha256Hex`,
  `isSha256Hex`, `asSha256Hex`, `SHA256_HEX_RE`) stay in `sha256.ts`; the
  `node:crypto`-using `sha256Hex()` moves to a new `sha256-node.ts`.
- **`thread-browser.ts` + `route-envelopes-node.ts`**: relocates the
  Node-only thread-content-hash re-derivation outside the browser graph
  while preserving strict re-derivation on the Node entry. Default
  `index.ts` consumers see the same public surface; the browser entry
  exposes the same wire codecs (`approval*`, `thread*` route envelopes,
  signed-approval-token types, branded IDs).
- **`src/browser.ts`** + `package.json` `exports` `"./browser"`: the
  curated browser surface.
- **`tests/browser-subpath.spec.ts`**: a static guard test that walks
  the transitive import graph of `src/browser.ts` via the TypeScript
  compiler API and FAILS if any `node:*` / bare `buffer` / bare `crypto`
  specifier creeps in. Future regressions are caught by `bun run test`.

## Commits

| Commit | What |
|---|---|
| `b1e29aaa` | Initial subpath + split |
| `51322608` | Merge of main (brings in #918 + #920 + the lint pin) |
| `3349c824` | Lift protocol branch coverage 89.51% → 90.59% (covers `thread-browser.ts` + `route-envelopes-thread-replay-check.ts` new branches) |
| `8171aef1` | Keep #920's replay-check codec on the node-free import graph (route via `receipt-types.ts`; export from `route-envelopes-node.ts`) |

[...]

PR #923, opened 2026-05-20
feat(desktop): R1 — renderer foundation (React + TanStack + Tailwind on the loopback broker)

FranDias:
## Summary

R1 of the Stream B renderer plan. Grows `apps/desktop/src/renderer/` from
its 350-LOC TypeScript status shell into a React application that talks
to the broker over the loopback HTTP/SSE listener via the just-shipped
`@wuphf/protocol/browser` subpath ([#922](https://github.com/nex-crm/wuphf/pull/922)).

This is the **prerequisite the RFC §15 Stream B plan assumed existed but
didn't** — R2 (`feat/work-board`) and R3 (`feat/unified-approvals-inline`)
each land surfaces on top of this foundation. The week-8 alpha gate is
the work-board + sticky-pin region being live for any approval-flow demo.

## Decisions taken (all confirmed with the lead engineer before scaffold)

| # | Decision | Rationale |
|---|---|---|
| 1B | **Main-process dynamic CSP injection** (not wildcard, not on-disk swap) | `session.webRequest.onHeadersReceived` re-derives the CSP per response from the supervisor's current broker URL. Survives broker restarts on new ports. Strict baseline meta-tag CSP without the broker origin acts as a fallback if the header injection fails. The broker URL is validated (HTTP, explicit port, loopback hostname, exact-origin match) before being added to `connect-src` — non-loopback URLs throw |
| 2A | **Aggressive coverage from day one** | Threshold stays at 98/98/98/98 (unchanged from baseline). Measured 98.23% branches comfortably exceeds the floor; +45 tests for 30 new files |
| 3A | **Single PR, three commits** | Coherent reviewable unit. Split would have R1a with nothing exercisable |

Plus: **first-party tailwind UI primitives** (Button, Card, StatusBadge —
~150 LOC total) mirroring `web/src/components/ui/`'s pattern. No
`@radix-ui/*`, no shadcn — `web/` doesn't use them either, and the
lightest option matches the request.

## Three commits

| Group | Commit | What |
|---|---|---|
| 1 | `0dd6756f` `build(desktop): add renderer react foundation` | deps + electron-vite renderer plugin + Tailwind + tsconfig.renderer JSX |
| 2 | `217fd84a` `fix(desktop): inject broker csp dynamically` | `src/main/csp.ts` — `installDynamicRendererCsp`, validated origin, `+ csp-injection.test.ts` covering broker-URL present / absent / restart / non-loopback-rejected variants |
| 3 | `efae621d` `feat(desktop): add renderer app shell` | React shell + `BrokerBootstrapProvider` (wraps the existing `getBrokerStatus → /api-token → bearer` flow) + TanStack Router skeleton + TanStack Query + `api/` layer over `/api/v1/*` using `@wuphf/protocol/browser` codecs + authenticated fetch-stream SSE hook + first-party UI primitives + every component/hook/provider/api-client tested |

### One implementation note worth flagging

The SSE hook uses **`fetch` with `AbortController`** instead of `EventSource`
because broker stream frames require a bearer token, and the standard
`EventSource` API can't carry an `Authorization` header. Putting the
bearer in the query string would expose it in URLs (logs, history, etc.);
fetch-streaming is the standard secure pattern. Tested with a mock fetch
returning a streamed body; reconnection on abort handled by re-running
the `useEffect`.

## Hard rules honoured (apps/desktop/AGENTS.md)

- Rule 4: contextBridge allowlist UNCHANGED. **No new IPC verb** added.
  `check:ipc-allowlist` passes.
- Rule 7: CSP stays strict. Dynamic header injection adds *only* the
  concrete broker origin to `connect-src`; everything else
  (`script-src 'self'`, no `unsafe-inline`, no `unsafe-eval`, no remote
  script sources) preserved.
- Rule 10: **No app data IPC.** Renderer reaches `/api/v1/*` over loopback
  HTTP with the bearer; main process touches nothing in `~/.wuphf/`.
- Rule 13: Strict TypeScript. No `any`, no `@ts-ignore`, no biome-ignore.
- Rule 14: Coverage ratchet preserved.

## Test plan

- [x] `cd apps/desktop && bun run typecheck` (3 TS projects)
- [x] `cd apps/desktop && bun run lint`
- [x] `cd apps/desktop && bun run test` — 282 tests
- [x] `cd apps/desktop && bun run test:coverage` — stmts 99.17 / branches 98.23 / fns 100 / lines 99.56 (threshold 98/98/98/98)
- [x] `cd apps/desktop && bun run check:ipc-allowlist`
- [x] `cd apps/desktop && bun run check:invariants`
- [x] `cd apps/desktop && bun run check:build-externals`
- [x] `cd apps/desktop && bun run build` — electron-vite build green
- [x] `bash scripts/check-file-size.sh` no failures
- [x] `cd packages/protocol && bunx tsc --noEmit && bash ../../scripts/test-protocol.sh` — protocol still green
- [x] Pre-R1 → post-R1 desktop coverage: 98.63% branches → 98.23% branches (margin shrinkage from +30 new files, still well above the 98 ratchet)

## What's NOT in R1 (intentional)

- No work-board surface (that's **R2 `feat/work-board`**).
- No thread-detail / sticky-pin (that's **R3 `feat/unified-approvals-inline`**).
- No new IPC verb. No new app-data IPC.
- The `npx wuphf` distribution stays deferred — the alpha is the
  Electron app.

## Summary by CodeRabbit

* **New Features**
  * Desktop app: full React app shell with routing, broker HTTP API client, SSE-based real-time events, React Query integration, and new UI components (Button, Card, StatusBadge).

* **Improvements**
  * Strengthened security: dynamic Content-Security-Policy injection and clarified loopback HTTP/SSE boundary; clearer bootstrap/retry and dev vs packaged behavior.
  * Docs/README updated to reflect architecture and data flow.

* **Tests**
  * Extensive unit/integration suites plus new Playwright e2e covering bootstrap, CSP, SSE, API client, and UI.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nex-crm/wuphf/pull/923?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


FranDias:
## Manual verification — R1 boots end-to-end

Walked the R1 renderer through a full Electron boot against a real broker subprocess after landing 4 small dev-runtime fixes (commit `2b3df9b4`). The screenshots below were captured from a packaged-style launch:

```sh
bun --cwd apps/desktop run build
WUPHF_RECEIPT_STORE_PATH= \
WUPHF_WEBAUTHN_STORE_PATH= \
WUPHF_RENDERER_DIST=$(pwd)/apps/desktop/out/renderer \
  bun --cwd apps/desktop run preview
```

The two empty env vars are the dev escape hatch added in `2b3df9b4` — they keep the broker on the in-memory receipt + webauthn stores so the better-sqlite3 ↔ Electron 42 ABI gap (see "out of scope" below) doesn't block boot.

### Ready state — full boot

The renderer mounts, the bootstrap probe fetches `/api-token` from the broker, the bearer-authed `GET /api/health` succeeds, and the index route reads the app version through the IPC bridge.

R1 ready state

Things this proves:
- broker utility process forks cleanly and reports `broker_ready`
- the renderer loads from `${brokerUrl}/` (same-origin) and the dynamic CSP from `installDynamicRendererCsp` reaches `connect-src 'self' http://localhost:51011`
- `BrokerBootstrapProvider` resolves `loading → ready` with a real bearer
- `Card`, `Button`, and `StatusBadge` primitives render with the right variants
- the version badge proves `window.wuphf.getAppVersion` round-trips through the contextBridge allowlist

### Error path — broker subprocess gone

Killing the broker after boot puts the renderer into the error path; this is the same UI as a cold "broker never came up" failure. The retry button re-invokes the bootstrap probe.

### Out of scope, tracked separately

Two real product gaps surfaced while booting and are filed for follow-up PRs rather than expanded into this one:

| # | Gap | Why deferred |
|---|---|---|
| 1 | better-sqlite3 12.x does not have a prebuild for Electron 42's V8 ABI (146) and its C++ source no longer compiles against the new `v8::External::New` signature | Upstream issue; needs a better-sqlite3 release or an Electron downgrade. The `WUPHF_RECEIPT_STORE_PATH=` escape hatch makes dev boot until then. |
| 2 | `packages/broker` does not emit `Access-Control-Allow-Origin` even when the request's `Origin` is in the trusted-origins allowlist. `bun run dev` (vite at `localhost:5173`, broker at `127.0.0.1:PORT`) fails the bootstrap probe with CORS. | Broker-side change, not renderer; will land in a dedicated broker PR. `bun run preview` (renderer served from `${brokerUrl}/`, same-origin) works today. |

### Coverage check

`bun --cwd apps/desktop run test:coverage` after the dev-runtime commit:
```
Statements   : 99.39% ( 1310/1318 )
Branches     :  98.62% ( 715/725 )
Functions    : 100.00% ( 279/279 )
Lines        :  99.59% ( 1244/1249 )
```
Above the 98/98/98/98 ratchet floor; 300 tests pass (3 new for `DEV_RENDERER_CSP` + `isPackaged` plumbing).

FranDias:
## Review pass — addressed CodeRabbit + convergent triangulation HIGHs (commit `659cfcd4`)

### CodeRabbit (9/9 addressed)

| # | File | Disposition |
|---|---|---|
| 1 | `apps/desktop/AGENTS.md:80-83` — "trusted" wording | **FIXED** — clarified the renderer is packaged-first-party but still treated as untrusted (sandbox + closed allowlist) |
| 2 | `apps/desktop/docs/modules/security-model.md:84` — `base-uri 'self'` mismatch | **FIXED** — aligned doc with runtime `base-uri 'none'` |
| 3 | `apps/desktop/electron.vite.config.ts:67` — silent CSP no-op | **FIXED** — `devRendererCspPlugin` throws if the meta tag is missing |
| 4 | `apps/desktop/scripts/check-invariants.sh:90` — case-sensitive grep | **FIXED** — `-Fi`, also added `*.tsx` |
| 5 | `apps/desktop/src/renderer/app/routes/index.tsx:25` — unhandled rejection | **FIXED** — `.catch()` keeps the chip hidden on IPC failure |
| 6 | `apps/desktop/src/renderer/sse/useBrokerEvents.ts:246` — per-chunk CRLF | **FIXED** — normalize on accumulated buffer instead |
| 7 | `apps/desktop/tests/renderer/index-route.test.tsx:83` — spy restore | **FIXED** — `try/finally` |
| 8 | `packages/protocol/src/browser.ts:233` — untyped key allowlist | **FIXED** — `as const satisfies readonly (keyof T)[]` tuples for all 4 SSE allowlists |
| 9 | `apps/desktop/package.json:23` — typecheck skips e2e | **FIXED** — `typecheck` now runs `tsconfig.e2e.json` too |

### Triangulation findings — convergent HIGHs (2+ lenses agreeing) addressed

| # | Convergence | Disposition |
|---|---|---|
| A | **security-1 + architecture-1** — unpackaged preview gets the dev CSP | **FIXED** — replaced the `isPackaged` flag with `devRendererOrigin: string \| null`; the listener compares `details.url`'s origin to that value and applies `DEV_RENDERER_CSP` only on a match. Preview, broker-served renderer, and packaged loads all stay strict. The origin is validated (loopback http + explicit port + bare origin) before it can be honored. |
| B | **security-2 + sre-1 + architecture-3** — migration assets fragile | **FIXED** — `copyBrokerMigrationsPlugin` now reads an authoritative `BROKER_MIGRATION_FILES` list, refuses symlinks via `lstatSync`, and fails the build if Rollup chunking moves `broker-entry.js` out of `out/main/` (where the SQL files are emitted). Mutation-tested by the e2e: removing the migrations makes the test time out. |
| C | **security-3** — dev-origin env leak in packaged builds | **FIXED** — packaged builds `delete process.env.WUPHF_DEV_RENDERER_ORIGIN` before deriving the dev origin so a wrapper / CI runner cannot extend the broker's trusted-origin set in a shipped binary. |

### Triangulation findings — single-lens, **DEFERRED** with rationale

These are real, but cost > benefit at the R1-foundation stage; tracked for dedicated follow-ups so the R1 PR doesn't grow into "everything that touches the renderer."

| Lens | Finding | Defer rationale |
|---|---|---|
| api/types | `parseBootstrap` in `src/renderer/bootstrap.ts` duplicates `apiBootstrapFromJson`, `ApiToken`, `BrokerUrl` brand definitions | requires extending `@wuphf/protocol/browser` with the bootstrap codec + brands. Belongs with the broader "drop renderer-local protocol mirrors" PR (will pair with the SSE id-vs-data.id verification). Risk today is bounded: the renderer-local parser is byte-for-byte cross-checked against the protocol's vectors via the e2e bootstrap probe. |
| sre/architecture | dev renderer doesn't rebind after a broker restart (`url1 -> null -> url2` leaves the window on `url1`) | the packaged broker-pinned-window path already handles this; dev-mode windows are flagged by SRE as a known gap. Fix requires the bootstrap provider to subscribe to `getBrokerStatus()`. Out of scope for R1's renderer foundation. |
| sre | no `AbortSignal` / timeout on bootstrap + API fetches | needs end-to-end signal threading across `loadBrokerBootstrap`, `BrokerApiClient`, and the TanStack Query helpers. Worth a dedicated PR with regression tests. |
| sre | renderer load/render failures (e.g. `loadURL` rejection) aren't first-class events | needs a top-level renderer ErrorBoundary + main-process structured event for `loadURL` rejections. R2 will need these too; better to land alongside R2's first real surface. |
| perf | initial bundle pays for unused approval / NFKC code (~1,044 kB raw / ~218 kB gzip) | requires narrowing protocol's `/browser` subpath into key-only + per-feature surfaces. Owned by the same protocol-codec follow-up as the api/types finding. |
| perf | SSE invalidation calls `invalidateQueries` synchronously per frame | requires batching at `useBrokerEvents`'s consumer layer; will land with R2 where the actual cache keys exist. |
| perf | SSE frame parser allocates heavily | hot path under 100x event rate. Optimize once we have a real workload to measure against (R2/R3). |
| perf | CSP listener runs for every response | small allocation per response; gated behind `webRequest.onHeadersReceived` filters when there is something to optimize. |
| security | dev SQLite override accepts unsafe filesystem paths | dev-only escape hatch; the `better-sqlite3` ↔ Electron 42 ABI gap that motivated it is itself the priority follow-up. |
| architecture | renderer CSP mode has multiple sources of truth (config, main, index.html) | the convergent fix above narrowed the surface; centralization belongs with the bundle-size and protocol-codec PR. |
| coverage | `copyBrokerMigrationsPlugin` + `devRendererCspPlugin` have no unit tests | the e2e mutation-tested both end-to-end; a unit-test pair would catch regressions earlier. Will land with the protocol-codec follow-up. |

### Guards

`bun --cwd apps/desktop run typecheck` ✓ (now includes `tsconfig.e2e.json`)
`bun --cwd apps/desktop run lint` ✓
`bun --cwd apps/desktop run test` ✓ 302/302
`bun --cwd apps/desktop run test:coverage` ✓ 99.10 / 98.37 / 99.64 / 99.44 (above 98 floor)
`bun --cwd apps/desktop run build` ✓
`bun --cwd apps/desktop run test:e2e` ✓ 1/1 (1.0s)
`bun --cwd packages/protocol run test` ✓ 966/966

FranDias:
Superseded by #927 — the R1 renderer foundation commits were folded into the work-board branch rebuild so we could land R1 + R2 + threads-subsystem wiring in a single tidy merge. All R1 changes are now in main.

Issue #939, opened 2026-05-21
fix(onboarding): wizard step counter consistency (B-05 + U-04 + Q-05)


najmuzzaman-mohammad:
Findings **B-05**, **U-04**, **Q-05** in `docs/qa/2026-05-20-icp-walkthrough.md`. Epic #930.

Wizard headers today read: `STEP 1 OF 5`, `STEP 2 OF 5 · WHO YOU ARE` (but the prompt is "What does Acme Postmark do?" — a company description, not who-you-are), `PHASE: WEBSITE` (no step counter), `PHASE: SCAN` (no step counter), `STEP 3 OF 5 · PICK A STARTING BLUEPRINT`, `STEP 5 OF 5 · FIRST TASK` (step 4 — team trim — silently skipped on the scratch path).

**Fix:** Either drop the step counter entirely and let the chat-mode wizard be conversational (preferred — matches the product principle of feeling like a colleague, not a form), or fix the numerator/denominator + label so it stays consistent on every path.

**Severity:** P2 — cosmetic but breaks the immersion.

Issue #935, opened 2026-05-21
fix(wiki): article + notebook loaders stuck on 'Loading article…' (B-08 + B-17 + U-08)


najmuzzaman-mohammad:
Findings **B-08**, **B-17**, and **U-08** in `docs/qa/2026-05-20-icp-walkthrough.md`. Epic #930.

`#/wiki/team/about/README.md` shows "Loading article…" forever even when the article exists on disk. `#/wiki/notebooks` does the same. There is no timeout, no error message, no retry button. The empty-state voice across the app is also inconsistent — Inbox uses "Inbox zero." and the Activity log uses "No agents are visibly active right now.", while Wiki uses an indefinite spinner.

**Fix:**
1. Diagnose the failing article fetch (path encoding? wrong API route for nested `team/about/`?).
2. Add a 5s loader timeout that surfaces the real error + a Retry button.
3. Notebooks empty state should be `Agents will write here as they work — nothing yet.` not a spinner.
4. Pick a single empty-state voice and apply it across Wiki, Notebooks, Inbox, Activity, Skills.

**Severity:** P1 — silently broken loaders look like the whole app is broken.

PR #970, opened 2026-05-21
fix(onboarding): swap scan chip for next pending suggestion on terminal status (#936)

### LINKED ISSUE #936: fix(onboarding): clear scan chip when phase auto-advances after failure (B-10)
Finding **B-10** in `docs/qa/2026-05-20-icp-walkthrough.md`. Epic #930.

After a scan failure, the backend correctly advances to `PhaseBlueprint` and writes a `blueprint-pick` pending suggestion — but the frontend keeps the "Scanning https://example.com…" chip pinned in the composer area with no buttons. The user is stranded until a hard browser reload, after which the blueprint chips render correctly.

**Fix:**
1. When a `ceo_scan_chip` reaches terminal status (`done`/`failed`), the composer area must drop the chip and pick up the next `PendingSuggestion` automatically.
2. Add a Playwright assertion: "after the scan fails, the blueprint chip-row buttons render without a page reload".

**Severity:** P0 — silently strands the user with no recovery path other than F5.

najmuzzaman-mohammad:
## Summary

After a website scan fails during onboarding, the broker correctly transitions to `PhaseBlueprint` and writes a new `PendingSuggestion` (`blueprint-pick`), but the frontend kept showing the read-only "Scanning..." chip until a hard reload. The user was stranded with no buttons.

Two surgical changes on the frontend (Go broker is left alone — it correctly writes the new `PendingSuggestion` and broadcasts the new card message):

1. **`useBrokerEvents` now invalidates `["onboarding-state"]`** alongside `["messages"]` on every SSE `message` event. The broker emits the new blueprint chip-row as a CEO DM message in the same critical section that updates `PendingSuggestion`, so any message arrival is a reliable signal that the onboarding state may have moved. The query has no observers outside onboarding, so the invalidate is a no-op on the regular shell.
2. **`CeoCardSection` keys its inner `StructuredMessageCard` on the suggestion id** and exposes `data-suggestion-id` on the section element. Keying is defense in depth on top of the existing `useEffect`-based stage reset: a forced unmount guarantees no stale child state (focus, controlled inputs, sub-component refs) leaks across the `scan-chip → blueprint-pick` swap.

## Test plan

- [x] `bunx tsc --noEmit` clean
- [x] `bunx biome check` clean (only pre-existing warnings on unrelated lines in InterviewBar.tsx)
- [x] `bash scripts/test-web.sh web/src/components/messages/InterviewBar.scanChipSwap.test.tsx` — new regression test passes
- [x] `bash scripts/test-web.sh` — full web suite (170 files, 1617 tests) passes
- [x] `go build -o /tmp/wuphf-scan-chip-fix ./cmd/wuphf` clean
- [x] Screenshot harness `web/e2e/screenshots/scan-chip-desync-fix.mjs` captures the before/after states with stubbed `/onboarding/state` swaps (mirrors the SSE-driven invalidate)

## Regression test

`web/src/components/messages/InterviewBar.scanChipSwap.test.tsx` mounts `CeoCardSection` with the scan chip as the initial pending suggestion, swaps to the blueprint chip-row via the `OnboardingDMContext`, and asserts the scan chip is gone and the Bookkeeping / Start-from-scratch options render.

## Why frontend-only

The Go broker side of this loop is correct: `runScanPhase` → `advanceAfterScan` → `advancePhase(PhaseBlueprint)` writes the new `PendingSuggestion`, appends the blueprint card as a CEO DM message, and the message append fires the existing SSE `message` event. The bug was purely that the frontend cache for `/onboarding/state` did not listen for that event.

Closes #936

## Screenshots

Before the fix — composer slot stuck on the scanning chip with no buttons (`data-phase="scan"`):

01-stranded-before-fix

After the fix — slot swaps to the blueprint chip-row (`data-phase="blueprint"`), Bookkeeping / Niche CRM / YouTube Factory / Start from scratch options render:

02-recovered-after-fix

## Release Notes

* **Bug Fixes**
  * Fixed an issue where onboarding card elements would not properly reset when transitioning between different scanning phases.

* **Tests**
  * Added regression test and E2E screenshot tests to verify onboarding UI correctly re-renders during phase transitions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nex-crm/wuphf/pull/970?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/web/e2e/screenshots/scan-chip-desync-fix.mjs
+++ b/web/e2e/screenshots/scan-chip-desync-fix.mjs
@@ -1,0 +1,192 @@
+// Capture the scan-chip → blueprint chip-row swap that fixes #936.
+//
+// Two frames:
+//   01-stranded — what users saw BEFORE the fix: scanning chip pinned in the
+//                 composer slot with no buttons. Reproduced by stubbing
+//                 /onboarding/state to return the scan chip as the pending
+//                 suggestion even though the broker has moved on.
+//   02-recovered — what users see AFTER the fix lands: the blueprint chip-row
+//                  swaps into the slot automatically (no hard reload),
+//                  surfacing the Bookkeeping / Niche CRM / scratch options.
+//
+// The two frames use the same mock backend layer; only the
+// /onboarding/state response differs between mounts, which is exactly what
+// the SSE-driven onboarding-state invalidate (useBrokerEvents.ts) causes
+// in production.
+
+import process from "node:process";
+
+import {
+  DEFAULT_BASE,
+  installCommonMocks,
+  launchBrowser,
+  shotPage,
+} from "./lib.mjs";
+
+const OUT = process.env.WUPHF_SCREENSHOTS_OUT;
+if (!OUT) {
+  console.error("WUPHF_SCREENSHOTS_OUT is not set; run via publish.sh");
+  process.exit(2);
+}
+
+const baseAnswers = {
+  company_name: "Acme Billing",
+  description: "Subscription billing for indie SaaS",
+  website_url: "https://example.com",
+};
+
+// Frame 1 — the broken state: scan chip pinned with no buttons.
+const STATE_STRANDED = {
+  onboarded: false,
+  phase: "scan",
+  form_answers: baseAnswers,
+  pending_suggestion: {
+    id: "scan-progress-example-com",
+    phase: "scan",
+    kind: "ceo_scan_chip",
+    payload: {
+      url: "https://example.com",
+      status: "scanning",
+      scanning_label: "Scanning example.com…",
+    },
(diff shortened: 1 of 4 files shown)


najmuzzaman-mohammad:
Verified on verify/fix-wave-3 with all 12 PRs merged. Go tests green (internal/team 96s + workspaces + cmd). Web tests 1630/1630. Browser walkthrough confirmed: step counter gone with WHAT YOU DO header (#939), blueprint chips appear without reload after scan failure (#936), wiki article renders + notebooks shows list + timeout-retry UI works (#935). Marking ready.

FranDias (approved):

PR #996, opened 2026-05-28
[release] fix(workspaces): auto-pause running workspace on shred

najmuzzaman-mohammad:
## Summary

Shredding a workspace from the rail used to silently fail when its broker was still running — the icon stayed in the sidebar, the broker kept the port bound, and the toast was easy to miss against the modal backdrop.

The `Shred()` orchestrator already knew the right answer (call `Pause`, which runs the SIGTERM → SIGKILL → tmux ladder) but punted back to the caller with `"workspace is running; pause it first"`. No UI surface wired up that two-step, so confirmed shreds on a live workspace just bounced.

This collapses the two-step into the one intent. Confirmed shred now:
1. Calls `Pause` if the workspace is running or its port still answers.
2. Re-reads the registry post-pause so state is fresh.
3. Continues with the existing backup + registry-removal flow.

If pause genuinely can't kill the broker (survives SIGKILL), Shred surfaces a wrapped error instead of leaving the workspace half-dead.

## Test plan

- [x] `go test -race ./internal/workspaces/ -run "Shred|Pause"` — all green, including the new `TestShredAutoPausesRunningWorkspace`
- [ ] Live verify: spawn a non-main workspace, confirm shred from the rail, verify icon disappears and `lsof -iTCP:<port>` shows the port released

## Summary by CodeRabbit

* **Bug Fixes**
  * Workspace deletion now automatically pauses running workspaces so removal proceeds without error; if the workspace disappears during this step, deletion reports it as not found.
* **Tests**
  * Added automated coverage verifying deletion succeeds for workspaces that were running and ensures runtime data is removed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nex-crm/wuphf/pull/996?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/workspaces/orchestrator.go
+++ b/internal/workspaces/orchestrator.go
@@ -394,7 +394,16 @@ func Shred(ctx context.Context, name string, permanent bool) (string, error) {
 	// behind shred keeps the UX promise that a confirmed shred actually
 	// teardown everything (process + port + tree).
 	if target.State == StateRunning || target.State == StateStarting || probePort(target.BrokerPort) {
-		if err := Pause(ctx, name); err != nil {
+		// Detach Pause from the caller's context. A canceled request (browser
+		// tab closed, network hiccup) mid-teardown would abort Pause inside
+		// its SIGTERM/SIGKILL poll loop and leave the workspace in
+		// StateStopping with the broker still alive — exactly the half-dead
+		// state this auto-pause exists to prevent. The bounded timeout still
+		// caps how long Pause can hang; it just doesn't ride on ctx.
+		pauseCtx, cancel := context.WithTimeout(context.Background(), pauseWallClockTimeout+15*time.Second)
+		err := Pause(pauseCtx, name)
+		cancel()
+		if err != nil {
 			return "", fmt.Errorf("workspaces: shred %q: pause running broker: %w", name, err)
 		}
 		// Re-read the registry so target reflects the post-pause state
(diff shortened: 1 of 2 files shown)


najmuzzaman-mohammad:
CodeRabbit dispositions:

**Finding 1 — `orchestrator_test.go` hard-coded BrokerPort 29984: SKIPPED.**
Port 29984 sits in the test-only high range (29980-29985) that the sibling Pause tests already use (`TestPauseHappyPathBrokerExitsCleanly` → 29980, `TestPauseFallsThroughWhenTokenMissingButProcessGone` → 29982). The convention is deliberate: the tests need the port *unbound* so `probePort` returns false on the first tick of Pause's poll loop. An ephemeral `net.Listen(":0")` then `Close()` introduces a TOCTOU race window where another process could grab the port between close and test execution — strictly worse than the current static-high-port pattern. If we later see CI flakes here, the fix is a shared helper across all five tests, not a one-off.

**Finding 2 — `orchestrator.go` Pause inherits request ctx: FIXED in c547df21.**
Valid catch. Replaced `Pause(ctx, name)` with a detached `context.WithTimeout(context.Background(), pauseWallClockTimeout+15*time.Second)` so a canceled request can't abort the SIGTERM/SIGKILL ladder mid-flight and leave the workspace in StateStopping. The 15s margin covers the post-loop tmux teardown + final liveness check.

gcampton:
Thanks Naj.

PR #1012, opened 2026-05-29
feat(inbox): add Needs action filter as default tab

najmuzzaman-mohammad:
## Summary

The Inbox previously defaulted to **All**, which mixed every actionable item without distinguishing what was waiting on the human right now from what was just lingering (rejected tasks, reviews bounced back to the author). Operators read this as "everything is mixed".

This adds a **Needs action** tab at position 0 and makes it the default. It is a strict subset of the existing actionability filter:

- **Requests**: every blocking request (already 1:1 with awaiting human).
- **Reviews**: `pending` / `in-review` only — `changes_requested` means the submitter holds the ball, so it's actionable but not on the human.
- **Tasks**: `decision` / `review` / `changes_requested` — `rejected` is a terminal state that needs no action.

The existing **All** tab keeps its current semantics so terminal + back-with-author rows remain one click away.

Filter order is now: `[Needs action] [All] [Unread] [Decisions] [Requests] [Reviews] [Rejected]`.

## Test plan
- [x] `bash scripts/test-web.sh web/src/components/lifecycle/DecisionInbox.test.tsx` — 7 pass, 2 pre-existing skips
- [x] `bunx tsc --noEmit`
- [x] New test covers default tab, rejected/changes_requested exclusion, click-to-All reveal
- [ ] Manual: open the Inbox, confirm Needs action is the default and that flipping to All surfaces rejected / changes_requested rows
- [ ] Screenshots: see published `screenshots/pr-<n>` branch

## Summary by CodeRabbit

* **New Features**
  * Introduced a "Needs action" Inbox filter that shows only items requiring immediate attention and is now the default view.
  * Filter counts updated to include the new "Needs action" bucket.

* **Tests**
  * Added unit tests validating default filter selection and visibility toggling between "Needs action" and "All".
  * Added an end-to-end screenshot script that captures the Inbox in the default "Needs action" state and after switching to "All".

### DIFF
--- a/web/e2e/screenshots/inbox-needs-action.mjs
+++ b/web/e2e/screenshots/inbox-needs-action.mjs
@@ -1,0 +1,171 @@
+// Capture the Inbox surface before vs. after the Needs action filter
+// becomes the default tab. Mocks /api/inbox/items with three pending
+// rows + two terminal/back-with-author rows (rejected task, review in
+// changes_requested) so the All vs. Needs action delta is visible.
+//
+// Run via:
+//   web/e2e/screenshots/publish.sh inbox-needs-action <pr-number>
+
+import process from "node:process";
+
+import {
+  DEFAULT_BASE,
+  flipStore,
+  installCommonMocks,
+  launchBrowser,
+  shotElement,
+} from "./lib.mjs";
+
+const OUT = process.env.WUPHF_SCREENSHOTS_OUT;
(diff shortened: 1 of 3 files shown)


najmuzzaman-mohammad:
## Screenshots

01-needs-action-default

02-all-shows-everything

FranDias (approved):

PR #1016, opened 2026-06-01
[release] fix(routines): exclude one-shot watchdog jobs from the Routines list

najmuzzaman-mohammad:
## What

The **Routines** surface was flooded with one-off "Follow up on …" and "Recheck task …" rows that no one created. Those are one-shot watchdog jobs the broker auto-enqueues on every task-lifecycle transition (`task_follow_up`, `request_follow_up`, `recheck`) — internal plumbing, not routines. `RoutinesApp` already *intended* to exclude them (its comment: *"One-shot follow-ups … would just be noise"*), but the classifier let every one through.

## Root cause — a wire-shape trap

`schedulerJob.IntervalMinutes` (Go) is serialized **without `omitempty`**, so every job hits the wire as `"interval_minutes": 0`, watchdogs included. `isCadenceSchedulerJob` gated on:

```ts
typeof job.interval_minutes === "number"   // typeof 0 === "number" → true for ALL jobs
```

So the "is this recurring?" check passed for everything, and each per-task watchdog rendered as a routine. The unit test missed it because its fixture **omitted** `interval_minutes` (`typeof undefined` is falsy), so it never exercised the real wire shape.

## Fix

A routine now requires a **positive** interval, a cron expression, or `system_managed`:

```ts
job.system_managed === true || hasPositiveInterval(job) || hasCronExpression(job)
```

One-shot watchdogs (interval `0`, no cron, not system-managed) are excluded entirely — they never appear, even under *Show system*, because they are transient plumbing, not crons.

Verified safe to filter all interval-0/no-cron/non-system jobs: the `ScheduleBuilder` composer can only emit recurring frequencies (interval / hourly / daily / weekdays / weekly / monthly), so a human can never create a legitimate one-shot routine that this would hide.

## Test plan

- `schedulerJobClassification.test.ts` — fixture now defaults `interval_minutes: 0` (mirrors the wire, so it actually guards the regression) + explicit follow-up / recheck cases. **green**
- `RoutinesApp.test.tsx` — added a one-shot follow-up to the fixture; asserts it never renders as a row, even with *Show system* on. **7 passed** (was 5)
- `SystemSchedulesPanel.test.tsx` (other classifier consumer) — **14 passed**, no collateral
- `bunx tsc --noEmit` clean · `bunx biome check` clean

## Notes

- `SystemSchedulesPanel.tsx` carries the same latent `typeof interval_minutes === "number"` pattern but is **dead code** (nothing mounts it since `CalendarApp` was deleted). This fix covers it for free; worth deleting separately.

## Screenshots

Routines list seeded with 2 recurring routines + 1 system cron + 4 one-shot watchdog jobs.

**Before** — four "One-shot / Unassigned" follow-up & recheck rows clutter the list (count: 6):

before

**After** — only the two real recurring routines remain (count: 2); watchdogs are filtered out:

after

## Summary by CodeRabbit

* **Bug Fixes**
  - One-shot system watchdog/maintenance jobs that previously appeared in the Routines list are now correctly filtered out and hidden from the UI.

* **Tests**
  - Expanded unit/regression tests and a new end-to-end screenshot spec to ensure these one-shot jobs remain excluded and the Routines view renders consistently.

### DIFF
--- a/web/e2e/screenshots/routines-filter.mjs
+++ b/web/e2e/screenshots/routines-filter.mjs
@@ -1,0 +1,180 @@
+// Capture the Routines list before/after the one-shot-watchdog filter fix.
+//
+// The broker enqueues one-shot task_follow_up / request_follow_up / recheck
+// jobs on every task-lifecycle transition. They ride the wire with
+// interval_minutes: 0, and the old classifier treated `typeof 0 === "number"`
+// as a cadence, so they flooded the Routines list. This spec seeds a realistic
+// /scheduler payload (2 real routines + 1 system cron + 4 one-shot watchdogs)
+// and shoots the Routines surface.
+//
+// Run order is controlled by the wrapper / driver, which sets the code state
+// (fixed vs reverted) and passes WUPHF_SHOT_LABEL so the before/after PNGs
+// land side by side in one orphan branch:
+//
+//   WUPHF_SHOT_LABEL=01-before  (classifier reverted to origin/main)
+//   WUPHF_SHOT_LABEL=02-after   (classifier fixed)
+
+import process from "node:process";
+
+import {
+  flipStore,
+  installCommonMocks,
+  launchBrowser,
+  shotElement,
+} from "./lib.mjs";
+
+const BASE = process.env.BASE_URL ?? "http://localhost:5273";
+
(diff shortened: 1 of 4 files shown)


FranDias (approved):

Issue #1004, opened 2026-05-28
Skills: held-out validation gate for enhance / new-skill proposals


najmuzzaman-mohammad:
## Why

Largest open gap from the [SkillOpt paper](https://arxiv.org/abs/2605.23904) (arXiv 2605.23904) — and the highest-leverage follow-up to PR #998.

SkillOpt's central finding: a self-editing skill loop is only safe if every accepted edit **strictly improves a held-out validation score**, with ties rejected. Removing this gate is what separates "self-improving" from "self-drifting." Their best skills land with 1–4 accepted edits total. PR #998 added the static gates (depth, repetition, heading shape, edit-count bound, invariant block). The dynamic gate is the next layer.

## What

For every NEW or ENHANCE proposal, before accepting:

1. Resolve a per-workspace fixture set of 5–10 representative tasks the skill is supposed to help with.
2. Run the agent loop with the candidate skill vs the current skill (or no skill, for new proposals) against the fixture set.
3. Accept only when the candidate strictly improves aggregate task success. Ties rejected.

## What it is NOT

- It is not a stub that always passes. A no-op gate is worse than no gate — false confidence in a self-editing loop.
- It is not a generic auto-grader. Our domain (sales, ops, open-ended judgment) needs a hybrid:
  - (a) task-outcome heuristics where we have them (e.g. issue resolved with no rework),
  - (b) LLM-as-judge with a strict rubric for everything else,
  - (c) human-in-the-loop for the most ambiguous cases.

## Open design questions

- Where do fixture tasks live? Per-workspace under \`team/skills/.fixtures/<slug>.md\`?
- Who authors them — a one-time CEO-chat onboarding step, or grown from real Issue/Task outcomes?
- Cost model: synthesis is offline, but running 5–10 fixture tasks per candidate is not free. Cache results keyed on (skill SHA, fixture SHA)?
- What does "strict improvement" mean when fixtures are LLM-judged and the judge has variance? Sample N times and require majority? Use a paired statistical test?

## Acceptance criteria

- A workspace can declare fixture tasks for any skill slug.
- The synthesizer runs the candidate against fixtures and gates accept/reject on strict improvement.
- The gate is wired through both ENHANCE and RENAME paths.
- Failures surface in the SynthError slice with the per-fixture comparison.

## References

- SkillOpt: arXiv 2605.23904 §3 — validation gate is "the only thing that matters in a self-editing loop"
- PR #998 — the static gates this dynamic gate builds on
[...]

PR #1025, opened 2026-06-03
feat(skill): held-out validation gate for new/enhance proposals (#1004)

abeeb-rai:
## Summary

Implements the dynamic quality gate for Stage B skill synthesis described in issue #1004.

Before a `NEW` or `ENHANCE` proposal is written to the wiki, the gate compares the candidate body against a baseline (empty for new skills, existing body for enhance) on representative task fixtures using an LLM-as-judge. Acceptance requires **zero regressions on any fixture and at least one win**. The gate gracefully skips when no fixture file exists for the slug.

### Design decisions

- **No fixture → skip, not reject**: `(false, nil)` increments `ValidationGateNoFixtures` so the metric is observable without blocking early adoption.
- **All-judge-failure → reject**: When every HTTP call fails, all fixtures degrade to `equivalent` → 0 wins, 0 losses → "no improvement" rejection. Conservative by design.
- **Mutex: acquire-read-release before gate call**: Gate's LLM call can be slow; the mutex is held only for the skill lookup, not the full gate execution.
- **`slog.Info` not `Warn` on gate rejection**: A rejection is an expected outcome, not an anomaly.

### Known gaps / follow-ups

- No tooling to prompt agents to author fixture tasks when a new skill is promoted (fixture authorship trigger).
- No-API-key path silently rejects all proposals after the first `sync.Once` warning — should surface this more clearly.
- `enhance` fall-through: if the enhance target disappears between synthesis and gate execution, the gate runs with `baselineBody=""` (treated as new-skill comparison).

### DIFF
--- a/internal/team/skill_compile.go
+++ b/internal/team/skill_compile.go
@@ -91,6 +91,12 @@ type SkillCompileMetrics struct {
 	// SkillEnhancementsTotal counts proposals that enhanced an existing
 	// skill instead of being discarded or created as new.
 	SkillEnhancementsTotal int64
+	// ValidationGateRejections counts Stage B proposals rejected by the
+	// held-out fixture validation gate (regression or no improvement found).
+	ValidationGateRejections int64
+	// ValidationGateNoFixtures counts Stage B proposals where the gate was
+	// called but no fixture file existed (gate skipped — graceful degradation).
+	ValidationGateNoFixtures int64
 }
 
 // snapshotSkillCompileMetrics returns a copy of m suitable for serialization.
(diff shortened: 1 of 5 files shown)


FranDias:
Closing — author (abeeb-rai) has left. Triangulation found a security **BLOCK**: the held-out validation gate was **bypassed for enhance/rename proposals** (`skill_validation_gate.go:71/215/234`), defeating its purpose; plus real lint/commitlint CI failures, judge-prompt injection with parse-failures masking regressions, transient judge failures cached as 'equivalent', and unbounded validation cost. As a large (+2255) feature with no remaining owner, closing as orphaned rather than carrying the security holes.

PR #1042, opened 2026-06-08
perf(broker): add channelIndex map for O(1) channel lookups

### LINKED ISSUE #1040: perf: findChannelLocked is O(N) — add channelIndex map for O(1) lookup
## Problem

`findChannelLocked` in `broker.go:1174-1182` does a linear scan over `b.channels` on every call:

```go
func (b *Broker) findChannelLocked(slug string) *teamChannel {
    slug = normalizeChannelSlug(slug)
    for i := range b.channels {           // O(N) every time
        if b.channels[i].Slug == slug {
            return &b.channels[i]
        }
    }
    return nil
}
```

This function is called **51 times across 21 files** — it's the hottest lookup path in the broker. There is no `channelIndex` map (confirmed via grep), unlike `findMemberLocked` which already uses a `memberIndex map[string]int`.

## Benchmark evidence

A side-by-side benchmark (`BenchmarkFindChannelLocked` vs `BenchmarkFindMemberLocked`) confirms the issue:

### findChannelLocked — linear scaling (O(N))

| Channels | ns/op |
|----------|-------|
| 10 | 90 |
| 50 | 184 |
| 100 | 320 |
| 500 | 1,157 |
| 1000 | 4,596 |

**~51x slower** at 1000 channels vs 10.

### findMemberLocked — flat (O(1) via map)

| Members | ns/op |
|---------|-------|
| 10 | 97 |
| 50 | 93 |
| 100 | 93 |
| 500 | 80 |
| 1000 | 72 |

Flat regardless of size.

## Proposed fix (~20 LOC)

1. Add `channelIndex map[string]int` to the `Broker` struct (mirrors `memberIndex` at line 66).
2. Add `rebuildChannelIndexLocked()` (mirrors `rebuildMemberIndexLocked` at line 1237).
3. Update `findChannelLocked` to use map lookup with lazy rebuild (same pattern as `findMemberLocked`).
4. Rebuild index on channel add/remove mutations.

## Impact

O(N) → O(1) on every channel operation. With 50 channels × 51 call sites, this eliminates thousands of string comparisons per request cycle. Directly improves response latency for every API endpoint that touches channels (messages, DMs, agent routing, etc.).

abeeb-rai:
## Summary

- Add `channelIndex map[string]int` to the Broker struct, mirroring the existing `memberIndex` pattern
- `findChannelLocked` now uses O(1) map lookup with lazy rebuild (same pattern as `findMemberLocked`)
- Channel removal in `handleChannels` also uses the index instead of its own linear scan
- Includes benchmark tests comparing `findChannelLocked` vs `findMemberLocked`

## Problem

`findChannelLocked` (`broker.go:1174`) does a linear scan over `b.channels` on every call — **51 call sites across 21 files**. This is the hottest lookup path in the broker.

### Before (O(N) linear scan)
| Channels | ns/op |
|----------|-------|
| 10 | 90 |
| 100 | 320 |
| 1000 | **4,596** |

### After (O(1) map lookup)
| Channels | ns/op |
|----------|-------|
| 10 | 67 |
| 100 | 85 |
| 1000 | **75** |

**61x faster at 1000 channels.** Now flat across all sizes, matching `findMemberLocked`.

Closes #1040

## Test plan

- [x] All channel-related tests pass (`Channel|DM|channel` — 30+ tests)
- [x] Benchmark confirms O(1) scaling
- [x] Pre-commit hooks pass (gofmt, golangci-lint, go-vet, secrets)

## Summary by CodeRabbit

* **Chores**
  * Optimized channel lookup performance for improved responsiveness when managing channels.

* **Tests**
  * Added benchmark tests to measure performance characteristics of channel and member lookup operations.

### DIFF
--- a/internal/team/broker.go
+++ b/internal/team/broker.go
@@ -67,6 +67,7 @@ type Broker struct {
 	memberPresence    map[string]memberPresenceRecord // slug → presence; guarded by mu, populated via brokerTransportHost
 	presenceKeyToSlug map[string]string               // "adapter:key" → slug; guarded by mu
 	channels          []teamChannel
+	channelIndex      map[string]int // slug → index into channels; guarded by mu
 	sessionMode       string
 	oneOnOneAgent     string
 	focusMode         bool
@@ -1173,10 +1174,11 @@ func (b *Broker) FocusModeEnabled() bool {
 
 func (b *Broker) findChannelLocked(slug string) *teamChannel {
 	slug = normalizeChannelSlug(slug)
-	for i := range b.channels {
-		if b.channels[i].Slug == slug {
-			return &b.channels[i]
-		}
+	if len(b.channelIndex) != len(b.channels) {
+		b.rebuildChannelIndexLocked()
+	}
+	if i, ok := b.channelIndex[slug]; ok && i < len(b.channels) && b.channels[i].Slug == slug {
+		return &b.channels[i]
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


FranDias (approved):

FranDias:
Closing — author (abeeb-rai) has left. Note for the record: although this was **approved**, triangulation found a real correctness bug — `channelIndex` goes **stale on same-length channel replacement** (and the new map changed duplicate-slug semantics from first-match to last-match), so lookups could falsely return not-found. Closing as orphaned; a fresh O(1)-channel-lookup change should rebuild/update the index on every channel mutation path.

PR #1044, opened 2026-06-08
fix(team): use atomic writes in playbook_commit.go

### LINKED ISSUE #1043: fix: use atomic writes in playbook_commit.go
## Summary

`CommitPlaybookSkill` and `CommitPlaybookExecution` in `internal/team/playbook_commit.go` (lines 54 and 110) use `os.WriteFile`, which truncates-then-writes. A process crash between truncation and write completion could leave a zero-byte or partially-written `SKILL.md` / `.executions.jsonl` file.

The same package already has a `writeFileAtomic` helper in `learning_commit.go` (line 108) that uses the standard temp-file + `os.Rename` pattern. Both call sites should use it for consistency and correctness.

## Impact

Low probability but easy to fix. The mutex and immediate git-commit mitigate most real-world risk, but using the atomic helper aligns with the rest of the package and eliminates the truncation window entirely.

## Fix

Replace `os.WriteFile` with `writeFileAtomic` at lines 54 and 110 of `internal/team/playbook_commit.go`.

abeeb-rai:
## Problem

`CommitPlaybookSkill` and `CommitPlaybookExecution` in `internal/team/playbook_commit.go` use `os.WriteFile` to write `SKILL.md` and `.executions.jsonl` files. `os.WriteFile` truncates the file to zero bytes *before* writing new content. If the process crashes between truncation and write completion, the file is left at zero bytes or partially written.

Playbooks are how agents learn from past work — `SKILL.md` is the compiled skill definition and `.executions.jsonl` is the execution history. Silent corruption here degrades agent quality over time and is hard to detect because the file exists but contains garbage or nothing.

## Solution

Replace both `os.WriteFile` calls with `writeFileAtomic`, a helper already in the same package (`learning_commit.go:108`). It writes to a temp file in the same directory, then does an `os.Rename` — which is atomic on POSIX filesystems. The file is either fully written or untouched; there is no partial-write window.

No new code introduced — just reusing an existing internal helper for consistency.

Closes #1043

## Test plan

- [x] All 39 Go packages pass (`internal/team` included)
- [x] Pre-commit hooks (gofmt, go vet, golangci-lint, secretlint) pass
- [x] Pre-push hooks (build, go-unit, smoke) pass

### DIFF
--- a/internal/team/playbook_commit.go
+++ b/internal/team/playbook_commit.go
@@ -51,7 +51,7 @@ func (r *Repo) CommitPlaybookSkill(ctx context.Context, slug, relPath, content, 
 	if err := os.MkdirAll(filepath.Dir(fullPath), 0o700); err != nil {
 		return "", 0, fmt.Errorf("playbook commit: mkdir: %w", err)
 	}
-	if err := os.WriteFile(fullPath, []byte(content), 0o600); err != nil {
+	if err := writeFileAtomic(fullPath, []byte(content), 0o600); err != nil {
 		return "", 0, fmt.Errorf("playbook commit: write: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


FranDias:
Closing — author (abeeb-rai) has left. Triangulation rated this the cleanest (MEDIUM only): temp+rename gives reader-atomicity but not crash durability (no fsync of file+dir). The atomic-write correctness improvement is genuinely useful — if we want it, a fresh PR should add `fsync` of the temp file + parent dir (and apply the same to `learning_commit.go`).

PR #1057, opened 2026-06-09
feat(tasks): CEO decomposes-and-delegates; new agents need human approval (Phase 3)

najmuzzaman-mohammad:
## What

Phase 3 of the multi-agent overhaul. Addresses the report that **the CEO doesn't break work down, spin off sub-agents, or create specialists** — and adds the rule that **the CEO must always get human approval before spinning up any new agent.**

The re-audit (like Phase 2) found the machinery already exists: sub-tasks via `parent_issue_id`, `team_member` create, `team_plan`, `task_created`→owner auto-wake, and Phase 2 concurrent lanes. The gap was **behavioral** (the CEO was told to do the work itself) plus **one missing safety gate**. So this is backend + prompt only.

### 1. Decompose-and-delegate (behavioral)
`BuildTaskExecutionPacket` now gives the **lead** an explicit branch when it executes an owned task:
- break large or cross-functional work into sub-tasks (`team_task action=create`, `parent_issue_id` = this task),
- assign each to the best **existing** specialist (reuse-first),
- let them spin off and run concurrently on their own lanes (rides the existing `task_created`→`deliverTaskNotification` auto-wake + Phase 2 lanes),
- aggregate at the parent, and don't close it while children are open.

A specialist never sees this block — it still gets the direct-implementation framing. Previously the execution packet pushed *solo direct-implementation* for everyone, which is right for a specialist in a worktree but wrong for the coordinator.

### 2. New-agent creation gated behind mandatory human approval (structural)
Spinning up a new agent is durable and persistent, so it **always** requires explicit human approval — implemented structurally so it can't be skipped by prompt drift:
- New `requireTeamMemberApproval` (`member_approval.go`) mirrors the proven external-action approval gate: raises a blocking `kind:"approval"` request and polls `/interview/answer` until the human decides.
- `handleTeamMember` create calls it **first** and only creates the member on approve; on reject/timeout it returns guidance to **reuse an existing specialist**.
- `WUPHF_UNSAFE=1` is the only bypass (matches the action gate / `--unsafe`). Human-initiated creation (UI → `/office-members`) doesn't reach this path, so only agent-initiated creates are gated.
- The approval surfaces via the **existing** approval card (same `kind:"approval"` request the external-action gate and `team_request` already render) — no new web.

### 3. Prompt reconcile
`team_member` tool description + lead rules 16f/17/18 reworded to "reuse an existing specialist first; a new agent needs explicit human approval."

## Decisions (from the steering on this lane)
- **Robust architecture** over a prompt-only nudge → the gate is structural (in the tool path), and spin-off is the deterministic `task_created` auto-wake, not best-effort prompting.
- **Capability-gap, reuse-first** → decomposition assigns existing specialists; new agents only when no teammate fits.
- **Always human approval before any new agent** → hard blocking gate.

## Out of scope (deliberate)
No separate broker rollup/aggregation logic: a child's completion broadcast already wakes the CEO via the existing message loop, and the new prompt tells it to aggregate at the parent. Parent/child linkage + `SubTasksList`/`ParentTaskBreadcrumb` already render.

## Tests
- Member gate: bypass (`WUPHF_UNSAFE`), approve, and reject paths (httptest mock broker).
- Decompose branch: the lead's execution packet contains the decompose-and-delegate block; a specialist's does not.
- Full `teammcp` + `team` suites green (rebased onto current `main`, which now includes #1049). `gofmt`/`vet` clean.

Backend + prompt only — no web surfaces, so no screenshots.

## Summary by CodeRabbit

* **New Features**
  * Require explicit human approval before creating new office-wide team members; declined requests let work be reassigned to existing specialists.
  * Updated lead guidance and prompts: leads act as coordinators who decompose larger tasks into owned sub-tasks, prefer reusing existing specialists, and keep parent tasks as umbrellas until children complete.

* **Tests**
  * Added tests covering the human-approval flow and lead task decomposition behavior.


FranDias (dismissed):

FranDias (approved):

najmuzzaman-mohammad:
Addressed the 3 CodeRabbit review findings in `internal/teammcp/` (commit 3b5ae996):

1. **Advertised options whose semantics were discarded** (`member_approval.go`): the gate used the stock `approval` option set, which includes `approve_with_note`, `reject_with_steer`, and `hold` — but the create path only has a binary nil/error outcome and reads no typed guidance, so `approve_with_note`'s note was silently dropped. Now passes an explicit minimal `[]HumanInterviewOption{{approve}, {reject}}` to `normalizeHumanRequestOptions` so it stops auto-injecting the steering options (labels/descriptions still get enriched by ID from the approval defaults). The card no longer offers an outcome the handler ignores. Did NOT plumb the note into `handleTeamMember` (out of scope).

2. **Unstoppable timeout** (`member_approval.go`): replaced `time.After(actionApprovalTimeout)` with `time.NewTimer(...)` + `defer timeout.Stop()` and switched the select case to `case <-timeout.C:`, so the approve/reject early-return path no longer leaks a live timer per call.

3. **Data race on `sawRequest`** (`member_approval_test.go`): the value was written in the httptest handler goroutine and read in the test goroutine. Switched to `sync/atomic.Bool` (`Store`/`Load`).

Validated: `gofmt -l` clean, `go build`, `go vet`, and `go test ./internal/teammcp/ -race -count=1` all pass (incl. the proceed + reject member-approval paths under -race).

PR #1056, opened 2026-06-09
feat(team): wire context-packer to the broker (BrainHandle adapter) [PR-B]

najmuzzaman-mohammad:
**Stacked on #1051** (`feat/slack-context-packer`). PR-B of the Slack agent-coordination series.

Implements the packer's brain/validator/sink seams over the live broker, so the egress core (#1051) can run against real office data.

## What
- **`packerBrain`** (`packer.BrainHandle`): `PlanStep` renders the human-approved `IssueDraftSpec` — **never** the raw free-form `Details` (which the egress policy denies); `TaskLearnings` returns only **trusted**, exact-task-scoped learnings (uses the new `LearningSearchFilters.TaskID`); `TaskWikiRefs` resolves the task's explicitly-linked `WikiRefs` to article bodies (never a free wiki search); `Roster` lists owner + reviewers.
- **`packerSnapshotValidator`**: re-checks task freshness (`UpdatedAt`) at delivery time.
- **`PackerInjectionSink`**: in-memory append-only egress audit with idempotency lookup + transition history.

OSS self-hosted seam; Nex cloud provides a per-tenant adapter over the same interfaces.

## Test plan
- [x] `go test ./internal/team/ -run TestPacker` — 8 tests green (plan-step-not-details, trusted+scoped learnings, linked-wiki resolution, roster, validator fresh/stale/missing, sink).
- [x] `gofmt` / `go vet` / `golangci-lint` (diff) — 0 issues; `go build ./...` OK.

## Not in this PR
The Slack bridge transport (PR-C) calls `Pack`/`Deliver` with these seams; the wiki-link surface (PR-E) populates `WikiRefs`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1051, opened 2026-06-09
feat(packer): Slack inbound context-packer — egress core + prerequisites

najmuzzaman-mohammad:
## What

First slice of the **Slack inbound context-packer** — the CEO-membrane component
that lets agents-from-anywhere compound on the office brain inside a company
Slack. This PR lands the **spec**, the **three code prerequisites** the review
proved were missing, and the **packer's egress core** (the security boundary).
It does **not** include the Slack bridge adapter or Block Kit gate cards — those
are a separate spec/PR (the packer's `BrainHandle`/`SlackBridge`/`SnapshotValidator`/
`InjectionSink` are interfaces; the team-side adapters land with the bridge).

Design context: `docs/specs/slack-context-packer.md`. OSS self-hosted scope; the
multi-tenant host is Nex cloud, which shares this same kernel.

## Why this shape

The packer ships internal office-brain content into a foreign LLM we do not
control. That makes the **egress boundary** (`Classify` + `EgressPolicy` +
redaction) the one place a miss leaks customer data. The design was reviewed by
codex (16 findings), then by a triangulation pass (security/API/architecture)
plus an adversarial verification agent — which caught that the spec's "redaction
failure denies the item" had **no backing** in the real scanner. The three
prerequisites here close exactly the gaps that review surfaced.

## Commits

1. `docs(slack-packer)` — the spec (v3, triangulation-hardened, with both review
   disposition tables).
2. `feat(scanner)` — **Prerequisite 1**: `RedactForEgress`, a fail-closed wrapper
   over the display redactor. Denies (emits no content) when the scan is
   `Poisoned` or hit the entropy cap; pattern redaction is exhaustive, so a high
   pattern count alone is not a deny signal.
3. `feat(learnings)` — **Prerequisite 2**: an exact `TaskID` filter on
   `LearningSearchFilters`. Previously only fuzzy scope/file/query existed;
   task-scoped learning egress requires an exact AND-scope so a foreign bot never
   gets unrelated durable learnings. Empty `TaskID` preserves existing behavior.
4. `feat(tasks)` — **Prerequisite 3**: `WikiRefs []string` on `teamTask` (the
   task→article backing link), wired through both custom JSON marshalers and
   `TaskPostRequest`, applied on create + mutation via `dedupePaths`. Without it,
   "explicitly task-linked wiki refs" would have forced the implementer onto the
   forbidden free-search path. Empty by default ⇒ "no wiki egress".
[...]

PR #1061, opened 2026-06-09
feat(slack): real Slack bridge transport (Socket Mode) [PR-C]

najmuzzaman-mohammad:
**Stacked on #1056 → #1051.** PR-C of the Slack agent-coordination series — the real Slack bridge that makes the packer end-to-end.

## What
A real Slack transport (`internal/team/slack_transport.go`) modeled on `telegram.go`, plus the packer's egress delivery (`slack_bridge.go`):

- **Inbound: Socket Mode** (`xapp-` app token) — no public URL / webhook needed, ideal for a self-hosted OSS broker. EventsAPI message events route to the office via the `Host` contract (`UpsertParticipant` then `ReceiveMessage`). Skips subtyped events, bot/self messages (echo-loop guard via `auth.test`), empty text, and unmapped channels. Events are Ack'd before processing.
- **Outbound: Web API** `chat.postMessage` via the `xoxb-` bot token. `FormatOutbound` is a pure slug→channel mapping; `Send` threads on `ThreadKey`.
- **`packer.SlackBridge.Post`** (`slack_bridge.go`): posts the egress-classified mention to the delegation's channel/thread and the thread-context block as a reply, returning the mention `ts` as the audit anchor. **Posts with `escapeText=true`** — the egress boundary, so injected Slack control sequences (`<!channel>`/`<!here>`, fake `<url|label>` links) in a tainted Ask are neutralized.
- **Seams for testing:** a narrow `slackAPI` interface (4 Web API calls) + a `socketRunner` seam, so the transport, the bridge, and the routing logic are unit-tested against a fake with **no network**.
- **Config:** `ResolveSlackBotToken()` / `ResolveSlackAppToken()` (env → config file, mirroring the Telegram resolver). **Registration:** a Slack block in `launcher_transports.go` (starts only when both tokens + ≥1 connected slack channel exist; skips silently otherwise).

## Test plan
- [x] `go test ./internal/team/ -run Slack -race` — green. Inbound mapping + upsert-before-receive ordering, all skip cases, user resolution (human/bot/fallback), `Send`/`FormatOutbound` (+unmapped, +API error), `SlackBridge.Post` (mention-only, threaded reply, existing-thread anchoring, guards, API errors, follow-up-error-still-returns-ts), and control-sequence escaping.
- [x] full `internal/team` suite green; `gofmt` / `go vet` / `golangci-lint` (diff) — 0 issues; `go build ./...` OK.
- [x] Adversarial verification agent on the Slack surface — dispositions below.

## Verification dispositions
Built by a subagent, then independently re-verified (build/vet/lint/`-race`) and read end-to-end by me. An adversarial verification agent on the Slack surface found 6 issues; **all fixed** in the `fix(slack)` hardening commit, each with a test.

| # | Finding | Sev | Disposition |
|---|---------|-----|-------------|
| F1 | Office outbound (`Send`/`formatSlackOutbound`) could inject Slack control sequences | HIGH | FIXED — escape every dynamic field via `slackutilsx.EscapeMessage`, keep own markup literal |
| F2 | `auth.test` failure left the self-echo guard off | LOW | FIXED — retry with backoff; `bot_id`/subtype remain primary; regression test for `botUserID==""` |
| F3 | Events Ack'd before processing → a transient Host failure silently dropped the message | MEDIUM | FIXED — Ack **after** success; failed Host write leaves it un-Ack'd so Slack redelivers (broker dedupes by `ExternalID`) |
| F4 | `Run` returned on cancel without waiting for the socket goroutine | MEDIUM | FIXED — cancel + bounded 5s wait so the Host isn't used after `Run` returns |
| F5 | Failed thread follow-up returned an error → Deliver marks FAILED → retry duplicates the mention | MEDIUM | FIXED — follow-up is best-effort: return `(mentionTS, nil)` + log |
| F6 | `UserMap` cached name only → a warmed bot could later be attributed human | LOW | FIXED — cache `{name, human}` |

I also caught + fixed (before the agent ran) the egress bridge posting with `escapeText=false` (commit `26c9c8ce`). Controls the agent confirmed held: no token interpolation in logs/errors, mutexed map access, `UpsertParticipant` before `ReceiveMessage`.

## Live run
Needs `SLACK_BOT_TOKEN` (`xoxb-`) + `SLACK_APP_TOKEN` (`xapp-`, Socket Mode) + a workspace/app. Everything above is tested against a mock Slack; the live run happens in the merged end-to-end test.

## Not in this PR
The Slack Block Kit approval gate cards (PR-D, stacked on the deterministic-integrations lane) use this transport's interactivity surface.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/jackc/pgx/v5 v5.9.2
 	github.com/modelcontextprotocol/go-sdk v1.6.0
 	github.com/sahilm/fuzzy v0.1.2
+	github.com/slack-go/slack v0.25.0
 	go.uber.org/goleak v1.3.0
 	golang.org/x/net v0.55.0
 	golang.org/x/sync v0.20.0
--- a/go.sum
+++ b/go.sum
@@ -101,6 +101,8 @@ github.com/erikgeiser/coninput v0.0.0-20211004153227-1c3628e74d0f h1:Y/CXytFA4m6
 github.com/erikgeiser/coninput v0.0.0-20211004153227-1c3628e74d0f/go.mod h1:vw97MGsxSvLiUE2X8qFplwetxpGLQrlU1Q9AUEIzCaM=
 github.com/fsnotify/fsnotify v1.10.1 h1:b0/UzAf9yR5rhf3RPm9gf3ehBPpf0oZKIjtpKrx59Ho=
 github.com/fsnotify/fsnotify v1.10.1/go.mod h1:TLheqan6HD6GBK6PrDWyDPBaEV8LspOxvPSjC+bVfgo=
+github.com/go-test/deep v1.1.1 h1:0r/53hagsehfO4bzD2Pgr/+RgHqhmf+k1Bpse2cTu1U=
+github.com/go-test/deep v1.1.1/go.mod h1:5C2ZWiW0ErCdrYzpqxLbTX7MG14M9iiw8DgHncVwcsE=
 github.com/golang-jwt/jwt/v5 v5.3.1 h1:kYf81DTWFe7t+1VvL7eS+jKFVWaUnK9cB1qbwn63YCY=
 github.com/golang-jwt/jwt/v5 v5.3.1/go.mod h1:fxCRLWMO43lRc8nhHWY6LGqRcf+1gQWArsqaEUEa5bE=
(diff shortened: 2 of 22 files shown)


najmuzzaman-mohammad:
Superseded by #1078, which folds this work in on top of the main (#1062) merge as a single integration PR. Closing to consolidate the Slack agent-coordination lane.

PR #1082, opened 2026-06-14
fix(web): scope the interview bar to the chat where the request originated

najmuzzaman-mohammad:
## Problem

The human-interview bar (an agent's clarifying question / external-action approval above the composer) showed up **globally**. It was mounted once in `Shell` and read the office-wide request queue (`useRequests`, `scope=all`), so a question raised in one task channel surfaced on **every** surface — the home composer, other channels, the wiki — blocking the human where the request had nothing to do with them.

## Fix

Scope the bar to the chat the human is currently viewing.

- **`InterviewBar`** now takes a required `channelSlug: string | null` prop and filters the queue to requests whose channel matches. Empty/absent request channel ⇒ `#general`, mirroring the broker's `normalizeChannelSlug` default, so legacy channel-less requests don't vanish. `null` (a non-chat surface) shows nothing.
- **`useActiveChannelSlug()`** (new, in `useCurrentRoute.ts`) resolves the active chat's channel:
  - channel route ⇒ its slug
  - task-detail route ⇒ the owning task's channel (via the already-cached `useOfficeTasks`)
  - home ⇒ `#general` (the home composer posts there)
  - everything else (wiki, agents, settings, inbox, …) ⇒ `null`
- **`Shell`**, **`DMView`**, and **`OnboardingChat`** pass their channel explicitly.

Cross-channel triage still lives in the **Inbox**, so narrowing the inline bar never strands a request.

## Why a prop instead of reading the route inside `InterviewBar`

`DMView` is reused across surfaces where `useChannelSlug()` would not reflect the DM's channel, so the channel is passed in explicitly. This also keeps `InterviewBar` a pure, easily-tested presentational component with no hidden routing dependency.

## Tests

- Added `<InterviewBar> channel scoping` tests: only same-channel requests render, channel-less ⇒ `#general`, `null` surface ⇒ nothing. Verified **red** before the filter (showed `1/2` instead of `1/1`).
- Existing `InterviewBar` tests updated to pass the new prop.
- Full web suite green: **2152 passed, 14 skipped, 0 failed**. `tsc --noEmit` clean. `bun run build` clean.

## Screenshots

The visible delta is the bar's *presence/absence per channel*, which needs a live multi-channel scenario with a pending interview to stage. Capturing via the screenshot harness is a follow-up; the behavior is covered by the unit tests above.

## Summary by CodeRabbit

* **New Features**
  * Interview request pending queues are now scoped to the currently active conversation, so only items from the viewed channel appear. Channel-less requests default to **general**; non-chat surfaces suppress the pending queue while keeping CEO content visible.
  * The DM drawer and onboarding chat now pass the correct channel context to the Interview bar.

* **Tests**
  * Updated Interview bar tests for explicit channel scoping and added coverage for null behavior.
  * Added unit tests for the inbox notification sound using Web Audio fakes.

* **Refactor**
  * Updated the inbox chime to a new synthesized bell-like sound and improved handling when audio starts suspended.

### DIFF
--- a/web/src/components/layout/Shell.tsx
+++ b/web/src/components/layout/Shell.tsx
@@ -1,5 +1,6 @@
 import type { ReactNode } from "react";
 
+import { useActiveChannelSlug } from "../../routes/useCurrentRoute";
 import { AgentPanel } from "../agents/AgentPanel";
 import { CommandPaletteHost } from "../command/CommandPalette";
 import { TeamMemberWelcome } from "../join/TeamMemberWelcome";
@@ -24,6 +25,7 @@ export function Shell({ children }: ShellProps) {
   // — both rails are flex children of `.office`. The rail is 56px wide
   // and the channel sidebar keeps its own width; the layout reflows
   // automatically.
+  const activeChannelSlug = useActiveChannelSlug();
   return (
     <div className="office">
       <WorkspaceRail />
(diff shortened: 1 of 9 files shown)


FranDias (dismissed):

najmuzzaman-mohammad:
### Follow-up in this PR: louder, reliable inbox chime

Per the decision that **blocking approvals also stay in their origin chat** (not surfaced globally), the inbox notification chime is now the main office-wide "needs you" signal, so I hardened it:

- **Louder + richer** — `web/src/lib/notificationSound.ts` now plays a bright three-note A-major arpeggio (A5·C#6·E6) with soft octave overtones. Peak amplitude went from `0.05` → ~`0.4` at the destination (≈7× louder, still clip-safe).
- **Fixed a dropped first ding** — a suspended `AudioContext` (autoplay policy / backgrounded tab) is now resumed and the chime emitted in the resolved callback, instead of bailing on a synchronous `state` check right after the async `resume()`. The old code swallowed the first ding after the tab regained focus.
- **Tests** — added `notificationSound.test.ts` (loudness floor, resume-then-play, no-Web-Audio no-op). The resume-then-play test was verified **failing** against the old logic.

Trigger is unchanged: `InboxButton` plays it when the broker's `inbox_attention` count rises (`/office/stats`, 10s poll). Worth a listen to confirm the loudness/timbre feel right — easy to tune `CHIME_PEAK` or the notes if you want it punchier or softer.

FranDias (approved):

PR #1094, opened 2026-06-14
fix(tasks): remove channel selection from the new-task modal

najmuzzaman-mohammad:
## What

The task creation modal (`TaskCreateDialog`, opened from the **+ New task** button on the tasks page) still exposed a `#channel` picker chip. Channels are no longer a user-facing concept — every task gets its own channel — so this control was stale and confusing.

This PR removes the channel selection:

- Drop the `#channel` PropertyChip, the eyebrow `· #channel` segment, the `defaultChannel` prop, and all channel state from `TaskCreateDialog`.
- `useCreateTask` now treats `channel` as **optional**; the broker still defaults it to `"general"` under the hood, so the wire shape is unchanged.
- Update the Storybook stories (drop the channel sample data / `PreselectedChannel` / `NoChannelsYet` stories).

## Tests

- New `TaskCreateDialog.test.tsx` regression guard:
  - asserts the modal renders **no** channel selector
  - asserts submit no longer sends a user-chosen `channel`
  - both assertions **fail** against the pre-fix modal and pass with the fix
- Full web suite green: **234 files / 2172 passed, 14 skipped**.
- `tsc --noEmit` clean, `biome check` clean.

## Screenshots

⏳ Pending — needs a running broker + web instance to capture the modal before/after. Will add via `web/e2e/screenshots/publish.sh` before marking ready.

## Summary by CodeRabbit

* **Refactor**
  * Task creation dialog simplified: Channel selection removed; tasks now default to the general channel.

* **Tests**
  * Added test coverage for the updated task creation flow.

### DIFF
--- a/web/src/components/tasks/TaskCreateDialog.stories.tsx
+++ b/web/src/components/tasks/TaskCreateDialog.stories.tsx
@@ -2,16 +2,9 @@ import { useEffect, useState } from "react";
 import type { Meta, StoryObj } from "@storybook/react-vite";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 
-import type { Channel, OfficeMember } from "../../api/client";
+import type { OfficeMember } from "../../api/client";
 import { Button } from "../ui/Button";
 import { TaskCreateDialog } from "./TaskCreateDialog";
-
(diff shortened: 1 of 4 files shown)


FranDias (approved):

PR #1107, opened 2026-06-16
fix(tasks): new Issues created inside a task chat get their own channel

najmuzzaman-mohammad:
## Problem (recurrence)

Reported: "latest release is still creating tasks but those tasks still share the same chat." Confirmed against v0.227.0 (which includes the earlier sub-task fix) by inspecting the live workspace:

```
OFFICE-22  parent=-  channel=task-office-22   "Daily Digest from email"
OFFICE-28  parent=-  channel=task-office-22   "Outbound email reply agent"   ← shares OFFICE-22's chat
```

Both are **top-level Issues with no `parent_issue_id`**, so the previous sub-task fix never applied. `OFFICE-28` was created while the CEO was talking inside `OFFICE-22`'s chat, so it inherited `task-office-22`.

## Root cause

`shouldMintPerTaskChannel` only minted a dedicated channel when the incoming channel was `general`. A new top-level Issue created from inside another Issue's `task-<id>` channel hit the "non-general → don't mint" guard and silently shared that Issue's timeline.

## Fix

Mint a dedicated channel for a real top-level task when the incoming channel is `general` **OR** is another task's per-task channel — detected via the channel's linked `TaskID` (`channelOwnedByAnotherTaskLocked`), not a slug-prefix heuristic. Genuinely explicit, non-per-task shared channels (project / bridged) are still respected. Threaded through all four create paths (mutation, plan ×2, lifecycle).

## Tests

- **Live HTTP regression** `TestBrokerSecondTopLevelTaskFromTaskChannelMintsOwnChannel`: a second top-level Issue created from the first Issue's channel must mint its own `task-<id>`. Verified to **fail pre-fix** (shares the first's channel) and **pass post-fix**.
- `TestShouldMintPerTaskChannelGuards` extended for the new signature + cases ("top-level task created in another task's channel mints its own", "system task in another task's channel still doesn't mint").
- Full `internal/team` + `internal/teammcp` pass (the only failure locally is a pre-existing `TestOfficeEvals` TempDir-cleanup flake that also fails on clean `main` and is green in CI).

## Test plan

- [x] `go build ./...`, `go vet`, `gofmt`
- [x] new regression test fails pre-fix / passes post-fix
- [x] `go test ./internal/team ./internal/teammcp` (minus the pre-existing OfficeEvals teardown flake)
- [x] `-race` on the channel/task tests

## Bug Fixes
- Fixed task-channel allocation so newly created top-level tasks initiated from within another task’s channel (and tasks that effectively resolve to the default “general” channel) now receive their own dedicated per-task channel instead of sharing the parent’s channel.

## Improvements
- Added stricter validation of per-task LLM runtime override settings at the boundary during task creation and planning for more consistent behavior.

## Tests
- Added an integration test reproducing the prior channel-sharing issue and expanded channel-minting contract coverage.

### DIFF
--- a/internal/team/broker_tasks_lifecycle.go
+++ b/internal/team/broker_tasks_lifecycle.go
@@ -278,9 +278,9 @@ func (b *Broker) EnsureTask(channel, title, details, owner, createdBy, threadID 
 	// the per-task channel deterministically.
 	b.counter++
 	taskID := b.allocateIssueIDLocked()
-	// Mint a dedicated channel for new business-objective tasks that
-	// defaulted to "general".
-	if shouldMintPerTaskChannel(channel, &teamTask{
+	// Mint a dedicated channel for a task that defaulted to "general" or was
+	// created from inside another task's chat (so it never shares one).
+	if shouldMintPerTaskChannel(channel, b.channelOwnedByAnotherTaskLocked(channel), &teamTask{
 		Title:   title,
 		Details: strings.TrimSpace(details),
 		Owner:   strings.TrimSpace(owner),
--- a/internal/team/broker_tasks_mutation_service.go
+++ b/internal/team/broker_tasks_mutation_service.go
@@ -704,11 +704,13 @@ func (b *Broker) MutateTask(body TaskPostRequest) (TaskResponse, error) {
 		// name the per-task channel "task-<id>" deterministically.
 		b.counter++
 		taskID := b.allocateIssueIDLocked()
-		// For new business-objective tasks that defaulted to "general",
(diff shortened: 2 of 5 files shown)


FranDias (dismissed):

najmuzzaman-mohammad:
Addressed CodeRabbit's outside-diff note in 52d8bad1 (FIXED): reworded the `shouldMintPerTaskChannel` doc comment to state the System / incident guards are evaluated before `ParentIssueID`, so a sub-issue that is also a system/incident task stays in #general — matching the guard test. Comment-only, no behaviour change.

FranDias (approved):

PR #1108, opened 2026-06-17
fix(tasks): migrate existing Issues squatting in another task's channel

najmuzzaman-mohammad:
## Why

#1107 stopped **new** Issues from being created into another Issue's chat, but Issues already persisted with a shared channel stay shared. Live example from the workspace:

```
OFFICE-22  channel=task-office-22   (owner of the channel)
OFFICE-28  channel=task-office-22   ← still squatting after #1107
```

## What

One-shot, idempotent boot migration `reconcileSharedTaskChannelsLocked`, run from `normalizeLoadedStateLocked` (same place as the other load-time reconcilers). For each task whose `Channel` is a per-task channel owned by a **different** task (`channel.TaskID != task.ID`), it mints the task's own `task-<id>` channel and moves the link.

**Left untouched:** the task that legitimately owns the channel, tasks in `#general`, tasks in non-per-task shared channels (no `TaskID` — project/bridged), and system/incident tasks. Only the task→channel link moves — historical messages stay in the old channel, so new activity lands in the task's own chat.

**Idempotent:** after the first pass each task owns its channel, so subsequent boots are no-ops.

## Test

`TestReconcileSharedTaskChannelsLocked_RehomesSquatters` reproduces the OFFICE-22/OFFICE-28 case: the squatter is re-homed to its own channel, the owner is untouched, `#general` and non-per-task channels are left alone, and a second pass changes nothing.

## Test plan

- [x] `go build ./...`, `go vet`, `gofmt`
- [x] `go test ./internal/team ./internal/teammcp` (exit 0)
- [x] migration unit test (re-home + idempotency + leave-alone cases)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where some tasks were incorrectly assigned to shared channels. Affected tasks are now automatically relocated to their proper dedicated channels during system startup.

### DIFF
--- a/internal/team/broker_defaults.go
+++ b/internal/team/broker_defaults.go
@@ -339,7 +339,57 @@ func (b *Broker) normalizeLoadedStateLocked() {
 		_ = b.syncTaskWorktreeLocked(&b.tasks[i])
 	}
 	b.reconcileOrphanedBlockedTasksLocked()
+	b.reconcileSharedTaskChannelsLocked()
 	b.pendingInterview = firstBlockingRequest(b.requests)
+}
+
+// reconcileSharedTaskChannelsLocked re-homes tasks that are squatting in
+// another task's dedicated per-task channel. Before the channel-minting fix,
+// a top-level Issue created from inside an existing Issue's chat inherited that
+// chat's channel, so several tasks could share one task-<id> channel (e.g.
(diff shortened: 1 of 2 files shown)


FranDias (approved):

PR #1115, opened 2026-06-18
fix(team): stop agents silently stalling — release turn slot on panic + honest note on chat-reply failure

najmuzzaman-mohammad:
Fixes for agents (CEO included) that **silently stall and never reply**. Three verified fixes; one investigated candidate dismissed as a non-bug.

---

## Fix 1 — leaked turn slot on panic (the big one)

The per-lane headless worker (`runHeadlessCodexQueue`) called `finishHeadlessTurn(lane)` as the **last statement** inside its panic-recovered closure. Any panic in `headlessCodexRunTurn`, the recovery helpers, or `recordTaskLedgerEntry` was swallowed by `recoverPanicTo` (logs a stack, returns) so `finishHeadlessTurn` **never ran**.

`finishHeadlessTurn` (1) clears `active[lane]` → frees the global/per-agent concurrency-cap slot, (2) respawns lanes parked behind the cap, and (3) **wakes the lead/CEO after a specialist completes**. Skipping it on panic leaked the active slot forever: under a cap every other agent's lane stayed parked, and the CEO never reacted to completions → silent stall, no reply, no surfaced error.

**Fix:** `defer l.finishHeadlessTurn(lane)`, registered right after a turn begins. Cleanup always runs, even on panic; `recoverPanicTo` (deferred earlier, runs last) still logs. Happy path unchanged.

**Test:** `TestHeadlessTurnPanicFreesActiveSlot` — under a global cap of 1, a panicking turn must still free its slot so a later turn runs. Fails before, passes after.

## Fix 2 — chat/DM reply that times out or errors posts nothing

When a turn is a plain chat reply (no task) and it **times out** or **fails**, `recoverTimedOutHeadlessTurn` / `recoverFailedHeadlessTurn` early-return with only an internal log line. Task turns surface via `BlockTask` + the decision inbox, but a chat reply has no task — so the user who DM'd the CEO got silence.

**Fix:** `noteChatTurnStall` posts one honest line into the channel when the failing turn has no task. Wired into the worker's timeout branch and its non-killed error branch (killed already posts its own note).

**Tests:** `TestNoteChatTurnStallTasklessVsTask`, `TestHeadlessQueueChatTimeoutPostsStallNote`.

## Fix 3 — human-prompted turn that succeeds but posts no reply

The murkier silence: a turn returns success yet posts nothing. For non-coding agents the durability guard is skipped, so it's treated as complete and the user gets silence. The provider runners already salvage a forgotten broadcast (`postHeadlessFinalMessageIfSilent` posts the model's final text when it ran but didn't call `team_broadcast`), so the only remaining case is a turn that returned with **no user-facing output at all** — which the human-priority prompt contract says should never happen (it instructs the agent to always reply, status, or acknowledge).

**Fix:** `noteChatTurnNoReply`, gated three ways to avoid false positives on intentional silence:
- `turn.FromHuman` only — agent-to-agent turns may legitimately stay quiet;
- taskless only;
- only when the agent posted no substantive message in **any** channel since the turn began (`agentPostedSubstantiveMessageSince` scans all channels, so an agent that answered elsewhere or via the salvage path isn't flagged).

Wired into the worker's `err==nil` branch.

**Tests:** `TestNoteChatTurnNoReplyGating` (full gating matrix), `TestHeadlessQueueHumanChatSilentSuccessPostsNote`.

## Investigated and dismissed — provider call ignoring `ctx.Done()`

A sub-agent flagged the OpenAI-compat streaming reader as a "critical hang" because `http.Client.Timeout` is 0. Verified against the code: the request is built with `http.NewRequestWithContext(ctx, …)` from the turn ctx (4–12m deadline) over a standard `http.Transport`. In Go's `net/http` the request context governs the **entire** response lifetime, so when the deadline fires the connection closes and the body read unblocks. The other three runners use `exec.CommandContext` + an explicit `ctx.Done()` terminate-goroutine. **No real hang** — the claim rested on a `net/http` misconception. No change made.

## Verification

- New tests fail-before / pass-after where applicable; all green under `-race`
- `bash scripts/test-go.sh ./internal/team` — green (one intermittent pre-existing timing flake, not reproducible, unrelated)
- `go vet ./internal/team/` clean; `go build ./cmd/wuphf` OK

## Summary by CodeRabbit

* **New Features**
  * Headless agent now posts system notifications when human-initiated chat requests receive no agent reply.
  * Headless agent now posts system notifications to report failed chat turns.

* **Bug Fixes**
  * Improved system reliability for concurrent turn execution and resource cleanup during failure scenarios.

### DIFF
--- a/internal/team/headless_codex_queue.go
+++ b/internal/team/headless_codex_queue.go
@@ -427,6 +427,17 @@ func (l *Launcher) runHeadlessCodexQueue(lane headlessLane, stop <-chan struct{}
 				l.updateHeadlessProgress(slug, "idle", "idle", "waiting for work", headlessProgressMetrics{})
 				return
 			}
+			// Guarantee the active slot is released even if the turn body
+			// panics. finishHeadlessTurn clears active[lane] (freeing the
+			// concurrency-cap slot), respawns parked lanes, and wakes the
+			// lead/CEO after a specialist completes. If a panic in
+			// headlessCodexRunTurn, the recovery helpers, or
+			// recordTaskLedgerEntry skipped it, the slot leaked forever:
+			// under a cap every other agent's lane stayed parked and the CEO
+			// never reacted to completions — agents silently stalled with no
+			// reply. recoverPanicTo (deferred above, so it runs last) still
+			// swallows and logs the panic after this cleanup runs.
+			defer l.finishHeadlessTurn(lane)
 			appendHeadlessCodexLatency(slug, fmt.Sprintf("stage=started queue_wait_ms=%d", time.Since(turn.EnqueuedAt).Milliseconds()))
 			l.updateHeadlessProgress(slug, "active", "queued", "queued work packet received", headlessProgressMetrics{})
 
@@ -475,7 +486,6 @@ func (l *Launcher) runHeadlessCodexQueue(lane headlessLane, stop <-chan struct{}
 				l.recoverFailedHeadlessTurn(slug, turn, startedAt, detail)
 			}
 			l.recordTaskLedgerEntry(slug, turn, startedAt, err)
-			l.finishHeadlessTurn(lane)
 		}()
 		l.headless.mu.Lock()
 		_, stillRunning := l.headless.workers[lane]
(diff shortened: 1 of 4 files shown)


FranDias (approved):

PR #1159, opened 2026-07-02
feat(operator): UI tab owns the app + artifacts, Composio onboarding gate, demo call builds the real agent

najmuzzaman-mohammad:
## What

Three operator-surface changes from the same testing session, per founder direction.

### 1. Artifacts move back into the UI tab — one app UI
The agent detail's first tab is **UI** again: the agent has exactly ONE app UI, and it IS the tab — not an artifact chip among its outputs.
- The UI tab renders the live app directly (sealed `CustomAppFrame` when ready, `AppLivePreview` while building), staying mounted across tab switches.
- The artifacts the agent's runs produce (md/html/pdf routine outcomes) collect in an `Artifacts` strip **below the app**; the section only renders when at least one exists.
- The synthetic `app`-type artifact and the `renderApp` slot are removed (the wire never carried `app`). Same restructure in the mock detail.

### 2. Composio onboarding gate on the Integrations tab
Without a Composio key there is no catalog to browse. When `composio_key_set` is false, the operator Integrations tab renders the shipped `ComposioOnboarding` (broker-driven CLI sign-in, manual key-paste fallback); the catalog query only runs once a key exists.

### 3. Demo-workflow-to-Nex builds the real AGENT
- The realtime call's instructions now frame the work as building an **agent** with three parts — app UI, routines, tools — and `draft_workflow` captures which the demo calls for (`kind: app/routine/both`), the routine's name + prompt + schedule, and the tools the agent will need.
- The build handoff routes to the **real agent build** (app-builder engine) instead of the legacy mock workflow builder — so when the build finishes, "Open the agent" lands on the app inside the agent's UI tab.
- As the new agent's id resolves, the captured routine is created on the broker scheduler and each captured tool is authored through the agent service, with honest progress messages in the build chat.

## Test plan

- [x] `bash scripts/test-web.sh` — full suite green (2346 passed)
- [x] `bunx tsc --noEmit` clean; `bunx biome check --write` on changed files (no new findings beyond repo baseline)
- [x] New/updated tests: UI tab app+artifacts composition; empty-artifacts state; ArtifactsTab outputs-only; ToolIntegrations onboarding gate (no key → onboarding, no catalog fetch); demoCapture kind/routine/tools coercion + seed; AppBuilderChat demo auto-start + routine/tool creation
- [x] Visual check: Composio onboarding renders inside the operator shell (mocked `composio_key_set: false`)
- [x] Screenshots (below)

## Screenshots

01-ui-tab-app-with-artifacts

02-ui-tab-pdf-artifact

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated **UI** tab in the operator view, keeping the live app visible while showing related outputs below it.
  * Introduced a clearer **build your agent** flow with support for app, routine, or combined demo setups.
  * Added onboarding for Composio-connected integrations, including a clearer connect state before the catalog loads.

* **Bug Fixes**
  * Improved artifact browsing so markdown, HTML, and PDF outputs switch more reliably.
  * Updated screenshots and tests to cover the new UI and onboarding states.

### DIFF
--- a/web/e2e/screenshots/composio-gate.mjs
+++ b/web/e2e/screenshots/composio-gate.mjs
@@ -1,0 +1,50 @@
+// Capture the Composio onboarding gate (PR #1159): the operator Integrations
+// tab with no Composio key connected renders the shipped ComposioOnboarding
+// (sign-in primary, key-paste fallback) instead of an empty catalog. The
+// /api/config response is mocked to force the first-run state.
+
+import process from "node:process";
+
+import { installCommonMocks, launchBrowser, shotPage } from "./lib.mjs";
+
+const BASE = process.env.BASE_URL ?? "http://localhost:5273";
+
+const OUT = process.env.WUPHF_SCREENSHOTS_OUT;
+if (!OUT) {
+  console.error("WUPHF_SCREENSHOTS_OUT is not set; run via publish.sh / driver");
+  process.exit(2);
+}
+
+const { browser, context, page } = await launchBrowser({
+  viewport: { width: 1440, height: 900 },
+});
+
+try {
+  await installCommonMocks(context);
+  // Force the no-key state regardless of the workspace's real config.
+  await context.route("**/api/config", (r) => {
+    if (r.request().method() !== "GET") return r.fallback();
+    r.fulfill({
+      contentType: "application/json",
+      body: JSON.stringify({ composio_key_set: false }),
+    });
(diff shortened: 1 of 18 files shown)


najmuzzaman-mohammad:
## Screenshots

01-composio-onboarding-gate

FranDias (approved):

PR #1167, opened 2026-07-10
fix(gbrain): treat unresolvable WUPHF_GBRAIN_COMMAND as not installed

najmuzzaman-mohammad:
## What

An explicitly configured `WUPHF_GBRAIN_COMMAND` that no longer resolves to an executable is now treated as **gbrain not installed** instead of silently falling back to the `gbrain` on PATH. Applied to both resolution copies:

- `internal/gbrain.BinaryPath` — plus a once-per-process stderr warning naming the unresolvable value, so the misconfiguration is visible.
- `internal/config.gbrainBinaryInstalled` — the implicit memory-backend default probe, so a broken explicit command resolves the default to `markdown` instead of claiming gbrain readiness through the fallback binary.

Behavior with `WUPHF_GBRAIN_COMMAND` unset is unchanged (PATH lookup still applies).

## Why

Live incident on a QA stack: the wrapper script that `WUPHF_GBRAIN_COMMAND` pointed at was reaped by macOS's periodic /tmp cleanup. The silent fallback swapped in the user-global gbrain — whose brain was lock-held by a long-running `gbrain serve` — so every broker knowledge call became a 30s hung MCP connect (`connect gbrain mcp: context deadline exceeded`) and knowledge was silently served from the per-app file cache. An explicit command exists precisely to re-home the brain; substituting the global binary defeats that and turns a fast, diagnosable failure into a slow, invisible one.

## Test plan

- [x] `TestBinaryPathExplicitCommandMissingDoesNotFallBack` — fails against the old fallback behavior, passes with the fix
- [x] `TestBinaryPathExplicitCommandResolves`, `TestBinaryPathEmptyCommandUsesPathFallback` — explicit-resolves and unset-env paths unchanged
- [x] `TestResolveMemoryBackendExplicitCommandMissingDoesNotFallBack` — implicit default resolves to markdown when the explicit command is missing
- [x] `go vet`, `gofmt`, `golangci-lint run ./internal/gbrain/...` clean
- [x] Live-verified on the QA stack: with a valid wrapper, knowledge synthesis writes through gbrain (warm reads 1.6s vs the prior 30s timeout); unit-verified the fail-fast path

No web/ changes — no screenshots.

## Summary by CodeRabbit

* **Bug Fixes**
  * Explicitly configured gbrain commands are now authoritative. If the command cannot be found, the app no longer silently falls back to another `gbrain` executable on the system path.
  * Added a clear one-time warning when the configured command is unavailable.
  * Memory backend selection now correctly falls back to Markdown when the configured gbrain command cannot be resolved.

* **Tests**
  * Added coverage for explicit commands, PATH resolution, missing commands, and fallback behavior.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -328,10 +328,14 @@ func gbrainBackendReady() bool {
 }
 
 func gbrainBinaryInstalled() bool {
+	// An explicit WUPHF_GBRAIN_COMMAND is authoritative (mirrors
+	// gbrain.BinaryPath): if it does not resolve, gbrain is not installed —
+	// no PATH fallback. The explicit command usually re-homes the brain, and
+	// substituting the user-global binary would point the implicit default at
+	// the wrong (possibly locked) brain.
 	if cmd := strings.TrimSpace(os.Getenv("WUPHF_GBRAIN_COMMAND")); cmd != "" {
-		if _, err := exec.LookPath(cmd); err == nil {
-			return true
-		}
+		_, err := exec.LookPath(cmd)
+		return err == nil
 	}
 	if _, err := exec.LookPath("gbrain"); err == nil {
 		return true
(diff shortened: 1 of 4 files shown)


FranDias (approved):

Issue #990, opened 2026-05-23
release-drift.yml fails with SIGPIPE after printing 'all surfaces agree'


najmuzzaman-mohammad:
Every Release drift check run completes the surface comparison successfully (logs end with `All four surfaces agree at v$git_version. ✓`) but then exits 2 with `grep: write error: Broken pipe` and fails the workflow.

The script runs under `set -euo pipefail`. The trailing diagnostic block uses pipes like `git tag ... | grep ... | head -n1` where `head` exits early after the first match, sending SIGPIPE up to grep, which under pipefail propagates as workflow failure.

This masks real drift signal — operators see a red drift check after every release even though the four surfaces actually agree. The original #350-class bug this workflow was designed to catch would be invisible if a real drift were ever detected, because the dashboard is permanently red.

### Repro

Inspect any recent Release-drift run, e.g. https://github.com/nex-crm/wuphf/actions/runs/26335393576. Logs show 'All four surfaces agree' on the line preceding the SIGPIPE.

### Suggested fix

Wrap the trailing diagnostic pipes with `|| true` or restructure to avoid `head` short-circuiting under pipefail. Confirmed pre-existing — fails identically on schedule and workflow_run triggers since at least v0.212.0.

### Verified for v0.212.1 manually

- git tag: v0.212.1 ✓
- GH release: https://github.com/nex-crm/wuphf/releases/tag/v0.212.1 ✓
- npm view wuphf version: 0.212.1 ✓
- smoke-test-npx (macos + ubuntu): green ✓

FranDias:
## 🔭 Multi-view review — backlog sweep 2026-06-30

**Verdict:** Keep — a live, self-defeating CI bug: the release-drift monitor is red on every run, so it can no longer catch the drift it exists to catch. Cheap to fix, real to leave.

**🎯 Strategic fit (operator-era):** wuphf is the **public OSS dev-trust funnel** in the two-funnel model — proven ideas graduate to the paid app/core. A permanently-red release-health check on the OSS artifact is exactly the kind of trust erosion that undercuts the funnel. Guarding release integrity (git ↔ GitHub Release ↔ npm ↔ npx binary) keeps the public surface credible.

**🔧 Technical state (verified vs code):** Confirmed live — the last **8/8** `release-drift` runs failed (2026-06-23 → 06-30). Root cause is at `.github/workflows/release-drift.yml:68`:
```bash
[...]

PR #1173, opened 2026-07-18
fix(ci): prevent release-drift and auto-release pipeline SIGPIPE/EPIP…

Adityakk9031:
close:#990
### Summary
Fixes a recurring failure where the `Release drift check` (`release-drift.yml`) and `Auto Release` (`auto-release.yml`) workflows fail with `grep: write error: Broken pipe` (exit code `2` / `141`). 

### Root Cause
Both workflows utilized short-circuiting pipelines to get the latest tag:
- `release-drift.yml`: `git tag --merged origin/main --sort=-v:refname | grep -E '^v[0-9]' | head -n1`
- `auto-release.yml`: `git tag --sort=-v:refname | grep '^v' | head -1`

Under `set -o pipefail` (explicitly or via GitHub Actions' default shell configuration), `head` exits immediately after printing the first line and closes its side of the pipe. When the preceding `grep` tries to write subsequent lines to the closed pipe, it receives a `SIGPIPE` / `EPIPE`. 

Because GitHub Actions runners often inherit `SIG_IGN` on `SIGPIPE`, `grep` gets an `EPIPE` on its next write, prints `grep: write error: Broken pipe`, and exits with code `2`. Under `pipefail`, this non-zero exit code propagates and fails the entire workflow job.

### Changes
1. **Removed `grep` + `head`**: Switched to native git tag filtering using the `--list` parameter.
2. **Replaced with `awk 'NR==1'`**: Used `awk 'NR==1'` as the stream selector. Unlike `head`, `awk 'NR==1'` reads the entire stdin stream to completion (but only prints the first line). This prevents the writer (`git tag`) from encountering a closed pipe and guarantees a successful `0` exit code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automated release workflows’ logic for selecting the newest valid version tag.
  * Added stricter semver-like tag matching and deterministic newest-tag sorting to ensure release creation and drift checks consistently use the correct tag.

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Get latest tag
         id: latest
         run: |
-          TAG=$(git tag --sort=-v:refname | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.]+)?$' | awk 'NR==1')
+          TAG=$(git tag --merged origin/main --sort=-v:refname | awk '/^v[0-9]+\.[0-9]+\.[0-9]+([.-][0-9A-Za-z.]+)?$/ { if (n++ == 0) print }')
           echo "tag=${TAG:-v0.0.0}" >> "$GITHUB_OUTPUT"
 
       - name: Bump version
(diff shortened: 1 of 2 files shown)


Adityakk9031:
@FranDias have a look

Adityakk9031:
@FranDias have a look

PR #1180, opened 2026-07-31
fix unauthenticated rce via /tools/call

### LINKED ISSUE #1179: unauthenticated rce via post /tools/call
# Unauthenticated RCE: `POST /tools/call` executes request-body code in a non-isolating "sandbox"

## Summary

The pi-mono agent service (`agent/`) exposes `POST /tools/call`, which takes a
JavaScript tool body **directly from the HTTP request** and runs it. Two problems
compound into unauthenticated remote code execution:

1. **The "sandbox" does not isolate anything.** `toolSandboxWorker.ts` runs
   agent-authored code with dangerous globals (`process`, `fetch`, `Bun`,
   `Function`, `require`) "shadowed" as trailing `new Function(...)` parameters
   bound to `undefined`. That only shadows them lexically; code built via the
   `Function` constructor executes in the real global scope, so the standard
   prototype-chain hop reaches the genuine globals and walks straight past the
   shadowing.

2. **`/tools/call` has no authentication and runs body-supplied code.**
   `createServer` (`agent/src/service.ts`) has no auth on any route, and
   `Bun.serve` is started with no `hostname`, so it binds `0.0.0.0` (all
   interfaces). The `/tools/call` handler reads `body.tool.code` straight from
   the request and hands it to `runTool` — there is no check that the code
   matches a previously authored/persisted tool.

Combined: **anyone who can reach the service port runs arbitrary code as the
user** — read `process.env` (which includes `WUPHF_BROKER_TOKEN` and provider
API keys), read arbitrary files via `Bun.file`, and make outbound network calls
via the real `fetch`. The `approved: false` gate does not help: it only guards
`GATED_CAPABILITIES` (`crm.assign`, `nex.send`, `nex.browser`), never raw code
execution.

This is on `main` and the service is wired into production — the broker calls it
via `WUPHF_AGENT_URL` (default `http://127.0.0.1:8820`,
`internal/team/scheduler_operator_routines.go`).

## Affected code

- `agent/src/service.ts` — `/tools/call` handler runs `body.tool.code`; no auth; no `hostname` on `Bun.serve`.
- `agent/src/toolSandboxWorker.ts` — parameter-shadowing "isolation" that the `Function`-constructor escape defeats.
- `agent/src/toolRuntime.ts` — `scanReject` only blocks literal `import(`/`eval(`, which does not catch the escape.

The worker's own header comment describes the residual authority as merely
"reduced by strict-mode parameter shadowing," with a `TODO(security)`. In
practice the shadowing provides **zero** containment — this is a full escape,
not residual risk.

## Impact

- **Severity: Critical.** Unauthenticated RCE on the host running the agent
  service, reachable off-box because the listener binds all interfaces.
- Exfiltration of the broker token and any configured LLM/Composio API keys.
- Arbitrary local file read/write and outbound network egress.
- A softer variant reaches the same sink through `/tools/build` with
  `TOOL_AUTHOR_MODEL=1`: a prompt-injected tool description (e.g. from a
  malicious record summarized into chat) can have the model author malicious
  code that is persisted and later run through the same broken sandbox.

## Reproduction (redacted)

Verified end-to-end against the real service on `main`:

1. Boot the service with a marker secret in its environment (e.g.
   `WUPHF_BROKER_TOKEN`). `lsof` confirms it listens on `*:8820` (all
   interfaces), not loopback.
2. Send a single **unauthenticated** `POST /tools/call` whose `tool.code` uses a
   prototype-chain lookup to recover the real `process` global (the specific
   one-liner is withheld here) and returns `process.env.WUPHF_BROKER_TOKEN`.
3. The response is `{"status":"ok", ...}` and contains the marker secret
   verbatim. A second payload reads an arbitrary file via `Bun.file` and
   confirms the real `fetch` is reachable — all with `approved: false` and no
   `Authorization` header.

(The full weaponized payload is intentionally omitted from this public issue.
It can be shared privately with maintainers; a fix is in the accompanying PR.)

## Suggested fix

Defense in depth (see accompanying PR):

1. **Never execute request-body code.** `/tools/call` should reference a tool by
   `(agent, name)` and resolve the code from the per-agent store — the browser/
   caller sends inputs and args, never code. (The broker's routine path already
   persists tools before running them, so this does not regress that flow.)
2. **Bind loopback by default** (`WUPHF_AGENT_HOST`, default `127.0.0.1`) and
   **refuse to boot** a non-loopback bind without a token.
3. **Require a bearer token** (`WUPHF_AGENT_TOKEN`) on every route except
   `/health`, compared in constant time.
4. (Follow-up) Replace parameter-shadowing with real process-level isolation for
   the tool runtime; the current model cannot be made safe by patching.

ben564885:
closes #1179.

`post /tools/call` ran code straight from the request body in a sandbox that doesn't isolate, with no auth and bound to all interfaces — unauthenticated rce (token/key exfil, arbitrary file read, network egress).

two-layer fix:
- bind loopback by default and require a constant-time bearer token (`wuphf_agent_token`); refuse to boot a non-loopback bind without one.
- `/tools/call` now references a saved tool by `(agent, name)` and runs stored code — never code from the request body. frontend and wire type updated to match.

verified: replayed exploit now returns 401/404; agent suite 109 pass, web suites 18 pass, both typecheck clean. does not regress the broker routine flow.

## Summary by CodeRabbit

* **New Features**
  * Added optional bearer-token authentication for service requests.
  * Added configurable hostname binding with protection against unauthenticated remote access.
  * Tools are now selected by agent and tool name when invoked.

* **Bug Fixes**
  * Prevented request payloads from injecting or overriding executable tool code.
  * Added validation for missing, invalid, unknown, or incompatible tool references.
  * Kept health checks available without authentication.

### DIFF
--- a/agent/src/service.ts
+++ b/agent/src/service.ts
@@ -14,6 +14,7 @@
 //   POST /sessions                  new manual session; POST /sessions/<id>/message append
 //   GET  /artifacts?agent=          the agent's saved run artifacts
 
+import { timingSafeEqual } from "node:crypto";
 import { streamWorkflow } from "./buildAgent.js";
 import { buildCapabilities, capabilityConfigFromEnv } from "./capabilities.js";
 import { runWorkflow } from "./executor.js";
@@ -30,12 +31,40 @@ type BuildStream = (message: string, opts: { toolId?: string; signal?: AbortSign
 
 export interface ServerOptions {
 	port?: number;
+	// Interface to bind. Defaults to loopback (WUPHF_AGENT_HOST) — this service
+	// runs code and must never be reachable off-box without an explicit opt-in.
+	hostname?: string;
+	// Shared secret required as `Authorization: Bearer <token>` on every route
+	// except /health. Defaults to WUPHF_AGENT_TOKEN. When empty the service is
+	// unauthenticated and MUST stay loopback-bound (enforced in createServer).
+	authToken?: string;
 	// Override the build engine in tests so they never hit a live model.
 	buildStream?: BuildStream;
 	// Override the per-agent store (tests point it at a tmp dir).
 	store?: AgentStore;
 	// Override the pi-backed session layer (tests point it at a tmp dir).
 	sessions?: PiSessions;
+}
+
+/** True when `host` only ever resolves to the local machine. */
+function isLoopbackHost(host: string): boolean {
+	const h = host.trim().toLowerCase();
+	return h === "" || h === "127.0.0.1" || h === "::1" || h === "localhost";
+}
+
+/** Constant-time bearer-token check. Returns true only when a non-empty token
+ * is configured AND the request presents exactly that token. */
+function authorized(req: Request, token: string): boolean {
+	if (token === "") return false;
+	const header = req.headers.get("authorization") ?? "";
+	const prefix = "Bearer ";
+	if (!header.startsWith(prefix)) return false;
+	const presented = Buffer.from(header.slice(prefix.length));
+	const expected = Buffer.from(token);
+	// timingSafeEqual throws on length mismatch — guard first so length itself
+	// is not a fast-path oracle, then compare the bytes in constant time.
+	if (presented.length !== expected.length) return false;
+	return timingSafeEqual(presented, expected);
 }
 
 function json(data: unknown, status = 200): Response {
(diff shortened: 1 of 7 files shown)


najmuzzaman-mohammad:
Integrated via #1197 (rebased onto main, additive conflict with #1196 resolved, original commit preserved). Thanks @ben564885 — this closes a real unauthenticated-RCE surface. Closing as merged.

PR #1208, opened 2026-08-18
fix(data-tab): table rendered as flexbox not a grid (class collision)

najmuzzaman-mohammad:
A user hit this live: the Data tab table is unreadable with real multi-column data — headers and body cells don't line up.

**Root cause (two bugs):**
1. **Class collision** — the table `<tr>` used `opr-data-row`, which is *also* a `display:flex` key-value row in another component (operator-shell.css:4020). The tbody rows inherited that flex, so the header laid out as table columns (1695px wide) while the body packed left as flexbox (579px) — they never aligned. Renamed the table row to `opr-dt-row`.
2. **width:100% sort button** — forced the first column to grab all the slack in auto table-layout. Sizes to content now.

**Verified live:** header row width == body row width (1215px == 1215px), `tr` is `table-row`, columns align, dates and empty cells render cleanly.

Note: this was a pre-existing collision the Data-tab work inherited — my earlier data-tab '8' was overstated while the table was visually broken with real data.

## Test plan
- [x] AppDataTab.test.tsx 17 pass (text-based, unaffected by the rename); biome clean; verified live on a fresh workspace with 9-column data.

https://claude.ai/code/session_01V9n6uZhopKDdZoXQ37ry17

### DIFF
--- a/web/src/operator/surfaces/AppDataTab.tsx
+++ b/web/src/operator/surfaces/AppDataTab.tsx
@@ -586,7 +586,7 @@ function ModelTableView({ table }: { table: ModelTable }) {
                   return (
                     <Fragment key={`row-${start + i}`}>
                       <tr
-                        className={`opr-data-row${open ? " is-open" : ""}`}
+                        className={`opr-dt-row${open ? " is-open" : ""}`}
                         onClick={() => setOpenRow(open ? null : row)}
                       >
                         {table.columns.map((c) => (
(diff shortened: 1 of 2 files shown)


najmuzzaman-mohammad:
Landed directly on main as dabfb323b (cherry-picked; content identical). Direct-to-main is the working mode now — solo builder, no PR ceremony.

PR #1227, opened 2026-09-03
Rename the product concept Agent to Bot (wire stable)

najmuzzaman-mohammad:
The office's teammates are **Bots** now. The sidebar reads BOTS, the button reads New Bot, the wizard reads Create bot, and the prompts, docs and website say bot throughout.

Scope was **surface + code, wire stable**. The wire is what made this dangerous, so the rename ran as scripted passes over a clean tree rather than by hand, and two checkable rules did the safety work:

1. **Never rewrite inside a quoted string.** Struct tags, routes, disk paths and MCP tool names all live in strings.
2. **Never rewrite a token containing `_`.** snake_case IS the wire vocabulary in both languages.

Passes, in order: Go and TS identifiers → user-visible copy (a string a human reads has a space in it, with a denylist for headers, storage keys, routes and slugs) → JSX prose → filenames and directories with every import repointed → docs, prompts and website prose with fenced code blocks protected, because those quote the wire back verbatim.

## Deliberately still "agent"

Renaming any of these breaks a live install:

| | |
|---|---|
| json keys | `agent_slug`, `agent_id`, `agents`, `one_on_one_agent` |
| routes | `/api/agents/`, `/agent-stream/`, `/agent-logs` |
| MCP tools | `agent_message`, `agent_slug` |
| disk state | `~/.wuphf/agents/<slug>/` |
| env + header | `WUPHF_AGENT_SLUG`, `User-Agent` |
| external | OpenClaw session keys (`agent:main:main`), the `one` CLI's `--agent` flag, LobeHub's `agents/{name}.md` |
| TS wire fields | `agentSlug`, `agentId`, `agentName`, `agentRole` |

That last row is the uncomfortable seam, and it is deliberate. In TypeScript those spellings are simultaneously the commonest local names and live wire field names, and **neither tsc nor the test suite would catch a field renamed out from under the server**. Type names, component names and filenames rename freely; only the field spellings freeze.

Not the product concept, so untouched: `AGENTS.md`, `docs/agents/`, `.claude/agents/`, `CLAUDE.md`, and the `wuphf-agent` pi-mono build engine.

## Behaviour changes that fall out of this

- The slash commands are `/bot` and `/bots` now (the 1:1 blacklist and the TUI command set track them).
- The starter-routine label strips a trailing `Bot` **or** `Agent` from an app name, so apps built before this commit still get the same label.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt` — clean
- web `tsc --noEmit`, `biome check` — clean
- **2533/2533 web tests pass**
- Every Go package passes **except `internal/team`**, whose 9 failures are **pre-existing on clean `origin/main`** — upstream `2548575e0` seeds two system skills without updating the tests that assert skill counts. Confirmed by running those exact tests against a clean `origin/main` worktree. Filed as B7 in the backlog doc; left alone because that area is under active work in another session.

Live-checked in gstack browser against a real office: sidebar reads BOTS / New Bot, status bar reads "2 bots", wizard heading reads "Create bot" at 12:1 contrast. The only remaining "Agent" on screen is "Reporting Agent" — a user-created app name in office state, i.e. data, correctly untouched.

Screenshots to follow on this PR.


najmuzzaman-mohammad:
## Screenshots

01-sidebar-bots

02-create-bot-dark

03-blocking-ask-card

najmuzzaman-mohammad:
Follow-up: main was red before this branch (upstream `2548575e0` seeded two system skills without updating the tests that assert skill counts, plus a govet shadow). Rather than merge onto a red base, this PR now also fixes that — including two real defects it exposed:

- `brokerStateActivityScore` counted every skill, so an empty office scored above zero from the auto-seeded pair. That score gates the last-good snapshot, so a clobbered save could overwrite the snapshot protecting real work.
- The skill backfill swept even when the wiki backend had failed to init, creating directories under a root the caller believed was dead — which is what was racing `t.TempDir` cleanup.

All 13 checks green.

